PR #16, opened 2018-12-11
Add Travis CI

yongtang:
This PR adds Travis CI to invoke several tests.


yongtang:
@terrytangyuan I have updated the PR. Now Travis CI build both Python 2.7 and R and all passes.

There are still some issues with Python 3 but I think I could do it in a follow up PR.

PR #21, opened 2018-12-13
Support Apache Parquet format

### LINKED ISSUE #12: Support parquet format for tensorflow-io
There are quite some interests to support parquet for tensorflow and a PR (tensorflow/tensorflow#19461) has already been approved.

With the expected deprecation of `tf.contrib`, it makes sense to move PR tensorflow/tensorflow#19461 to tensorflow-io.

At the moment the blocking issue with the related PR is the bazel version. The PR requires bazel 1.7.1+ to incorporate boost library while tensorflow CI still uses bazel 1.5.0. We will need to find out a way to work out the bazel version issue. See https://github.com/tensorflow/tensorflow/issues/22449, https://github.com/tensorflow/tensorflow/pull/22964 for details.

yongtang:
This fix is the continuation of PR tensorflow/tensorflow#19461, which adds Apache Parquet format support in tensorflow-io.

The PR tensorflow/tensorflow#19461 essentially was cherry-picked with some additional fix ups to correct import path.

This fix fixes #12


yongtang:
The last commit in this PR is not necessary, once the other PR #22 is merged in.

yongtang:
The PR is updated to capture PR #22, think the PR is complete.

third_party/parquet.patch
```diff
@@ -0,0 +1,7719 @@
+diff -ru -p1 --new-file src/parquet/parquet_types.cpp src/parquet/parquet_types.cpp
```
BryanCutler: was this file intentional to add?

BryanCutler: nvm, I see it is applied as part of the parquet build

tensorflow_io/parquet/BUILD
```diff
@@ -0,0 +1,62 @@
...
+    name = 'python/ops/_parquet_ops.so',
+    srcs = [
+        "kernels/parquet_dataset_ops.cc",
+        "ops/dataset_ops.cc",
+    ],
+    linkshared = 1,
+    deps = [
+        "@local_config_tf//:libtensorflow_framework",
+        "@local_config_tf//:tf_header_lib",
+	"@parquet//:parquet",
```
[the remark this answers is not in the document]
yongtang: Thanks @BryanCutler, the alignment has been updated.

third_party/parquet.patch
yongtang: Thanks @BryanCutler. The patch file actually just added two files:
```
src/parquet/parquet_types.cpp
src/parquet/parquet_types.h
```

Those two files are thrift type files. The files are generated automatically. This is also a one time effort for each version and will not change when regenerate. However, it need both `flex` and `bison` to generate. I tried to build `flex` and `bison` through bazel but realized that they are more complicated than thrift itself. So I think maybe we could just build those two thrift files offline (and not to build flex and bison to generate those two files in compile time).

Also, I could not find an option in Bazel to support "override" or "additional extra" file on top of the `http_archive` to add those two files in the build. So I use a patch.

BryanCutler: Ah ok, sounds good. Thanks for the explanation!

yongtang:
@BryanCutler Thanks! Yes adding to root BUILD file is needed. The PR has been updated.

third_party/parquet.patch
mhong: Would it be useful to document how to build this patch, either as a comment block or a separate "build instructions" file? It might be useful if we want to refresh this patch down the road.

mhong: Also, do we want to check in the thrift type files used to generate this patch?

tensorflow_io/parquet/kernels/parquet_dataset_ops.cc
```diff
@@ -0,0 +1,331 @@
...
+
+      Status GetNextInternal(IteratorContext* ctx,
+                             std::vector<Tensor>* out_tensors,
+                             bool* end_of_sequence) override {
+        mutex_lock l(mu_);
+
+        // Loop until we find a row to read or there are no more files left to
+        // read
+        while (true) {
+          if (parquet_reader_) {
```
mhong: minor: consider inverting the if condition (`if (!parquet_reader_)`) to reduce indentation?

tensorflow_io/parquet/kernels/parquet_dataset_ops.cc
```diff
@@ -0,0 +1,331 @@
...
+      explicit Iterator(const Params& params)
+          : DatasetIterator<Dataset>(params) {}
+
+      Status GetNextInternal(IteratorContext* ctx,
+                             std::vector<Tensor>* out_tensors,
+                             bool* end_of_sequence) override {
+        mutex_lock l(mu_);
+
+        // Loop until we find a row to read or there are no more files left to
+        // read
```
mhong: consider extending this comment a bit more to describe the conceptual hierarchy. If I understand correctly, the hierarchical levels from coasest to finest are: current_file_index_ -> current_row_group_ -> current_row_?

tensorflow_io/parquet/kernels/parquet_dataset_ops.cc
mhong: Also to further reduce indentation, should we refactor some code into a helper function, say a helper to read all row groups from the current file (the body of the while loop below)?

tensorflow_io/parquet/kernels/parquet_dataset_ops.cc
```diff
@@ -0,0 +1,331 @@
...
+            *end_of_sequence = true;
+            return Status::OK();
+          }
+
+          TF_RETURN_IF_ERROR(SetupStreamsLocked(ctx->env()));
+        }
+      }
+
+      template <typename DType>
+      Status FillTensorValue(parquet::ColumnReader* column_reader,
```
mhong: should this and some functions below be private?

tensorflow_io/parquet/ops/dataset_ops.cc
```diff
@@ -0,0 +1,34 @@
...
+namespace tensorflow {
+
+REGISTER_OP("ParquetDataset")
+    .Input("filenames: string")
+    .Input("columns: int32")
+    .Output("handle: variant")
+    .Attr("output_types: list(type) >= 1")
+    .Attr("output_shapes: list(shape) >= 1")
+    .SetIsStateful()
+    .SetShapeFn(shape_inference::ScalarShape);  // TODO (yongtang): check that
```
mhong: i saw some shape checking code above. Has this TODO been addressed?

tensorflow_io/parquet/python/kernel_tests/parquet_test.py
```diff
@@ -0,0 +1,77 @@
...
+    with self.test_session() as sess:
+      sess.run(init_op)
+      for _ in range(num_repeats):  # Dataset is repeated.
+        for i in range(500): # 500 rows.
+          v0 = True if ((i % 2) == 0) else False
+          v1 = i
+          v2 = i * 1000 * 1000 * 1000 * 1000
+          v4 = 1.1 * i
+          v5 = 1.1111111 * i
+          self.assertAllClose((v0, v1, v2, v4, v5), sess.run(get_next))
```
mhong: very nice!

mhong:
LGTM. Thank you!

WORKSPACE
```diff
@@ -61,3 +61,74 @@ http_archive(
...
+
+# Parquet needs generated parquet_types.h and parquet_types.cpp which are generated
+# from src/parquet/parquet.thrift in apache-parquet-cpp-1.4.0.tar.gz.
+#
+# Generating parquet_types.h and parquet_types.cpp, however, needs both bison and flex
+# installed, which is really an unnecessary step.
+#
+# We use the following step to generate the parquet_types.h and parquet_types.cpp files:
+#  - In third_party directory, run `docker run -i -t --rm -v $PWD:/v -w /v ubuntu:16.04 bash -x /v/parquet.type`
+#  - Once complete, a parquet.patch file will be generated which could be used as a patch in bazel
```
BryanCutler: Thanks for adding this note!

PR #25, opened 2018-12-13
Update to build tensorflow-io pypi.org packages

yongtang:
This fix made several updates so that we could build tensorflow-io pypi.org packages:
- Use auditwheel to repair whl files so that they are compatible with pypi.org
- Update version string to 0.1.0
- Add release.sh to build 2.7, 3.4, 3.5, 3.6 packages. Commands to run:
```
docker run -it -v ${PWD}:/working_dir -w /working_dir  tensorflow/tensorflow:custom-op \
    bash -x /working_dir/release.sh 2.7
docker run -it -v ${PWD}:/working_dir -w /working_dir  tensorflow/tensorflow:custom-op \
    bash -x /working_dir/release.sh 3.4
docker run -it -v ${PWD}:/working_dir -w /working_dir  tensorflow/tensorflow:custom-op \
    bash -x /working_dir/release.sh 3.5
docker run -it -v ${PWD}:/working_dir -w /working_dir  tensorflow/tensorflow:custom-op \
    bash -x /working_dir/release.sh 3.6
```

I think we will be ready to cut a initial 0.1.0 version release after this PR.

### DIFF
--- a/build_pip_pkg.sh
+++ b/build_pip_pkg.sh
@@ -59,7 +59,7 @@ function main() {
   
   python setup.py bdist_wheel
 
-  cp dist/*.whl "${DEST}"
+  auditwheel repair -w "${DEST}" dist/*.whl
   popd
   rm -rf ${TMPDIR}
   echo $(date) : "=== Output wheel file is in: ${DEST}"
--- a/build_pip_pkg.sh
+++ b/build_pip_pkg.sh
@@ -36,6 +36,11 @@ function main() {
     exit 1
   fi
 
+  if [[ "$(auditwheel --version || true)" != *"installed"* ]]; then
+    echo "No auditwheel installed"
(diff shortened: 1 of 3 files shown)


build_pip_pkg.sh
```diff
@@ -59,7 +59,7 @@ function main() {
   
   python setup.py bdist_wheel
 
-  cp dist/*.whl "${DEST}"
+  auditwheel repair -w "${DEST}" dist/*.whl
```
terrytangyuan: We should probably note this at the beginning of the script as a requirement since most users might not have this installed.

yongtang:
Thanks @terrytangyuan, The PR has been updated with additional checks added.

terrytangyuan:
Looks good. Thanks!

PR #28, opened 2018-12-14
Update docs for installation

yongtang:
This fix updates docs for installation through pip (pypi.org).

The example code has also been updated.

This will make us ready to release 0.1.0.


yongtang:
Added additional description for modules supported.

yongtang:
Merge to create a release..

Issue #10, opened 2018-12-09
Support libsvm for tensorflow-io


yongtang:
In TensorFlow, libsvm format is supported through `tf.contrib.libsvm` which converts libsvm format into sparse tensor format. With the expected deprecation of `tf.contrib`, the libsvm will be deprecated as well.

It would be great to provide continued support of libsvm through tensorflow-io package. While the `tf.contrib.libsvm` converts libsvm format into sparse tensor, I think in tensorflow-io, we could convert libsvm to `tf.data` directly.

PR #32, opened 2018-12-19
Add libsvm dataset support

yupbank:
Address https://github.com/tensorflow/io/issues/10

add `make_libsvm_dataset` function, which returns a dataset contains (feature, label) per row.
```
def make_libsvm_dataset(file_names,
                         num_features,
                         dtype=None,
                         label_dtype=None,
                         batch_size=1,
                         compression_type='',
                         buffer_size=None,
                         num_parallel_parser_calls=None,
                         drop_final_batch=False,
                         prefetch_buffer_size=0):
```

### DIFF
--- a/tensorflow_io/libsvm/BUILD
+++ b/tensorflow_io/libsvm/BUILD
@@ -33,7 +33,7 @@ py_test(
     srcs = [
 	"python/kernel_tests/decode_libsvm_op_test.py"
     ],
-    main = "python/kernel_tests/decode_libsvm_ops_test.py",
+    main = "python/kernel_tests/decode_libsvm_op_test.py",
     deps = [
         ":libsvm_ops_py",
     ],
--- a/tensorflow_io/libsvm/BUILD
+++ b/tensorflow_io/libsvm/BUILD
@@ -12,7 +12,6 @@ cc_binary(
     deps = [
         "@local_config_tf//:libtensorflow_framework",
         "@local_config_tf//:tf_header_lib",
-        "@kafka//:kafka",
     ],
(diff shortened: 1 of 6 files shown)


tensorflow_io/libsvm/BUILD
```diff
@@ -0,0 +1,54 @@
...
+    name = "python/ops/_libsvm_ops.so",
+    srcs = [
+        "kernels/decode_libsvm_op.cc",
+        "ops/libsvm_ops.cc",
+    ],
+    linkshared = 1,
+    deps = [
+        "@local_config_tf//:libtensorflow_framework",
+        "@local_config_tf//:tf_header_lib",
+        "@kafka//:kafka",
```
yongtang: This this dips is not needed?

yongtang:
Overall looks good, though I am wondering if we could expose a class interface such as `class LibSVMDataset(dataset_ops.DatasetSource)`? Maybe we could add the class interface on top of the current implementation?

yupbank:
it is hard, since we only have a parsing kernel for now, we need to implement a datasource kernel to support that basically. 

if it is really worth it, i can make a second pr to port current paring kernel into datasource kernel 

and the function pattern is also tensorflow core https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/data/experimental/ops/readers.py#L311

tensorflow_io/libsvm/BUILD
yupbank: thank you!

yongtang:
@yupbank Added a PR #38 to fix some minor issues. Please take a look.

PR #34, opened 2018-12-20
Fix Ignite build so that .py files included into artifact

dmitrievanthony:
Currently when I try to use Ignite the following exception is thrown:

`ModuleNotFoundError: No module named 'tensorflow_io.ignite.python.ops.ignite_dataset_ops'`

The reason is that a file `ignite_dataset_ops` is not included into the result artifact. Changes in this PR fixes this issue.


yongtang:
Thanks for the fix!

PR #38, opened 2018-12-22
Add libsvm to BUILD file and fixed import path

yongtang:
so that it is part of the pip package


yupbank:
thank you!!

Issue #6, opened 2018-12-09
R Interface Prototype


terrytangyuan:
- [x] Package setup and testing infrastructure
- [x] Implement R wrappers + unit tests
- [x] Integration with [tfdatasets](https://github.com/rstudio/tfdatasets) R package
- [x] Setup CI for R package and make sure tests all pass
- [x] Examples/vignettes

terrytangyuan:
This is done. Next steps can be found in #18

PR #40, opened 2018-12-26
Added R wrappers for released dataset ops

terrytangyuan:
This PR is on top of https://github.com/tensorflow/io/pull/39. The wrappers were scaffolded using scripts created in #39 and then manually edited for cosmetic purposes as well as additional checks for arguments. 

* Added R wrappers for released dataset ops, including kafka, kinesis, hadoop, and ignite. Note that only wrappers for the released dataset ops are included here - parquet and libsvm support are not included yet. 
* Re-exported useful functions from `tfdatasets` package so `tfio` would play nicely with `tfdatasets`.
* Removed unnecessary dependencies and fixed `R CMD check` errors and warnings.

This is part of #6.


terrytangyuan:
Good call. I’ll add some test cases and instructions (maybe on top of your README PR). I am marking this PR as WIP for now and will ping when ready.

yongtang:
Thanks @terrytangyuan. The repo have a simple sequence file:
https://github.com/tensorflow/io/tree/master/tensorflow_io/hadoop/python/kernel_tests/testdata

The sample sequence file consists of key-value pairs:
https://github.com/tensorflow/io/blob/fc10234921c63080ac355fb49ca25b2f500b128f/tensorflow_io/hadoop/python/kernel_tests/hadoop_test.py#L37-L41

Let me know if I can be of any help. I am interested in playing with R as well 😃

terrytangyuan:
@yongtang Thanks. I fixed the error you encountered and added instruction to README.md as an alternative to the instruction you added in the other PR. Also added a very simple test case (we can add more strict ones later).

I haven't used `tensorflow/tensorflow:custom-op` image too often yet but the [r-base image](https://hub.docker.com/_/r-base/) is used more often in the R community and solves some issues with certain R package dependencies that we don't have to deal with. We can investigate some time to make `tensorflow/tensorflow:custom-op` work too later on.

yongtang:
@terrytangyuan I tried locally and it works. Thanks! We could add some additional test cases later.

PR #44, opened 2019-01-03
Update Apache Arrow to 0.11.1

yongtang:
This fix updates Apache Arrow to 0.11.1. Since Parquet is part of the Apache Arrow now, the Parquet library has been removed.

The Parquet headers are still generated like before, except it is included as part of the `arrow/cpp/src/parquet`.


yongtang:
@BryanCutler It looks like the Parquet Cpp library is now part of the Arrow library. I update the arrow dependency (with parquet related cpp files added) and removed the parquet dependency in the PR. Please take a look.

BryanCutler:
Thanks for doing this! For #36 I also needed a number of source files in `arrow/ipc` and `arrow/io` as well as the Flatbuffers library, which I added [here](https://github.com/tensorflow/io/pull/36/files#diff-3d9afc4dcae8956fef470785cc5cf75aR12).  Would you like me to make a PR to add them here or just update my branch once this is merged?

yongtang:
Thanks @BryanCutler. I think we could merge this PR now, then additional needed source files (in arrow.BUILD) could be added from your arrow PR there. Let me merge this PR.

BryanCutler:
> Thanks @BryanCutler. I think we could merge this PR now, then additional needed source files (in arrow.BUILD) could be added from your arrow PR there. Let me merge this PR.

Sounds good to me, thanks!

PR #47, opened 2019-01-08
Update flatbuffers to b99332e

yongtang:
Since google/flatbuffers#5104 has been merged, it is possible to use flatbuffers directly without patch. This fix update the flatbuffers to the latest one.

This fix is related to #36.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,6 +19,7 @@ matrix:
         - 2.7
       before_install:
         - bash -x ${TRAVIS_BUILD_DIR}/.travis/python.configure.sh
+        - pip install pyarrow==0.11.1
       script:
         - bazel test --show_progress=no -s --verbose_failures --test_output=errors -- //tensorflow_io/...
     - language: python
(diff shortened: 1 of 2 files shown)


yongtang:
@BryanCutler I created this PR, as google/flatbuffers#5104 has been merged and there is no need to apply a fix any more.

With this PR, you can just add the following to `flatbuffer_cc_library` in `arrow.BUILD`. I think that will be all needed.

```diff
-    out_prefix = "cpp/src/arrow/ipc/"
+    out_prefix = "cpp/src/arrow/ipc/",
+    flatc_args = [
+        "--no-union-value-namespacing",
+        "--gen-object-api",
+    ],
```

BryanCutler:
Ok, will do. Thanks @yongtang !

.travis/python.configure.sh
```diff
@@ -9,3 +9,4 @@ chmod +x install.sh
 rm -f install.sh
 # Configure TensorFlow
 ./configure.sh
+pip install pyarrow
```
BryanCutler: should this be fixed to `pyarrow==0.11.1`?

yongtang:
Thanks @BryanCutler. I updated the pyarrow version. It looks like pyarrow does not support python 3.4, so skipped the install for 3.4.

BryanCutler:
> It looks like pyarrow does not support python 3.4, so skipped the install for 3.4.

Yeah, pyarrow stopped building for Python 3.4 since it is scheduled to be end-of-life in March this year. I should make the tests that require pyarrow optional, since we are testing against this version currently.

PR #48, opened 2019-01-12
Add Google Cloud PubSub support

### LINKED ISSUE #31: Support Cloud PubSub for tensorflow-io
The tensorflow-io support a basic Kinesis integration which is a cloud-vendor specific (AWS) streaming platform.

A similar streaming platform provided by Google Cloud is the Cloud PubSub. It makes sense to provide support for Cloud PubSub in tensorflow-io as well.

One good thing with PubSub is that it has gRPC endpoint, which should make it a lot easier to implement in C++ than BigQuery (which only has RESTful API endpoint).

yongtang:
This PR is part of the effort to support Google Cloud PubSub, which is a vendor specific cloud streaming platform on GCP. The concept of PubSub is similar to Apache Kafka.

Google Cloud PubSub does not have a C++ SDK, though it has a gRPC endpoint so this implementation is based on gRPC.

This implementation is tested with google cloud pubsub emulator for now. Will need additional effort to bring GCP IAM support in place.

This PR fixes #31.


yongtang:
/cc @terrytangyuan This PR is also ready. The build is enabled in macOS and Linux. The test is done in Linux (skipped on macOS for now as there seems to be a bug in google SDK's macOS support?).

Please take a look

tests/test_kafka/kafka_test.sh
```diff
@@ -29,8 +29,7 @@ if [[ "$(uname)" == "Darwin" ]]; then
     (cd kafka_2.11-0.10.1.0/ && bin/kafka-server-start.sh -daemon config/server.properties)
     echo -e "D0\nD1\nD2\nD3\nD4\nD5\nD6\nD7\nD8\nD9" > kafka_2.11-0.10.1.0/test
     echo Wait 15 secs until kafka is up and running
-    # Note somehow it takes longer time in Darwin
-    sleep 15
+    sleep 5
```
terrytangyuan: This doesn't take longer anymore in Darwin?

PR #53, opened 2019-01-14
Update ffmpeg bazel rules

yongtang:
This fix updates ffmpeg bazel rules. The issue was that,
genrule (previously used) does not guarantee the files will
be generated from invoking native.cc_library. That means
occasionally `bazel build build_pip_pkg` does not work.

This fix add those stub .so files directly to avoid missing
.so files when invoking `bazel build build_pip_pkg`


yongtang:
Let's merge this one as otherwise the build will not only pass test yet the pip whl file generation will be broken.

Issue #55, opened 2019-01-14
TensorFlow 2.0 support


yongtang:
This issue is created to track TensorFlow 2.0 support for tensorflow-io. The most visible changes in TensorFlow 2.0 is the eager execution in default. Will need to test thoroughly and see if there are any impact on tensorflow-io.

yongtang:
When trying tensorflow-io's build with TF 1.13, I noticed that the generated binary for 1.13 may not be compatible with TF 2.0. Here is some core dump:
```
(gdb) where
#0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:51
#1  0x00007fd30249c801 in __GI_abort () at abort.c:79
#2  0x00007fd3024e5897 in __libc_message (action=action@entry=do_abort, fmt=fmt@entry=0x7fd302612b9a "%s\n") at ../sysdeps/posix/libc_fatal.c:181
#3  0x00007fd3024ec90a in malloc_printerr (str=str@entry=0x7fd302610d88 "free(): invalid pointer") at malloc.c:5350
#4  0x00007fd3024f3e1c in _int_free (have_lock=0, p=0x25f6798, av=0x7fd302847c40 <main_arena>) at malloc.c:4157
#5  __GI___libc_free (mem=0x25f67a8) at malloc.c:3124
#6  0x00007fd2e127b8d8 in __gnu_cxx::new_allocator<std::string>::deallocate(std::string*, unsigned long) ()
   from /usr/local/lib/python3.6/dist-packages/tensorflow_io/image/python/ops/_image_ops.so
#7  0x00007fd2e1279c28 in std::_Vector_base<std::string, std::allocator<std::string> >::_M_deallocate(std::string*, unsigned long) ()
   from /usr/local/lib/python3.6/dist-packages/tensorflow_io/image/python/ops/_image_ops.so
#8  0x00007fd2e12789fd in std::_Vector_base<std::string, std::allocator<std::string> >::~_Vector_base() ()
   from /usr/local/lib/python3.6/dist-packages/tensorflow_io/image/python/ops/_image_ops.so
#9  0x00007fd2e12765e7 in std::vector<std::string, std::allocator<std::string> >::~vector() ()
   from /usr/local/lib/python3.6/dist-packages/tensorflow_io/image/python/ops/_image_ops.so
#10 0x00007fd2e1280d32 in tensorflow::OpDefBuilder::~OpDefBuilder() ()
[...]

PR #56, opened 2019-01-15
TensorFlow 1.13 support

### LINKED ISSUE #54: TensorFlow 1.13 support
There are some `tf.data` API changes between TensorFlow 1.12 and 1.13 which breaks the build of tensorflow-io. This issue is created to track the the effort needed to provide TensorFlow 1.13 support for tensorflow-io.

yongtang:
This PR tries to add TensorFlow 1.13 support. As 1.13 has not been released yet, this PR is a working in progress until the release of 1.13.

Since 1.13 have breaking changes for Dataset API, this PR can only be merged with 1.13 released.

This PR fixes #54.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -52,7 +52,7 @@ from setuptools import setup
 from setuptools.dist import Distribution
 
 REQUIRED_PACKAGES = [
-    'tensorflow == 1.12.0',
+    'tensorflow == 1.13.0rc2',
 ]
 __version__ = '{}'
 project_name = '{}'
--- a/tensorflow_io/arrow/python/ops/arrow_dataset_ops.py
+++ b/tensorflow_io/arrow/python/ops/arrow_dataset_ops.py
@@ -19,15 +19,17 @@ from __future__ import print_function
 
 import io
 
+import tensorflow
+
+from tensorflow import dtypes
+from tensorflow.compat.v1 import data
 from tensorflow_io import _load_library
-
-from tensorflow.python.data.ops import dataset_ops
-from tensorflow.python.data.util import nest
-from tensorflow.python.framework import dtypes
-from tensorflow.python.framework import ops
-from tensorflow.python.framework import tensor_shape
-
 arrow_ops = _load_library('_arrow_ops.so')
+
+if hasattr(tensorflow, "nest"):
+  from tensorflow import nest
+else:
+  from tensorflow.python.data.util import nest
 
 
 def arrow_to_tensor_type(pa_t):
@@ -76,11 +78,11 @@ def arrow_schema_to_tensor_types(schema):
   """
   type_shape_list = [arrow_to_tensor_type(field.type) for field in schema]
   tensor_types, shape_dims = zip(*type_shape_list)
-  tensor_shapes = tuple(tensor_shape.TensorShape(s) for s in shape_dims)
+  tensor_shapes = tuple(tensorflow.TensorShape(s) for s in shape_dims)
   return tensor_types, tensor_shapes
 
 
-class ArrowBaseDataset(dataset_ops.DatasetSource):
+class ArrowBaseDataset(data.Dataset):
   """Base class for Arrow Datasets to provide columns used in record batches
   and corresponding output tensor types, shapes and classes.
   """
@@ -90,11 +92,15 @@ class ArrowBaseDataset(dataset_ops.DatasetSource):
     self._output_types = output_types
     self._output_shapes = output_shapes or \
         nest.map_structure(
-            lambda _: tensor_shape.TensorShape(None), self._output_types)
+            lambda _: tensorflow.TensorShape(None), self._output_types)
+    super(ArrowBaseDataset, self).__init__()
+
+  def _inputs(self):
+    return []
 
   @property
   def output_classes(self):
-    return nest.map_structure(lambda _: ops.Tensor, self._output_types)
+    return nest.map_structure(lambda _: tensorflow.Tensor, self._output_types)
 
   @property
   def output_shapes(self):
@@ -128,8 +134,12 @@ class ArrowDataset(ArrowBaseDataset):
       output_shapes: TensorShapes of the output tensors or None to
                      infer partial
     """
+    self._columns = columns
+    self._output_types = output_types
+    self._output_shapes = output_shapes or \
+        nest.map_structure(
+            lambda _: tensorflow.TensorShape(None), self._output_types)
     import pyarrow as pa
-    super(ArrowDataset, self).__init__(columns, output_types, output_shapes)
     if isinstance(record_batches, pa.RecordBatch):
       record_batches = [record_batches]
     assert record_batches
@@ -138,10 +148,11 @@ class ArrowDataset(ArrowBaseDataset):
     for batch in record_batches:
       writer.write_batch(batch)
     writer.close()
-    self._serialized_batches = ops.convert_to_tensor(
+    self._serialized_batches = tensorflow.convert_to_tensor(
         buf.getvalue(),
         dtype=dtypes.string,
         name="serialized_batches")
+    super(ArrowDataset, self).__init__(columns, output_types, output_shapes)
 
   def _as_variant_tensor(self):
     return arrow_ops.arrow_dataset(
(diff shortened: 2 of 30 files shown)


yongtang:
Looks like almost all tests passed. Though Travis CI's python 3.6 will crash when import tf-nightly.

We could wait until tf 1.13rc is out to take another look. We also could decide which version or platform we plan to support.

yongtang:
Think I found where ` undefined symbol` comes from. In setup.py the requirement is `tensorflow`, not tf-nightly. So even if we install tf-nightly it will still reinstall tensorflow. That will cause the api mismatch.

Change setup.py to use tf-nightly will fix the issue.

tensorflow_io/hadoop/python/kernel_tests/hadoop_test.py
```diff
@@ -48,7 +49,7 @@ def test_sequence_file_dataset(self):
 
     dataset = hadoop_dataset_ops.SequenceFileDataset(filenames).repeat(
         num_repeats)
-    iterator = dataset.make_initializable_iterator()
+    iterator = dataset_ops.make_initializable_iterator(dataset)
```
dmitrievanthony: I guess it would be better to use `tf.compat.v1.data.make_one_shot_iterator(dataset)` as it's mentioned in https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/data/ops/dataset_ops.py#L1358. What do you think?

tensorflow_io/ignite/python/tests/ignite_dataset_test.py
```diff
@@ -66,7 +67,7 @@ def _check_dataset(self, dataset):
     self.assertEqual(dtypes.string, dataset.output_types["val"]["NAME"])
     self.assertEqual(dtypes.int64, dataset.output_types["val"]["VAL"])
 
-    it = dataset.make_one_shot_iterator()
+    it = dataset_ops.make_one_shot_iterator(dataset)
```
dmitrievanthony: Would be great to update README.md as well. We can do it in the same PR (you can take [this version](https://github.com/tensorflow/tensorflow/blob/4afa2ba93dd606fd3715ef5693dee86fc25e0c77/tensorflow/contrib/ignite/README.md) as a reference) or I can do it in a separate PR.

BryanCutler:
@yongtang I figured out the issue with the ArrowDatasets and will put up the fix soon

tensorflow_io/hadoop/python/kernel_tests/hadoop_test.py
yongtang: Thanks @dmitrievanthony. Updated and using `tensorflow.compat.v1.data` now.

yongtang:
I think we are almost complete for 1.13 support. Unfortunately R may need to be updated, but this could be done later.

Also this PR will not be merged until TensorFlow 1.13 is released (At the moment only `1.13rc0`). I assume it will happen in several weeks or a month or two.

yongtang:
Also cc @terrytangyuan @BryanCutler FYI.

yongtang:
I am continuing to play with tf 1.12.0 vs. 1.13.0. Noticed that the C++ binaries build against 1.12.0 will not work with 1.13.0, as the following has changed:
```
# 1.12.0:
tensorflow::kernel_factory::OpKernelRegistrar::InitInternal( \
  tensorflow::KernelDef const*, absl::string_view, \
  tensorflow::OpKernel* (*)(tensorflow::OpKernelConstruction*))
# 1.13.0
tensorflow::kernel_factory::OpKernelRegistrar::InitInternal( \
  tensorflow::KernelDef const*, absl::string_view, \
  std::unique_ptr<tensorflow::kernel_factory::OpKernelFactory, \
  std::default_delete<tensorflow::kernel_factory::OpKernelFactory> >)
```

That means we do have to release a binary against 1.13.0 even if we use tensorflow's public endpoint Python API.

yongtang:
All tests passes on 1.13.0rc0 except R. One thing to note is that 1.13.0rc0 is not compatible with 1.12.0 anyway so we likely will move forward with 1.13.0 for the next 0.3.0 release.

yongtang:
@dmitrievanthony @BryanCutler  @terrytangyuan This PR is ready for review.

With this PR the tensorflow-io could be build against tensorflow 1.13 (rc0) and the generated binary is compatible with tensorflow 2.0 (preview), as is shown on Travis CI setup.

The code will not be compatible with 1.12 since there are C++ API changes in TF (https://github.com/tensorflow/io/pull/56#issuecomment-461245725).

To build locally, you could install tensorflow 1.13.0rc0 before `./configure.sh`:
```
$ docker run -it -v ${PWD}:/working_dir -w /working_dir tensorflow/tensorflow:custom-op
$ # In docker
$ curl -OL https://github.com/bazelbuild/bazel/releases/download/0.20.0/bazel-0.20.0-installer-linux-x86_64.sh
$ chmod +x bazel-0.20.0-installer-linux-x86_64.sh
$ ./bazel-0.20.0-installer-linux-x86_64.sh
$ pip install tensorflow==1.13.0rc0
$ ./configure.sh
$ bazel build build_pip_pkg
$ bazel-bin/build_pip_pkg artifacts
```

At the moment all CI tests should pass, except for R. Not sure how to fix the R failure.

/cc @terrytangyuan maybe you could take a look at the R part?

terrytangyuan:
The class name has changed. We probably need to update this line: https://github.com/tensorflow/io/blob/master/R-package/R/dataset_utils.R#L10

Try change it to something like: 
```
inherits(x, c("tensorflow.python.data.ops.dataset_ops.Dataset", "tensorflow.compat.v1.data.Dataset")) || is_tfio_dataset(x)
```

BryanCutler:
Thanks for looking into this @yongtang !

terrytangyuan:
@yongtang I doubt that it's related to `tfdatasets` since we are using our own `is_dataset()`. If the fix doesn't work, feel free to have it return `TRUE` and track this in a separate issue. I'll help look into further later. As long as `class(dataset) <- c("tf_dataset", class(dataset))` logic is happening in `as_tf_dataset()`, this should be safe. This function is internal and for package developer only so I don't expect that random objects will be passed into `as_tf_dataset()`.

yongtang:
@terrytangyuan I tried locally with `TRUE` and still it thrown out error. On the other hand, if I install tfdatasets from the GitHub (master branch HEAD):
```
R -e 'library("devtools"); install_github("rstudio/tfdatasets", ref="c6fc59b")'
```

Then it seems the tests are passing. Wondering if this helps.

tensorflow_io/arrow/python/ops/arrow_dataset_ops.py
```diff
@@ -130,8 +135,12 @@ def __init__(self,
       output_shapes: TensorShapes of the output tensors or None to
                      infer partial
     """
+    self._columns = columns
+    self._output_types = output_types
+    self._output_shapes = output_shapes or \
+        nest.map_structure(
+            lambda _: tensorflow.TensorShape(None), self._output_types)
```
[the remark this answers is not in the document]
yongtang: @BryanCutler I just updated the PR and removed unneeded initialization (other than one place which is needed).

terrytangyuan:
@yongtang I see. Then it might be using the is_dataset() from tfdatasets. I’ll double check on this. In the meantime, let’s just use your fix for now.

setup.py
```diff
@@ -21,9 +21,9 @@
 from setuptools import setup
 from setuptools.dist import Distribution
 
-__version__ = '0.2.0'
+__version__ = '0.3.0'
 REQUIRED_PACKAGES = [
-    'tensorflow == 1.12.0',
+    'tensorflow >= 1.13.0rc0',
```
dmitrievanthony: Does it makes sense to add upper bound as well?

dmitrievanthony:
BTW, @yongtang are you planning to port these changes into `tensorflow/contrib`? 

I see that last changes I made to recover `ignite` and `kafka` functionality in `tensorflow` were cherry-picked into `1.13` and so that `1.13.0rc0` works well. Unfortunately, it looks like folks changed underlying API again and our code is broken in `1.13.0rc1`. I could fix this issue ad-hock, but we could also cherry-pick changes from this request into `tensorflow/contrib` also. What will be the best chose, what do you think?

tensorflow_io/arrow/python/ops/arrow_dataset_ops.py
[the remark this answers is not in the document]
BryanCutler: Thanks, looks good!

yongtang:
@dmitrievanthony My understanding is that tensorflow 1.13 will likely be the last release before 2.0. And starting from tensorflow 2.0, the contrib directory will be deprecated and removed from tensorflow repo.

In that case, an ad-hoc fix on `tensorflow/contrib` to r1.13 branch directly (so that some ops are working fine) might be an easier path.

/cc @martinwicke @ewilderj to comment on the timeline of 1.13 and 2.0.

martinwicke:
Sadly, it's very likely we'll have at least one more 1.x release before 2.0 actually lands (outside of pre-releases).

yongtang:
Both TF 1.13.0rc2 and TF 2.0 (tf-nightly-2.0-preview) build with tensorflow-io succeed now. Think we are ready for 1.13 and 2.0.

Note:
1) This PR is on top of another PR #94 
2) The code base is compatible with 1.13 and 2.0. However, 
3) the binary built with TF 1.13.0rc2 is not compatible with TF 2.0 (C++ API issue?), see https://github.com/tensorflow/io/issues/55#issuecomment-464492690

I think the plan will be to rebuild another version of tensorflow-io when 2.0 is released.

yongtang:
Will merge this PR in for now. We want to start testing tf 1.13 now (as tf 1.13 release is imminent).

I will create a follow up PR to change the pip requirement to `tensorflow>=1.13.0, < 1.14.0` once TF 1.13 is released.

PR #63, opened 2019-01-23
Add correct impl of AsGraphDefInternal to Arrow Datasets

BryanCutler:
Added `AsGraphDefInternal` implementations to Arrow Datasets to properly define op outputs and node for serialization. This was causing a cryptic error when testing against TensorFlow v1.13. Also corrected `DebugString` to be a public method.


BryanCutler:
@yongtang , after updating the APIs for v1.13, this change fixes the tests so all are passing. I believe this should be fixed for 1.12 also, so I made this a separate PR.

BryanCutler:
Thanks @yongtang !

Issue #57, opened 2019-01-19
tensorflow-io 0.2.0 release


yongtang:
This issue tracks the efforts to release tensorflow-io 0.2.0.

Since TensorFlow 1.13 has not been released, tensorflow-io 0.2.0 will be based on TensorFlow 1.12.

- [x] Update `README.md` to add additional supported ops since 0.1.0 (@BryanCutler)
- [x] Update `README.md` to add link to individual ops' README.md if possible.  (@BryanCutler)
- [x] Change setup.py to fix `tensorflow==1.12.0`, as `1.13` is not compatible. (#65/@yongtang)
- [x] Add integration test in `tests` so that the build of `*.whl` files work correctly (#64/@yongtang)
- [x] Add examples in `doc/tutorials` so that end user could see the use case (defer to tensorflow 1.13).
- [x] Update R-packages so that additional ops are supported. (#70/ @terrytangyuan)
- [x] Update `RELEASE.md` to capture 0.2.0 changes (#66 / @yongtang)
- [x] Build `*.whl` files and push to PyPI.org  (@BryanCutler)
- [x] Release R packages to CRAN.

If you want to help, you can add a comment to indicate the items you are working on.

BryanCutler:
I could take the first 2 `README.md` updates, and `Build/push `*.whl` files`

> Add examples in doc/tutorials so that end user could see the use case
what kind of format will we make these tutorials, markdowns files also in the github repo or is it possible to have sphinx/html, or something?

yongtang:
Thanks @BryanCutler! I think markdown files with sample code would be a good start really.

BryanCutler:
>could you create an account in PyPI.org and create a PR to add your name and PyPI.org handle to Current Release Team

Done, PR https://github.com/tensorflow/community/pull/59

yongtang:
I think we cover all items. The `doc/tutorials` probably could be deferred to the next release as with tensorflow 1.13 it is possible to combine `tf.data` with `tf.keras` and the examples will be more telling. With 1.12 there are still many bugs for tf.data to work with higher level apis.

yongtang:
Just tased v0.2.0. @BryanCutler Would you like to help build the *.whl files and push to PyPI.org?

The build method for *.whl files have been updated, see PR   tensorflow/community#58
```
$ docker run -it -e BAZEL_VERSION=0.20.0 --rm -v ${PWD}:/working_dir \
    -w /working_dir  tensorflow/tensorflow:custom-op \
[...]

PR #64, opened 2019-01-25
Add an integration test case for VideoDataset

yongtang:
In our build test we haven't had a test case yet
(only tested that `pip install tensorflow_io-*-*-*.whl` works),
this PR add an integration test case for VideoDataset.

This PR is part of #57.

### DIFF
--- a/tests/test_video.py
+++ b/tests/test_video.py
@@ -1,0 +1,56 @@
+# Copyright 2018 The TensorFlow Authors. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may not
+# use this file except in compliance with the License.  You may obtain a copy of
+# the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.  See the
+# License for the specific language governing permissions and limitations under
+# the License.
+# ==============================================================================
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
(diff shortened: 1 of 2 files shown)


tests/test_video.py
```diff
@@ -0,0 +1,56 @@
...
+
+import tensorflow as tf
+from tensorflow.keras.applications.resnet50 import ResNet50
+from tensorflow.keras.applications.resnet50 import preprocess_input, decode_predictions
+from tensorflow_io.video import VideoDataset
+
+video_path = os.path.abspath(
+    os.path.join(os.path.dirname(os.path.abspath(__file__)),
+    "..",
+    "tensorflow_io/video/python/kernel_tests/testdata/small.mp4"))
```
terrytangyuan: Should we move all test data to top level since it's used by top level `tests/` as well?

tests/test_video.py
```diff
@@ -0,0 +1,56 @@
...
+
+  dataset = VideoDataset([video_path]).repeat(num_repeats)
+  iterator = dataset.make_initializable_iterator()
+  init_op = iterator.initializer
+  get_next = iterator.get_next()
+
+  with tf.Session() as sess:
+    sess.run(init_op)
+    for _ in range(num_repeats):  # Dataset is repeated.
+      for _ in range(166): # 166 frames
```
terrytangyuan: You can remove the comment since the code is self-explanatory.

yongtang: Updated. Thanks.

tests/test_video.py
yongtang: The only issue is the bazel as `testdata` directory needs to be visible to the project e.g., `//tensorflow_io/video/BUILD`, haven't found out a successful way to make it work with bazel yet.

I can continue investigation and add a follow up PR once I am able to expose the testdata in bazel.

PR #65, opened 2019-01-25
Use tensorflow == 1.12.0 in setup.py

yongtang:
This PR use fixed value of `tensorflow == 1.12.0` in setup.py,
as from the tests of #56, `tensorflow == 1.13.0` will not work
without additional implementation changes.


terrytangyuan:
@yongtang You can change it here: https://github.com/tensorflow/io/blob/master/R-package/R/package.R#L52-L53

yongtang:
Thanks @terrytangyuan. The PR has been updated. Please take a look.

PR #66, opened 2019-01-26
tensorflow-io 0.2.0 release

yongtang:
This PR is for tensorflow-io 0.2.0 release

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,3 +1,21 @@
+# Release 0.2.0
+
+## Major Features
+* `tensorflow_io.arrow`: Apache Arrow Datasets.
+* `tensorflow_io.image`: WebP image Dataset.
+* `tensorflow_io.libsvm`: LIBSVM Dataset.
+* `tensorflow_io.parquet`: Apache Parquet Dataset.
+* `tensorflow_io.video`: Video Dataset (from FFmpeg).
+
+## Thanks to our Contributors
+
+This release contains contributions from many people:
+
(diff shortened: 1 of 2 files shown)


terrytangyuan:
@yongtang I am working on preparing v0.2.0 release for R package, which includes version bump and some other fixes/enhancements. I’ll open up a PR when it’s ready. Please focus on Python package release for this PR.

yongtang:
@terrytangyuan Thanks. I will put this PR on hold.

terrytangyuan:
I just submitted a PR for v0.1.0 release after receiving CRAN's first round of review. 

This is taking much longer than I originally expected due to new year schedule. I propose that we can go ahead and create release tags on Github whenever the Python package is ready without waiting for R package. CRAN has its own archiving mechanism (for example, here are the archives for [reticulate package](https://cran.r-project.org/src/contrib/Archive/reticulate/)) so I don't think Github releases would affect R users that much.

yongtang:
@terrytangyuan Thanks, that makes sense. We could decouple the release of python and R. We could even have a special tag (e.g., something like`0.1.0-r`) if needed.

yongtang:
I think we are almost there for 0.2.0 release. Only need to update README.md before the tag.

Also cc @BryanCutler

BryanCutler:
I'm working on the README updates now, I'll have a PR soon

yongtang:
/cc @terrytangyuan @BryanCutler Just rebased and updated. I think we should be ready for 0.2.0.

@BryanCutler I also added you to the PyPI.org project so you should have access. Once the PR is merged then we could just tag and release 0.2.0.

PR #70, opened 2019-01-27
Prepare R package v0.2.0 release

terrytangyuan:
This PR prepares R package v0.2.0 release, which includes:
* Added some missing support such as Arrow and LIBSVM. Note that `ArrowDataset.from_pandas()` is intentionally not supported since there's no `pandas` in R. We can think about supporting native R `data.frame` and `data.table::data.table` later on. 
* Bumped version string, updated `NEWS.md` to be consistent with https://github.com/tensorflow/io/pull/66, and updated `cran-comments.md`.

Also note this may not be the final version of 0.2.0 R package since there maybe additional things we need to address from CRAN's feedback. I am planning to submit this release once new Python package release is on PyPI.


BryanCutler:
LGTM, thanks for doing this!  I like the idea of adding support for an R `data.frame` in the future!!

Issue #73, opened 2019-01-30
Migrate LMDBDataset from tensorflow to tensorflow-io


yongtang:
`LMDBDataset` is part of the `tf.contrib.data` and it naturally fits the tensorflow-io context. It makes sense to migrate LMDBDataset to tensorflow-io.

terrytangyuan:
I can work on this one if nobody else is working on it.

yongtang:
Thanks @terrytangyuan that would be great! Let me know if you encounter any issues with LMDBDataset.

yongtang:
This issue is related to tensorflow/datasets#34

terrytangyuan:
@yongtang the support for LMDB is now ready for review: https://github.com/tensorflow/io/pull/74

PR #74, opened 2019-02-01
Add LMDB Support

terrytangyuan:
This PR adds support for [Lightning Memory-Mapped Database (LMDB)](https://en.wikipedia.org/wiki/Lightning_Memory-Mapped_Database). See https://github.com/tensorflow/io/issues/73 for background.


terrytangyuan:
Yes, I agree that we should switch to use public APIs. We can do it altogether with other dataset ops in separate PRs.

PR #76, opened 2019-02-02
Fix missing column index when using ArrowDataset.from_pandas with preserve_index

### LINKED ISSUE #75: ArrowDataset.from_pandas fails when preserve_index=True
When making an ArrowDataset from a pandas.DataFrame, if the `preserve_index` flag is set to `True` the Dataset iterator will fail with the error:
```
InternalError: Missing 2-th output from node IteratorGetNext (defined at <ipython-input-42-624c3010a782>:1)  = IteratorGetNext[output_shapes=[[], [], []], output_types=[DT_DOUBLE, DT_INT64, DT_INT64], _device="/job:localhost/replica:0/task:0/device:CPU:0"](OneShotIterator)
```

This is because the `preserve_index` flag will add an additional column in the record batch and there is not a corresponding column index sent to the op.

BryanCutler:
This change fixes an error caused when ArrowDataset.from_pandas `preserve_index` flag is True. Instead of making column indices based on the DataFrame, they are now made based on the pyarrow.RecordBatch, which includes the index column. Also, the DataFrame columns are now properly selected using `iloc`.  Added regression tests with `preserve_index=True`.

Fixes #75

### DIFF
--- a/tensorflow_io/arrow/python/kernel_tests/arrow_test.py
+++ b/tensorflow_io/arrow/python/kernel_tests/arrow_test.py
@@ -189,6 +189,40 @@ class ArrowDatasetTest(test.TestCase):
     self.run_test_case(dataset, test_data)
 
   @unittest.skipIf(not _have_pyarrow, _pyarrow_requirement_message)
+  def testFromPandasPreserveIndex(self):
+    data = [
+        [1.0, 2.0, 3.0],
+        [0.2, 0.4, 0.8],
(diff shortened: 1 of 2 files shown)


BryanCutler:
Looks like I had missed testing this case.. @yongtang @terrytangyuan please take a look when you can, thanks!

BryanCutler:
Thanks @terrytangyuan !

PR #79, opened 2019-02-05
Add boolean data type support to Arrow Datasets

### LINKED ISSUE #78: Add boolean type support to Arrow Datasets
Arrow datasets are missing support for boolean data types

BryanCutler:
This change adds boolean data type support to Arrow based Datasets, both for scalar and vector tensors.  Additionally, checking Arrow batch column type matches the expected tensor output type has been moved from checking at each iteration, to only once per batch stream. Checking once per stream is all that is required since a set schema applies to all batches within that stream.

Fixes #78

### DIFF
--- a/tensorflow_io/arrow/python/kernel_tests/arrow_test.py
+++ b/tensorflow_io/arrow/python/kernel_tests/arrow_test.py
@@ -48,6 +48,7 @@ class ArrowDatasetTest(test.TestCase):
   def setUpClass(cls):
 
     cls.scalar_data = [
+        [True, False, True, True],
         [1, 2, -3, 4],
         [1, 2, -3, 4],
         [1, 2, -3, 4],
@@ -60,6 +61,7 @@ class ArrowDatasetTest(test.TestCase):
         [1.1, 2.2, 3.3, 4.4],
     ]
     cls.scalar_dtypes = (
+        dtypes.bool,
         dtypes.int8,
         dtypes.int16,
         dtypes.int32,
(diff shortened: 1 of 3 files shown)


BryanCutler:
@yongtang @terrytangyuan please take a look when you have time, thank you!

tensorflow_io/arrow/python/kernel_tests/arrow_test.py
```diff
@@ -305,6 +308,40 @@ def run_server(num_batches):
 
     server.join()
 
+  @unittest.skipIf(not _have_pyarrow, _pyarrow_requirement_message)
+  def testBoolArrayType(self):
+
+    # NOTE: need to test this seperately because to_pandas fails with
+    # ArrowNotImplementedError: Not implemented type for list in
```
xhochy: Is this a known issue?

BryanCutler: @xhochy  I'm not sure, I just noticed it with this pr. I'll look into it and make a jira if one doesn't exist already.

BryanCutler: Found the issue https://issues.apache.org/jira/browse/ARROW-4370

xhochy: @BryanCutler Thanks for digging into this!

terrytangyuan: Could you add the link to the issue in the comment?

BryanCutler: yup, good idea

BryanCutler:
Thanks for reviewing @terrytangyuan !

PR #84, opened 2019-02-11
Travis CI improvement

yongtang:
Travis CI build is getting longer and longer, so we have to think about improvement with various methods.

PyPI.org's nightly build is also setup.


yongtang:
This PR is ready for review. Several enhancement has been made:
1) The PR splits builds and tests process, so that we have more room in Travis CI (50 min limit) with additional testing.
2) All future PR will first generate 4 "nightly" build in test.pypi.org with python 2.7, 3.4, 3.5, 3.6 first. Only after all tests (Python & R) passed, then an true "nightly" binary will be released.
https://travis-ci.org/tensorflow/io/builds/491754340

In other words, we are using `test.pypi.org` as a temporary storage for Travis CI. Travis CI does not comes with a sharable storage by default (unless we use paid S3/GCS).

There are quite some code duplications. I tried over the weekend to play with Travis CI. However, it turns out Travis CI's "Matrix" expansion have several bugs that will not correct render the right "matrix" as I expect. So this is for now. In the future, we may have additional enhancement to reduce the code duplication in Travis CI setup.

Finally, our nightly build is out and is part of the Travis CI process automatically:
https://pypi.org/project/tensorflow-io-nightly/

/cc @terrytangyuan @dmitrievanthony @BryanCutler

yongtang:
Once this PR is merged, we should be able to move to 1.13.0 build.

yongtang:
Thanks @BryanCutler. The PR should have been fixed. The PR build number was used on purpose, as we need to have a way to reliably find out the version to test for different jobs within the same build (the build number is the same across different jobs).

Here are some background about Travis CI:
1) Travis CI limit each job to 50 mins. We need to build 2.7, 3.4, 3.5, 3.6 which takes more than one hour to just build. And we haven't take into consideration of 3.7 yet.
2) So we are forced to split the build with smaller jobs. Now 2.7, 3.4, 3.5, 3.6 are build on 4 jobs.
3) Even we split the builds, we still need to do testing. And testing has to be done in different jobs as well.
4) Unfortunately, Travis CI does not provide a way to share storage across jobs so we could not use the build in one job for test in another jobs.
5) Eventually, I just use test.pypi.org as a temporarily (free) storage so that every job within the same build could access (a timestamp will not work as we don't know which version to pull).

This is kinds of not optimal. If we want to avoid the situation above we could:
1) Use a CI provider that have a execution limit longer than 50min. I played with CircleCI and AzurePipeline during the weekend and had a couple of test runs with them, but
  - CircleCI is nice and convenience, but it comes with a limit of 1000 min per month (my understanding).
  - AzurePipeline is also nice and powerful, but it asks my write permission to write the file to GitHub repo directly (instead of me pushing and it pulling). I am not convinced that they need write permission to do that.

2) So for now we may need to stick with Travis CI. In order to solve the issue we need a free storage so that build binaries could be shared. AWS and GCS come with cost. Looks like test.pytest.org could be a not-so-ideal-but-works solution.

Some other notes: test.pytest.org does not allow you to overwrite already uploaded files. Files could be deleted, but version will remain even if deleted.

For now, I think the solution so far might be enough for the use. In the future I guess the whole CI could be revisited to see if better solution could be used.

yongtang:
Also, there was one build failure in Python ubuntu 16.04 test. The failure is from `more_itertools` which released `6.0.0` a couple of hours ago that breaks `pytest` for python 2.7:

```
...
  File "/usr/local/lib/python2.7/dist-packages/_pytest/fixtures.py", line 16, in <module>
    from more_itertools import flatten
  File "/usr/local/lib/python2.7/dist-packages/more_itertools/__init__.py", line 1, in <module>
    from more_itertools.more import *  # noqa
  File "/usr/local/lib/python2.7/dist-packages/more_itertools/more.py", line 329
    def _collate(*iterables, key=lambda a: a, reverse=False):
                               ^
SyntaxError: invalid syntax

```

I have pinned the `more-itertools==5.0.0` so all tests should pass now.

PR #88, opened 2019-02-13
Use localstack:localstack emulator to test kinesis

yongtang:
The test for Kinesis implementations used to be done manually and this is very error prune and out-of-sync with code changes.

This commit utilize the localstack:localstack emulator which is a container that emulates AWS services (not just limited to Kinesis but also many other services).

This commit changes the test so that Kinesis code changes are well covered.


yongtang:
Thanks @terrytangyuan. Once this PR is merged, I think we have enough new features to release 0.3.0.

PR #89, opened 2019-02-13
Add `pip install pyarrow==0.11.1` back to Travis CI

yongtang:
Just realized that recent Travis CI enhancement to add
the nightly build actually removes the
`pip install pyarrow==0.11.1` line in the middle.
Need to add this line back as otherwise arrow test
may not be invoked correctly.


yongtang:
@BryanCutler this PR is to reenable arrow test. During the refactor of Travis CI the `pip install pyarrow`  was missed, which results in arrow tests not run at all. I made some changes to arrow tests so that pyarrow is always required unless python==3.4. This will help prevent code changes untested in the future. Please take a look.

PR #90, opened 2019-02-13
Quick fix: Re-enable python 2.7 ad 3.4 build/test on Ubuntu 14.04

yongtang:
Looks like the early fix actually missed the test case of
Python 2.7 + Ubuntu 14.04
Python 3.4 + Ubuntu 14.04

This is a quick fix to re-enable it.


yongtang:
Thanks @terrytangyuan, added an issue #92 to track the Travis CI enhancement.

PR #100, opened 2019-02-18
Fix PyTest warning due to class TestData

BryanCutler:
This change renames class `TestData` in `test_arrow.py` to prevent the below warning from PyTest during test discovery because the name starts with "Test".

`/working_dir/tests/test_arrow.py:0: PytestWarning: cannot collect test class 'TestData' because it has a __new__ constructor`


BryanCutler:
Thanks @terrytangyuan !

PR #101, opened 2019-02-19
Add write_kafka ops for streaming output

yongtang:
This PR adds a very early implementation of write_kafka ops
so that we could do a streaming output.

The basic idea of write_kafka is to implement in C++ WriteKafkaOp
which takes a message as input, write the message to a Kafka server,
and returns the original message. The return of original message
is on purpose so that write_kafka is not a terminal op.
This helps to do further processing.

So far most of the tensorflow-io ops are input, while the output
has been mentioned in discussions. This implementation could be
a start for future implementation of output ops.

### DIFF
--- a/tensorflow_io/kafka/kernels/kafka_dataset_ops.cc
+++ b/tensorflow_io/kafka/kernels/kafka_dataset_ops.cc
@@ -317,4 +317,84 @@ class KafkaDatasetOp : public DatasetOpKernel {
 REGISTER_KERNEL_BUILDER(Name("KafkaDataset").Device(DEVICE_CPU),
                         KafkaDatasetOp);
 
+class WriteKafkaOp : public OpKernel {
+ public:
+  using OpKernel::OpKernel;
+  void Compute(OpKernelContext* context) override {
+    const Tensor* message_tensor;
+    const Tensor* topic_tensor;
+    const Tensor* servers_tensor;
+    OP_REQUIRES_OK(context, context->input("message", &message_tensor));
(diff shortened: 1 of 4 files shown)


yongtang:
Note in TensorFlow there is a `tf.write_file` op. But the `tf.write_file` does not play well with out datasets, as `tf.write_file` override the same file content.

Instead, I think at some point we could rewrite `write_file` to allow append to a binary file, or append record to a structured file (like video, etc). That will be more useful to combine with higher level APIs such as Keras I think.

tensorflow_io/kafka/kernels/kafka_dataset_ops.cc
```diff
@@ -317,4 +317,84 @@ class KafkaDatasetOp : public DatasetOpKernel {
...
+    OP_REQUIRES(context, (err == RdKafka::ERR_NO_ERROR),
+        errors::Internal("Failed to produce message:", RdKafka::err2str(err)));
+
+    err = producer->flush(timeout_);
+    OP_REQUIRES(context, (err == RdKafka::ERR_NO_ERROR),
+        errors::Internal("Failed to flush message:", RdKafka::err2str(err)));
+    context->set_output(0, context->input(0));
+  }
+private:
+  static const int timeout_ = 5000;
```
BryanCutler: Would it be better if this was configurable?

yongtang:
Thanks for the review. Let's merge this one for now. I will create a follow up PR to address additional issues in the next day or two 👍

terrytangyuan:
Great work! @yongtang

PR #102, opened 2019-02-19
Change _load_library to get datapath from environment variable for tests

BryanCutler:
This PR uses an environment variable `TFIO_DATAPATH` to specify an alternative path to load op libraries from, replacing the pytest option `--binary <path>`. Tests can now be run in the following way after building the C++ ops with Bazel:
```
TFIO_DATAPATH=bazel-bin pytest tests
```


BryanCutler:
I think we can specify the shared library path for tests using an environment variable, that way we don't have to pass the path around as an argument to pytest. The CI scripts would still to be updated if we want to use this. What do you guys think @yongtang and @terrytangyuan ?

Also, adding the `tests/__init__.py` makes pytest treat it as a package and I could then run tests by with just calling pytest, for example `TFIO_DATAPATH=bazel-bin pytest tests/test_arrow.py`

yongtang:
@BryanCutler I think it makes sense. Can you change the Travis CI script to capture the change so that test can pass? The file is located in `.travis/build.test.sh`

BryanCutler:
Sure, will do. Just wanted to check first if it sounded ok.

setup.py
```diff
@@ -118,11 +118,16 @@ def has_ext_modules(self):
...
   f.write(content.format(version, project))
 
+datapath = None
 if '--data' in sys.argv:
   data_idx = sys.argv.index('--data')
   datapath = sys.argv[data_idx + 1]
   sys.argv.remove('--data')
   sys.argv.pop(data_idx)
+else:
+  datapath = os.environ.get('TFIO_DATAPATH')
```
BryanCutler: @yongtang I added this as an alternative to `--data` option for when we might run `setup.py` manually and have the env var set. I didn't change the CI script to use this because the `--data` seems fine.

.travis/install.test.sh
```diff
@@ -33,11 +33,11 @@ R -e 'install.packages(c("forge"), quiet = TRUE)'
 CPYTHON_VERSION=$(python3 -c 'import sys; print(str(sys.version_info[0])+str(sys.version_info[1]))')
 python3 -m pip install wheelhouse/*-cp${CPYTHON_VERSION}-*.whl
 python3 -m pip install -q pytest boto3 pyarrow==0.11.1 pandas==0.19.2
-(cd tests && python3 -m pytest .)
+(cd tests && python3 -m pytest --import-mode=append .)
```
BryanCutler: Previous CI install.test job failed with:
```
_____________________ ERROR collecting tests/test_arrow.py _____________________
test_arrow.py:35: in <module>
    from tensorflow_io.arrow.python.ops import arrow_dataset_ops
/v/tensorflow_io/arrow/__init__.py:24: in <module>
    from tensorflow_io.arrow.python.ops.arrow_dataset_ops import ArrowDataset
/v/tensorflow_io/arrow/python/ops/arrow_dataset_ops.py:30: in <module>
    arrow_ops = _load_library('_arrow_ops.so')
/v/tensorflow_io/__init__.py:54: in _load_library
    raise NotImplementedError("unable to open file: ", filename)
E   NotImplementedError: ('unable to open file: ', '_arrow_ops.so')
```
So it seems to be picking up tensorflow_io from the source directory and not the installation. I could reproduce this locally and adding the option `--import-mode=append`. I'm not sure if this had anything to do with my changes or not.

.travis/python.release.sh
```diff
@@ -40,8 +40,12 @@ DEBIAN_FRONTEND=noninteractive apt-get -y -qq install \
 
 curl -sOL https://github.com/bazelbuild/bazel/releases/download/${BAZEL_VERSION}/bazel-${BAZEL_VERSION}-installer-${BAZEL_OS}-x86_64.sh
 chmod +x bazel-${BAZEL_VERSION}-installer-${BAZEL_OS}-x86_64.sh
-./bazel-${BAZEL_VERSION}-installer-${BAZEL_OS}-x86_64.sh
+
+# Install bazel, display log only if error
+./bazel-${BAZEL_VERSION}-installer-${BAZEL_OS}-x86_64.sh 2>&1 bazel_install.log || (cat bazel_install.log && false)
```
BryanCutler: The bazel install log is pretty long, this should only print the log if there was an error.

.travis/install.test.sh
BryanCutler: Ahh, figured it out. It was the addition of `tests/__init__.py`.

.travis/build.test.sh
```diff
@@ -28,8 +31,11 @@ DEBIAN_FRONTEND=noninteractive apt-get -y -qq install \
 
 curl -sOL https://github.com/bazelbuild/bazel/releases/download/${BAZEL_VERSION}/bazel-${BAZEL_VERSION}-installer-${BAZEL_OS}-x86_64.sh
 chmod +x bazel-${BAZEL_VERSION}-installer-${BAZEL_OS}-x86_64.sh
-./bazel-${BAZEL_VERSION}-installer-${BAZEL_OS}-x86_64.sh
+
+# Install bazel, display log only if error
+./bazel-${BAZEL_VERSION}-installer-${BAZEL_OS}-x86_64.sh 2>&1 > bazel_install.log || (cat bazel_install.log && false)
```
BryanCutler: The bazel install log is pretty long and I don't think we need it if install went fine, this should only print the log if there was an error. Does that seem ok?

.travis/python.release.sh
```diff
@@ -40,8 +40,12 @@ DEBIAN_FRONTEND=noninteractive apt-get -y -qq install \
 
 curl -sOL https://github.com/bazelbuild/bazel/releases/download/${BAZEL_VERSION}/bazel-${BAZEL_VERSION}-installer-${BAZEL_OS}-x86_64.sh
 chmod +x bazel-${BAZEL_VERSION}-installer-${BAZEL_OS}-x86_64.sh
-./bazel-${BAZEL_VERSION}-installer-${BAZEL_OS}-x86_64.sh
+
+# Install bazel, display log only if error
+./bazel-${BAZEL_VERSION}-installer-${BAZEL_OS}-x86_64.sh 2>&1 > bazel-install.log || (cat bazel-install.log && false)
```
BryanCutler: The bazel install log is pretty long and I don't think we need it if install went fine, this should only print the log if there was an error. Does that seem ok?

BryanCutler:
@yongtang the tests are passing now if you could please have another look, thanks!

setup.py
yongtang: @BryanCutler The change in setup.py is not needed. This file is only used for packaging the whl file. In the process of packaging, it always needs a source file and the .so file. At this stage, .so file is always in a separate directory than source (bazel). This setup.py file is also not part of the whl itself so it will not impact the behavior of `__init__.py`.

.travis/python.release.sh
yongtang: Looks good! 👍

yongtang:
@BryanCutler I think the change in setup.py is not needed, but otherwise it looks good.

setup.py
BryanCutler: I added this as a convenience if using the Docker image from #104. It sets the env var for the container, so if I want to build a wheel for my own testing I can use the command `python setup.py bdist_wheel`. It's just to make it slightly easier for the developer and the `--data` arg still takes precedence here. Would you still like me to remove it?

yongtang: OK. I think we could leave it alone.

yongtang:
@BryanCutler may need a rebase then we could merge I think.

yongtang:
The test failure:
```
Error: HTTP error 403.
  API rate limit exceeded for 35.192.187.174. (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.)
```
is related to #105 and has been fixed in master. So merging..

Thanks for the PR 👍 ❤️ !

BryanCutler:
Thanks @yongtang !

PR #104, opened 2019-02-20
Add Dockerfile for Python Development

### LINKED ISSUE #103: Make Dockerfile for Python Development
It would be nice to have a Dockerfile to build an image with all the configuration and packages for a developer to start building and testing right away. This would include:

* Required OS packages
* Bazel installation
* TensorFlow
* Required pip packages

BryanCutler:
This adds a docker file to be used for Python development. It includes required OS packages, Bazel installation, and setups up a conda environment `tfio-dev` with required pip packages. It also sets the environment variable `TFIO_DATAPATH` used for loading shared libraries needed to run PyTests.

The Docker image can be built with the command from `tensorflow/io` directory:
`docker build -f dev/Dockerfile -t tfio-dev .`

To run the container:
`docker run -it --rm -v $PWD:/v -w /v tfio-dev`

In the container, do the following:

To setup and install TensorFlow, if not installed already
`./configure.sh`

To build TensorFlow I/O C++ 
`bazel build -s --verbose_failures --spawn_strategy standalone //tensorflow_io/...`

To run tests
`pytest tests`

To build a distributable Python package
`python setup.py bdist_wheel`

Fixes #103

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,20 +77,40 @@ devtools::install_github("tensorflow/io", subdir = "R-package")
 
 ### Python
 
-The TensorFlow I/O package (`tensorflow-io`) could be built from source:
+For Python development, a reference Dockerfile [here](dev/Dockerfile) can be
+used to build the TensorFlow I/O package (`tensorflow-io`) from source:
 ```sh
-$ docker run -it -v ${PWD}:/working_dir -w /working_dir tensorflow/tensorflow:custom-op
-$ # In docker
-$ curl -OL https://github.com/bazelbuild/bazel/releases/download/0.20.0/bazel-0.20.0-installer-linux-x86_64.sh
-$ chmod +x bazel-0.20.0-installer-linux-x86_64.sh
-$ ./bazel-0.20.0-installer-linux-x86_64.sh
-
+$ # Build and run the Docker image
+$ docker build -f dev/Dockerfile -t tfio-dev .
+$ docker run -it -rm --net=host -v ${PWD}:/v -w /v tfio-dev
+$ # In Docker, configure will install TensorFlow or use existing install
 $ ./configure.sh
-$ bazel build //tensorflow_io/...
-$ python setup.py --data bazel-bin -q bdist_wheel
+$ # Build TensorFlow I/O C++
+$ bazel build -s --verbose_failures //tensorflow_io/...
+$ # Run tests with PyTest, note: some tests require launching additional containers to run (see below)
+$ pytest tests/
+$ # Build the TensorFlow I/O package
+$ python setup.py bdist_wheel
 ```
 
 A package file `dist/tensorflow_io-*.whl` will be generated after a build is successful.
+
+NOTE: When working in the Python development container, an environment variable
+`TFIO_DATAPATH` is automatically set to point tensorflow-io to the shared C++
+libraries built by Bazel to run `pytest` and build the `bdist_wheel`. Python
+`setup.py` can also accept `--data [path]` as an argument, for example
+`python setup.py --data bazel-bin bdist_wheel`.
+
+#### Starting Test Containers
+
+Some tests require launching a test container before running. In order
+to run all tests, execute the following commands:
+
+```
+$ bash -x -e tests/test_ignite/start_ignite.sh
+$ bash -x -e tests/test_kafka/kafka_test.sh start kafka
+$ bash -x -e tests/test_kinesis/kinesis_test.sh start kinesis
+```
 
 ### R
 
--- a/dev/Dockerfile
+++ b/dev/Dockerfile
@@ -1,0 +1,46 @@
+FROM ubuntu:18.04
+
+RUN apt-get update && \
+    apt-get install -y \
+    build-essential \
+    curl \
+    nano \
+    unzip \
+    ffmpeg
+
+ARG BAZEL_VERSION=0.20.0
+ARG BAZEL_OS=linux
+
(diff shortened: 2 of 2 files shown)


BryanCutler:
This is a WIP, still need to update README.md and the env var TFIO_DATAPATH depends on #102

README.md
```diff
@@ -71,19 +71,23 @@ devtools::install_github("tensorflow/io", subdir = "R-package")
...
+$ # Build and run the Docker image
+$ docker build -f dev/Dockerfile -t tfio-dev .
+$ docker run -it -rm -v ${PWD}:/v -w /v tfio-dev
+$ # In Docker, configure will install TensorFlow or use existing install
 $ ./configure.sh
-$ bazel build build_pip_pkg
-$ bazel-bin/build_pip_pkg artifacts
+$ # Build TensorFlow I/O C++
+$ bazel build -s --verbose_failures --spawn_strategy standalone //tensorflow_io/...
+$ # Run tests with PyTest, note: some tests might require starting services to run
```
BryanCutler: I actually had some trouble running ignite and kafka tests, I tried running the provided containers but couldn't connect to them. Do I need to forward some ports? Just wondering how you all run these..

BryanCutler:
I tested this by wtih:
```
docker build --build-arg CONDA_ADD_PACKAGES="ipython" -f dev/Dockerfile -t tfio-dev .
docker run -it --rm -v $PWD:/v -w /v tfio-dev
```

Then in docker:
```
pip install tf-nightly
./configure.sh
bazel build -s --verbose_failures --spawn_strategy standalone //tensorflow_io/...
pytest tests/test_arrow.py
python setup.py bdist_wheel
```

BryanCutler:
I added a conda env to the docker image, with Python 3.6 as default. I think this will make it easier to test out different combinations of Python and TensorFlow versions.  Please take a look when you can @yongtang @terrytangyuan , thanks!

README.md
yongtang: @BryanCutler You can pass `docker run -it -rm --net=host -v ${PWD}:/v -w /v tfio-dev` to get access to host port from within docker container.

yongtang:
@BryanCutler Normally to test every tests:
* Launch needed test containers:
  ```
  bash -x -e tests/test_ignite/start_ignite.sh
  bash -x -e tests/test_kafka/kafka_test.sh start kafka
  bash -x -e tests/test_kinesis/kinesis_test.sh start kinesis
  ```
* Run test with `docker run --net=host...`

dev/Dockerfile
```diff
@@ -0,0 +1,45 @@
...
+    echo "source activate tfio-dev" >> ~/.bashrc
+
+ARG PIP_ADD_PACKAGES=""
+
+RUN /bin/bash -c "source activate tfio-dev && python -m pip install \
+    pytest \
+    boto3 \
+    pyarrow==${ARROW_VERSION} \
+    pandas \
+    ${PIP_ADD_PACKAGES} \
```
terrytangyuan: We should probably put all of the dependencies with versions in a single `requirements.txt` file. Currently they also appear in Travis scripts.

README.md
BryanCutler: Ah, thanks I forgot the `--net` option. I'll update with the info

dev/Dockerfile
BryanCutler: So pyarrow is only an optional dependency to enable some functionality and the version doesn't need to be the same as Arrow C++ as long there remains binary compatibility in the format, which hasn't changed since 0.8 I believe.

Are there other hard requirements from other datasets?

yongtang: The `tensorflow-io` package itself only needs `tensorflow` package.

To test `tensorflow-io` (in `tests` directory) `pytest`, `pyarrow`, `boto3` , `pandas` (in python) and `ffmpeg` in Ubuntu are needed for now I think. Might be some others in the future.

BryanCutler: Should `apt-get install ffmpeg` be added to the list of packages here?

yongtang: Yes that should be enough.

BryanCutler:
Ok, I think this is ready. Let me know what you think @yongtang @terrytangyuan , thanks!

README.md
```diff
@@ -71,19 +71,34 @@ devtools::install_github("tensorflow/io", subdir = "R-package")
...
+$ # In Docker, configure will install TensorFlow or use existing install
 $ ./configure.sh
-$ bazel build build_pip_pkg
-$ bazel-bin/build_pip_pkg artifacts
+$ # Build TensorFlow I/O C++
+$ bazel build -s --verbose_failures //tensorflow_io/...
+$ # Run tests with PyTest, note: some tests require launching additional containers to run (see below)
+$ pytest tests/
+$ # Build the TensorFlow I/O package
+$ python setup.py bdist_wheel
```
yongtang: I think `python setup.py --data bazel-bin bdist_wheel` might be needed. Otherwise the PR looks good to me.

BryanCutler: Thanks @yongtang for the reveiw!  Actually, the `--data bazel-bin` option is not needed when working in the container since the env var `TFIO_DATAPATH` is set. This should be added as a note though.

BryanCutler:
@yongtang I rebased and added a note about `TFIO_DATAPATH` / `--data bazel-bin` usage. lmkwyt, thanks!

BryanCutler:
Thanks @yongtang !

PR #105, opened 2019-02-20
Add GITHUB_PAT to pass a token, to have a higher limit to pull from R repo

yongtang:
Travis CI set a limit to pull unless a GITHUB_PAT is passed. With
GITHUB_PAT we should have 5000/hour (vs. 60/hour).


yongtang:
cc @terrytangyuan, pulling R from GitHub is subject to GitHub's rate limiting. So added `GITHUB_PAT` to increase the limit to 5000/hour vs 60/hour.

PR #106, opened 2019-02-20
Enable TF 2.0 preview build

yongtang:
This PR enabled TF 2.0 preview build, as part of the effort for #55.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -111,7 +111,7 @@ if '--preview' in sys.argv:
   preview_idx = sys.argv.index('--preview')
   version = version + ".dev" + sys.argv[preview_idx + 1]
   package = 'tf-nightly-2.0-preview'
-  project = 'tensorflow-io-preview'
+  project = 'tensorflow-io-2.0-preview'
   sys.argv.remove('--preview')
   sys.argv.pop(preview_idx)
 
--- a/setup.py
+++ b/setup.py
@@ -52,7 +52,7 @@ from setuptools import setup
 from setuptools.dist import Distribution
 
 REQUIRED_PACKAGES = [
-    'tensorflow == 1.13.0rc2',
+    '{}',
 ]
 __version__ = '{}'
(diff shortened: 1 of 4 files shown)


yongtang:
/cc @terrytangyuan to take a look. The PR essentially allows us to build tensorflow-io with TF 2.0 (tf-nightly-2.0-preview). The whl package will be uploaded to tensorflow-io-preview once the is merged into the master branch.

There are several changes:
1) It looks like R is not working with TF 2.0 yet so I temporarily disabled it in case 2.0-preview is build.
2) I passed a token explicitly for rate limiting. This token has no access right for anything, only for rate limiting.

I realized that while we could pass a hidden environment variable (Like in my previous PR),  it will not be enabled in Travis CI if the PR is from a forked repo. That means hidden env will not work for contributors. We could always regenerate the token if needed, only need to make sure there is no access right for anything.

The same approach has been used in at least mini shift so I think we should be fine:
https://github.com/minishift/minishift/blob/master/.travis.yml

Once this PR is merged, we will have the 2.0 support.

setup.py
```diff
@@ -105,6 +107,14 @@ def has_ext_modules(self):
   sys.argv.remove('--nightly')
   sys.argv.pop(nightly_idx)
 
+if '--preview' in sys.argv:
+  preview_idx = sys.argv.index('--preview')
+  version = version + ".dev" + sys.argv[preview_idx + 1]
+  package = 'tf-nightly-2.0-preview'
+  project = 'tensorflow-io-preview'
```
terrytangyuan: Should we rename this to be `tensorflow-io-2.0-preview` to be consistent with [tensorflow-estimator-2.0-preview](https://pypi.org/project/tensorflow-estimator-2.0-preview/)?

yongtang: @terrytangyuan Updated and changed to `tensorflow-io-2.0-preview`.

PR #110, opened 2019-02-22
Improve error message when _load_library fails

BryanCutler:
When _load_library fails due to a problem loading the shared library, the error message can be misleading because it seems like it can not find the file. This change includes the cause of the load failure and the complete paths searched.

### DIFF
--- a/tensorflow_io/__init__.py
+++ b/tensorflow_io/__init__.py
@@ -44,11 +44,13 @@ def _load_library(filename, lib="op"):
       else lambda f: tensorflow.compat.v1.load_file_system_library(f) == None
 
   # Try to load all paths for file, fail if none succeed
+  errs = []
   for f in filenames:
     try:
       l = load_fn(f)
       if l is not None:
         return l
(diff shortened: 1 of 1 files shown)


BryanCutler:
cc @yongtang @terrytangyuan

BryanCutler:
FYI, here is what the error will look like
```
NotImplementedError: unable to open file: _arrow_ops.so, from paths: 
  ['/miniconda/envs/tfio-dev/lib/python3.6/site-packages/tensorflow_io/arrow/python/ops/_arrow_ops.so', 'bazel-bin/tensorflow_io/arrow/python/ops/_arrow_ops.so']
caused by: ['/miniconda/envs/tfio-dev/lib/python3.6/site-packages/tensorflow_io/arrow/python/ops/_arrow_ops.so: file too short', 
  'bazel-bin/tensorflow_io/arrow/python/ops/_arrow_ops.so: cannot open shared object file: No such file or directory']
```

BryanCutler:
Thanks @terrytangyuan !

PR #112, opened 2019-02-23
[WIP] Use buildpack-deps as the base image to speed up Travis CI

yongtang:
The official buildpack-deps has build tools installed already,
and are based on different os Ubuntu:14.04/16.04/18.04
so speed up should be possible.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,10 @@ git:
 services:
 - docker
 
+cache:
+  directories:
+  - .cache/pip
+
 addons:
   apt:
     update: true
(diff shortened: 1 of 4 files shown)


terrytangyuan:
How much speedup are we seeing here?

yongtang:
It's actually not much, around 2 minutes per job (total around 10 min). I am a little concerned with the addition of new ops. For example, PubSub will need to launch another emulator for setup and that is about 10 more minutes. And we are approaching the 50 minutes limit quickly again.

Let's hold on for this PR for now. I have some thought on additional optimization, will play with it first (together with Mac build).

yongtang:
This PR is mostly in #116 now so closing.

PR #116, opened 2019-02-24
Enable MacOS build on Travis CI

yongtang:
- Enabled macOS build for all ops (including kafka/kinesis/ignite/video)
- Enable macOS tests for all ops except video.
- Updated the script to launch ignite process, kafka/zookeeper process, kinesis localstack emulator on macOS (not through container as Travis does not have docker available on macOS).
- Changed port number from 42300 to 10800 for ignite, as on macOS there is no docker available so no port mapping. 
- Updated librdkafka to 0.11.6 (from 0.11.5)
- Changed the required tensorflow package to `>=1.13.0,<1.14.0`. Also, there is only one place to specify tensorflow package version and that is in setup.py. No multiple places like before to avoid inconsistency.
- Additional refactor of .travis.yaml to consolidate macOS and linux build. Note `.travis/python.release.sh` is still the command to run to build the release binary like before.


yongtang:
There are still some work to do and clean up, but I think we are very close to get macOS support. 

Right now the following ops has been enabled:
- arrow
- hadoop
- image
- libsvm
- lmdb
- parquet

Not done yet:
- ignite
- kafka
- kinesis
- video

For ignite, kafka, kinesis, we will need to setup servers or emulators (with or without docker) for testing on Travis CI.
For video, on macOS there is no standard install like ubuntu for FFmpeg. We have to stick with one or two versions of ffmpeg install (cannot support all).

Will do some cleanup but overall I think we may have a nightly build of macOS support pretty soon.

@damienpontifex as you might be interested.

yongtang:
/cc @terrytangyuan @dmitrievanthony @BryanCutler to take a look. We have enabled the build all ops for macOS, and the tests of most of the ops for macOS. Tests for kinesis/kafka/video/ignite are not done yet.

The PR also adds some enhancement and additional comments about each configuration in Travis CI:
- Developer Build: The builds are done in Ubuntu 16.04/18.04 to make sure anyone could clone the source code and build locally.
- Install Build: The whl binaries are generated with Ubuntu 14.04 (to match tensorflow), and tested with Ubuntu 16.04/18.04. This is to make sure that user could use pip to install the releases and run locally (with their ubuntu 16.04/18.04 machines)
- Preview Build: Nightly releases for TF 2.0 preview.
- Nightly Build: Nightly releases based on TF 1.13.1.

Please take a look

.travis/python.release.sh
```diff
@@ -16,63 +16,17 @@
...
-rm -rf get-pip.py
-python3 -m pip install -q auditwheel==1.5.0
-python3 -m pip install -q wheel==0.31.1
-
-if [[ ! -z ${TENSORFLOW_INSTALL} ]]; then
-  python -m pip install -q ${TENSORFLOW_INSTALL}
+# docker run -i -t --rm -v $PWD:/v -w /v --net=host buildpack-deps:14.04 /v/.travis/python.release.sh
+
+export TENSORFLOW_INSTALL="$(python setup.py --package-version)"
+export PYTHON_VERSION="python python3 python3.5 python3.6"
```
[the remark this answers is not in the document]
yongtang: @BryanCutler The build of whl that is pushed to PyPi.org, has to run on Ubuntu 14.04 as otherwise the `pip install tensorflow-io` will not work with `pip install tensorflow'.

If we build whl in Ubuntu 18.04, then some users that is using an old version of Ubuntu (16.04), will find out that they could run tensorflow but not tensorflow-io. This will not be a good experience for them.

So we stick with Ubuntu 14.04 for building whl. And the generated binary will be tested on Ubuntu 16.04/18.04.

On the other hand, some developers may not be interested in building whl themselves. They only want to build the repo and run locally, maybe debug something. They likely will not use Ubuntu 14.04 (too old). That is why we also test building the source with Ubuntu 16.04/18.04.

BryanCutler:
Ahh, ok makes sense thanks for the explanation. I didn't see that there
were three versions of Ubuntu, I thought it was all 16.04/18.04. I see why
we need to build on 14.04 then, sound fine to me.

yongtang:
@BryanCutler The Ubuntu 14.04 will be out of support (EOL) 04/2019, so likely TensorFlow will start building on Ubuntu 16.04. Then we could drop building on 14.04.

There are also discussions of moving the build of TensorFlow on CentOS 6, for the purpose of manylinux2010. Don't know if this will be the direction or not yet. Though we could always build tenworflow-io with the same environment as tensorflow to make sure it works for any scenario.

yongtang:
Looks like we are all good now and all tests passes on macOS (except for video/ffmpeg). Here is the list of changes in the PR:
- Enabled macOS build for all ops (including kafka/kinesis/ignite/video)
- Enable macOS tests for all ops except video.
- Updated the script to launch ignite process, kafka/zookeeper process, kinesis localstack emulator on macOS (not through container as Travis does not have docker available on macOS).
- Changed port number from 42300 to 10800 for ignite, as on macOS there is no docker available so no port mapping. 
- Updated librdkafka to 0.11.6 (from 0.11.5)
- Changed the required tensorflow package to `>=1.13.0,<1.14.0`. Also, there is only one place to specify tensorflow package version and that is in setup.py. No multiple places like before to avoid inconsistency.
- Additional refactor of .travis.yaml to consolidate macOS and linux build. Note `.travis/python.release.sh` is still the command to run to build the release binary like before.

tests/test_ignite/start_ignite.sh
```diff
@@ -17,13 +17,25 @@
...
+docker run -itd --name ignite-plain -p 10800:10800 \
 -v ${SCRIPT_PATH}:/data apacheignite/ignite:${IGNITE_VERSION} /data/bin/start-plain.sh
 
 # Start Apache Ignite with IGFS.
 docker run -itd --name ignite-igfs -p 10500:10500 \
 -v ${SCRIPT_PATH}:/data apacheignite/ignite:${IGNITE_VERSION} /data/bin/start-igfs.sh
 
 # Wait Apache Ignite to be started
-sleep 10
+#sleep 10
```
BryanCutler: is waiting no longer needed or done somewhere else?

yongtang: Thanks @BryanCutler. Just removed this one as Ignite is only going to be used after build is complete (takes about 10-15min) so I think no need to wait explicitly.

terrytangyuan:
Great job @yongtang!

damienpontifex:
Brilliant @yongtang!
I'm looking forward to trying this all out

yongtang:
@damienpontifex the build is based on macOS 10.12+xcode 9 (on Travis CI). Let me know if you encounter any issues.

PR #135, opened 2019-03-04
Add MNISTDataset which zip MNISTImageDataset and MNISTLabelDataset

yongtang:
This should help beginners to use tensorflow-io.

Also, updated the README.md to replace the example with a more
beginner-friendly MNIST example (matches the beginner tutorial
in tensorflow.org).

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,30 +35,39 @@ People who are a little more adventurous can also try our nightly binaries:
 $ pip install tensorflow-io-nightly
 ```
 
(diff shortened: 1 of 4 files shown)


yongtang:
/cc @terrytangyuan to take a look. The starter example has been replaced with MNIST (and matches the beginner tutorial in tensorflow.org).

That should make tensorflow-io more `beginner-friendly`.

I am thinking maybe it makes sense to change the R example to use MNIST as well if possible?

PR #140, opened 2019-03-10
Add CIFAR dataset support with compression

yongtang:
This PR adds the CIFAR data format support. Unlike MNIST data format,
which has been used in quite a few other places like:
- MNIST
- Fashion MNIST
- Kuzushiji-MNIST
- EMNIST

, the CIFAR is not widely used (at least not widely seen) other
than its original CIFAR10/CIFAR100 format. However, CIFAR is
a good candidate for us to look into the archive/compression
which will be used for other places. CIFAR is limited
so the impact of this implementation is limited. We could
easily revisit and rework on CIFAR once we have a better
understanding of compression/archive format.


yongtang:
The PR is ready for review now. One interesting point about CIFAR 10 is that CIFAR 10 train data consists of multiple objects within the zip file: `data_batch_[1-5].bin`. The complete set is a concatenation of those objects in the zip file. Think the template here will help in the future as well, as we will encounter similar scenarios for other formats, e.g., multiple video clips combined, etc.

/cc @terrytangyuan to take a look.

tensorflow_io/cifar/python/ops/cifar_dataset_ops.py
```diff
@@ -0,0 +1,148 @@
...
+
+  @property
+  def output_shapes(self):
+    return (tensorflow.TensorShape([32, 32, 3]), tensorflow.TensorShape([]))
+
+class _CIFAR100Dataset(data.Dataset):
+  """A CIFAR File Dataset that reads the cifar file."""
+
+  def __init__(self, filename):
+    """Create a `CIFARDataset`.
```
terrytangyuan: This should be CIFAR100

tensorflow_io/cifar/python/ops/cifar_dataset_ops.py
```diff
@@ -0,0 +1,148 @@
...
+
+  @property
+  def output_types(self):
+    return dtypes.uint8, dtypes.uint8, dtypes.uint8
+
+class CIFAR100Dataset(_CIFAR100Dataset):
+  """A CIFAR File Dataset that reads the cifar file."""
+
+  def __init__(self, filename, test=False, mode='fine'):
+    """Create a `CIFAR10Dataset`.
```
terrytangyuan: This should be CIFAR100

tensorflow_io/cifar/python/ops/cifar_dataset_ops.py
```diff
@@ -0,0 +1,148 @@
...
+  def output_types(self):
+    return dtypes.uint8, dtypes.uint8
+
+class CIFAR10Dataset(_CIFAR10Dataset):
+  """A CIFAR File Dataset that reads the cifar file."""
+
+  def __init__(self, filename, test=False):
+    """Create a `CIFAR10Dataset`.
+
+    Args:
```
terrytangyuan: Missing doc string for test

tensorflow_io/cifar/python/ops/cifar_dataset_ops.py
```diff
@@ -0,0 +1,148 @@
...
+  def output_types(self):
+    return dtypes.uint8, dtypes.uint8, dtypes.uint8
+
+class CIFAR100Dataset(_CIFAR100Dataset):
+  """A CIFAR File Dataset that reads the cifar file."""
+
+  def __init__(self, filename, test=False, mode='fine'):
+    """Create a `CIFAR10Dataset`.
+
+    Args:
```
terrytangyuan: Missing doc strings for the rest of the args

tensorflow_io/cifar/python/ops/cifar_dataset_ops.py
```diff
@@ -0,0 +1,148 @@
...
+            tensorflow.TensorShape([]),
+            tensorflow.TensorShape([3, 32, 32]))
+
+  @property
+  def output_types(self):
+    return dtypes.uint8, dtypes.uint8, dtypes.uint8
+
+class CIFAR100Dataset(_CIFAR100Dataset):
+  """A CIFAR File Dataset that reads the cifar file."""
+
```
terrytangyuan: Are we expecting other types than train and test? If so, we can introduce new argument for different types

tensorflow_io/cifar/python/ops/cifar_dataset_ops.py
```diff
@@ -0,0 +1,148 @@
...
+from __future__ import division
+from __future__ import print_function
+
+import tensorflow
+from tensorflow import dtypes
+from tensorflow.compat.v1 import data
+from tensorflow_io import _load_library
+cifar_ops = _load_library('_cifar_ops.so')
+
+class _CIFAR10Dataset(data.Dataset):
```
terrytangyuan: Could you clarify the differences between the class whose name starts with underscore and the other class? Maybe it would be better to combine them into one class?

yongtang:
Thanks @terrytangyuan for the review. The `_CIFAR10Dataset` and `CIFAR10Dataset` is a workaround for the limitations of `Dataset`. The `tf.data.Dataset` returns a `variant` (like `any`) to represent the handle and expect everything is done in graph. However, the native CIFAR format is `[C, H, W]` while we mostly render `[H, W, C]` in other classes. We certainly could add the `transpose` op in the graph in C++ but that seems like an overkill as the op `transpose` could be easily added in python. So the workaround is to have a `_CIFAR10Dataset` that generate native format, then `CIFAR10Dataset` apply additional `transpose` (see `as_variant_tensor` in `CIFAR10Dataset`):
```
  def _as_variant_tensor(self):
     return _CIFAR10Dataset(self._filename).map( # pylint: disable=protected-access
         lambda label, image: (tensorflow.transpose(image, [1, 2, 0]), label))._as_variant_tensor()
```

yongtang:
@terrytangyuan One enhancement in this PR is that the compression logic for file inputs has been baked in template, together with the creation of C++ op itself. In the future, new additions of other format only need to implement a `FromStream` and `ReadRecord` and use `InputDatasetOp<CustomInputType>` to register (`InputDatasetOp` is a template so no additional impl).

yongtang:
@terrytangyuan I am looking into the additional enhancement of simplification to add new custom ops. This PR is an early step, will follow up with additional PRs. Eventually the new ops creation should only need to provide a parsing logic of data and not worrying about anything else like graph definition, etc.

Issue #109, opened 2019-02-22
Idea:  DT_VARIANT type for ImageSets / ImageSource?


suphoff:
When looking through the current image operations and how to add functionality I quickly came to imagine lots and lots of per file format special operations - and did not really like that picture.

Currently all operations on image and video files (on file system or as strings) parse the files, extract one (or more) pictures as 2-3 dimensional data tensors and then throw away all parsing information.

This severely restricts the usability - all possible operations would have to be defined per file format - and each operation starts from scratch from an unopened file.   

If ImageSets (Single Pictures, Multiple Pictures (Tiff), Videos (Single or multi channel)) would be wrapped in a DT_VARIANT they could become first class objects in TF.

ImageSet Operations could then extract information from the ImageSet like cardinality, specific images, image sizes, ...

Example usage pattern in pseudo TF code:

```python
# Extract Random Image from TIF file
image_set = TiffImageSet('example.tif')
n = ImageSetCardinality(image_set)
index = tf.random.uniform(1, minval = 0; maxval = n, dtype=tf.dtypes.int64)
image = ImageSetGetImage(image_set, index)
```
A similar argument could be made to represent single potential images as an ImageSource Object.
This way meta data (focal length, resolution, ...) information could be  retrieved from the  ImageSource Object using special ImageSource Operations, used in calculations and then used as parameters to retrieve a cropped and scaled subset with another ImageSource Operation.

Example usage:
```python
image_source = Picture('car.jpg');
 # Extract Low RES picture from image_source and find license plate
license_plate_area = find_license_plate(image_source)
# Extract high Resolution patch from image_source and read license  
plate_image = ImageSourceExtractPatch(image_source, license_plate_area)
tag = read_tag(plate_image)
```

I believe ImageSource operations could also be made batch friendly to avoid unnecessary copy operations on batching. 

This may not be a realistic idea - but I wanted to bring it up as this may be the ideal time to specify a new generic interface. (There are enough existing formats in the repository to verify interface, but not too many to make the task unmanageable.)

Thoughts?

yongtang:
@suphoff These are definitely good ideas! May still need to see if these could be applied into TF's graph mode but it should work in eager mode I think.

The metadata information of those images (either as a standalone image, or as one frame within Video/TIFF, or maybe GIF animations) are very useful and could help extract images of interests. Maybe we could start building some individual ops to see how it could be applied?

suphoff:
[...]

PR #143, opened 2019-03-11
Add ImageInputSequence class for metadata information with a list of images

yongtang:
This PR adds an ImageInputSequence class so that metadata information
of a list of images (one image could have multiple frames) is stored
as a resource. The meta data information could later be
obtained without re-compute (as the information are static).

At the moment two methods are added: `count()` which returns
the number of frames, and `type()` which returns the fourcc
file type of each frame.

Additional meta data information could be added later.

This PR is inspired by #109.

### DIFF
--- a/tensorflow_io/image/BUILD
+++ b/tensorflow_io/image/BUILD
@@ -5,9 +5,14 @@ package(default_visibility = ["//visibility:public"])
 cc_binary(
     name = 'python/ops/_image_ops.so',
     srcs = [
+        "kernels/image_ops.h",
+        "kernels/image_ops.cc",
         "kernels/webp_dataset_ops.cc",
         "kernels/tiff_dataset_ops.cc",
         "ops/dataset_ops.cc",
+    ],
+    includes = [
+        ".",
     ],
     linkshared = 1,
     deps = [
--- a/tensorflow_io/image/__init__.py
+++ b/tensorflow_io/image/__init__.py
@@ -17,6 +17,7 @@
 @@WebPDataset
 @@TIFFDataset
 @@decode_webp
+@@ImageInputSequence
 """
 
 from __future__ import absolute_import
@@ -26,6 +27,7 @@ from __future__ import print_function
 from tensorflow_io.image.python.ops.image_dataset_ops import WebPDataset
 from tensorflow_io.image.python.ops.image_dataset_ops import TIFFDataset
 from tensorflow_io.image.python.ops.image_dataset_ops import decode_webp
+from tensorflow_io.image.python.ops.image_dataset_ops import ImageInputSequence
 
 from tensorflow.python.util.all_util import remove_undocumented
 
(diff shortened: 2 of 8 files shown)


suphoff:
@yongtang: I hope I can convince you to use Variants instead of Resources.
In my opinion Variants are a lot more versatile as they can be stored as tensors and passed around in a compute graph - just like any other values. One example operation I had in mind would take a vector (batch) of Imagesource variants and decode (potentially cropped, scaled) images with equal size directly into a batched output tensor.
In general I have the feeling that passing references around (via Variants) will allow future optimizations using lazy evaluation, All the benefits of Resources (persistence across runs) can easily be archived by storing variants tensors in TF variables - additionally allowing multiple variants to be stored in a persistent table.

I just pushed some utility code to make it easy to wrap objects in Variants to https://github.com/suphoff/io/tree/imagesource - it only contains the framework and no actual image related code as I got stuck last week looking at the vastly different capabilities of the standard low level image libraries. ( jpeg - discrete scaling only, webp - cropping only on even values for some encodeing, ..)
However I hope that you find it useful as it allows for easy separation of Image related code to common boiler plate code required to wrap objects into Variants.

yongtang:
@suphoff The variant type really makes sense for image operations. Though the resource approach I implemented, is actually focusing on a different problem about file compression/archive/containers.

As we adds more support for file formats and provides user a better experience, we want to automatically process compression, archiving, and containers. For example, when user specify a zip file, we should just atomically decompress the zip into a collection of files, and treat one zip file as a set of files. Zip file could be big and many contents might not be needed (e.g., user may only want to process jpeg files in zip and ignore and blob file or txt files).

Zip is a standard format for archive but there are other "container" types of files. One example is the video file. We could treat one video file as a standalone blob and decode, or we could treat one video file as a "container/archive" with many images. Decoding everything in a big video file (could be GBs), or even just copy this GBs of data into memory, is really unnecessary, and hard to manage in a graph. I think it fits naturally with resource as we could avoid passing GB's of protobuf-encoded message in graph around.

After looking into your idea, I think we could do it in this way:
- A resource that handles archives, directories, or containers (e.g., mp4) as a sequence of objects.
- The resource will expose an `getitem` types of API so that it is possible to extract one or multiple objects of interests. The returned value will be variant.
- At another level, we will process the variant (object), decode/crop the image or frame, or get the image meta data information like height, width, channel, exactly as you have been doing.

I think that will be a good plan to maintain separation of objects (variant) and the higher level processing (e.g., crop/bbox/etc).

yongtang:
@suphoff I will close this PR, and open another PR to keep the resource management with focus on archives, compressions, and containers. Will you be able to open a PR to focus on the image processing part? I think a PR (even if WIP) would be great.

suphoff:
@yongtang : I will try to push out a WIP this week.

I had never looked at the zip format internals before  - have you considered implementing a stacked zip filesystem on top of the tensorflow filesystem?

PR #152, opened 2019-03-18
Add output support for Text files

yongtang:
This PR add the output support for text files. It is served
as an experimental for later addtions of streaming support
such as Kfaka.

The overall idea is that a resource is associated with the
text output file. A write operation is a tuple of (index, item)
to make sure the output sequence is in order. Inside the kernel
the output sequence is cached (if out of order items are received
first). Cached items will always be flushed to text file as soon
as part of them are in order. Note that out-of-order output sequence
could happen in graph mode.

Some additional work might be needed, e.g., limit the cache size.

The test is to show the usage with fashion mnist tutorial on
tensorflow.org, through tf.keras's output callback.

Combined with output support, it should be possible to have
an inference infrastructue that generate the results to stream sink
(e.g., kafka producer). Kafka output support will be added in
follow up PR.


yongtang:
All tests pass now. Think the PR is ready for review.

PR #158, opened 2019-03-25
Part 1: migration of contrib.bigtable to tensorflow_io/bigtable

henrytansetiawan:
This Part 1, focusing on migrating the c++ ops/kernels of contrib/bigtable and the initial build plumbing, adapting the build scaffolding from tensorflow/tensorflow to tensorflow_io.


yongtang:
@henrytansetiawan The PR looks good! Most of the tests pass except for TF 2.0 tests. As far as I can see the reason for the test failure is because TF 2.0 changed the signature of `ResourceBase:: DebugString()`. Before 2.0, there is no `const` and after 2.0 there is a `const`.

I think using the following will fix the issue in TF2.0 and all tests should pass.

```c++
#if TF_MAJOR_VERSION==1&&TF_MINOR_VERSION==13
  virtual string DebugString() {
#else
  virtual string DebugString() const {
#endif
```

yongtang:
@henrytansetiawan Sorry, I forgot to mention, the `TF_MAJOR_VERSION` and `TF_MINOR_VERSION` are defined in `version.h` and this header file may not be included. So you may need to add:
```
#include "tensorflow/core/public/version.h"
```
at the beginning to have access to `TF_MAJOR_VERSION` and `TF_MINOR_VERSION`.

henrytansetiawan:
The blocker has been fixed.

yongtang:
@henrytansetiawan You may need to rebase against the master to update, so that the pubsub fix in #159 is included to pass Travis CI.

PR #159, opened 2019-03-26
Fix google-cloud-pubsub to 0.39.1 and google/cloud-sdk to 236.0.0

yongtang:
This PR is to fix google cloud pubsub hanging issue so that future PR could be unblocked.

/cc @henrytansetiawan


yongtang:
All tests passed after Pubsub version fixed. This is a master fix so I will merge it to unblock future PRs.

PR #168, opened 2019-04-03
Update to use tensorflow==2.0.0alpha for preview version

yongtang:
Change preview to use tensorflow==2.0.0alpha to stabilize and use 2.0 alpha version.


yongtang:
This one is to fix the 2.0 version to 2.0.0alpha to make sure the build will not break. (Some changes in .so in tf makes tf-nightly-2.0-preview not compatible).

yongtang:
Merge now to fix the build

PR #169, opened 2019-04-03
Fix ArrowDatasets repeating values after extra calls to Iterator GetNext 

### LINKED ISSUE #165: ArrowDataset repeats the entries when using with map and batch
I'm using ```tensorflow==1.13.1``` and ```tensorflow-io==0.4.0``` from pip

Here is the code I used to confirm the behavior

```python
import tensorflow as tf
from tensorflow_io.arrow import ArrowDataset
import pandas as pd

print('Using Range')

ds = tf.data.Dataset.range(10)
ds = ds.map(lambda x: x)
ds = ds.batch(4)
it = ds.make_one_shot_iterator()
d = it.get_next()

c = tf.ConfigProto()
c.gpu_options.allow_growth = True

with tf.Session(config=c) as sess:
    try:
        for i in range(1, 12):
            data = sess.run(d)
            print(data)
    except:
        print('Out of Range')

print('Using ArrowDataset')

ddf = pd.DataFrame({'A': range(10)})
ds = ArrowDataset.from_pandas(ddf, preserve_index=False)
ds = ds.map(lambda x: x)
ds = ds.batch(4)
it = ds.make_one_shot_iterator()
d = it.get_next()

c = tf.ConfigProto()
c.gpu_options.allow_growth = True

with tf.Session(config=c) as sess:
    try:
        for i in range(1, 12):
            data = sess.run(d)
            print(data)
    except:
        print('Out of Range')
```

Output

```
Using Range
[0 1 2 3]
[4 5 6 7]
[8 9]
Out of Range
Using ArrowDataset
[0 1 2 3]
[4 5 6 7]
[8 9]
[1 2 3 4]
[5 6 7 8]
[9]
[2 3 4 5]
[6 7 8 9]
Out of Range
```

It repeats the other elements for some reason.

BryanCutler:
This fixes a bug where an ArrowDataset.Iterator reaches the end of sequence, but the caller makes additional calls to `GetNext`, causing the ArrowDataset to reset and repeat values.  This was observed when used with `map().batch()` which creates a `map_and_batch_dataset_op`.

Now there is a final state that is set in ArrowDatasets, so additional calls to `GetNext` will only set `end_of_sequence=true` again.

A unit test was added that uses `ArrowDataset().map().batch()`

Fixes #165

### DIFF
--- a/tensorflow_io/arrow/kernels/arrow_dataset_ops.cc
+++ b/tensorflow_io/arrow/kernels/arrow_dataset_ops.cc
@@ -212,13 +212,14 @@ class ArrowDatasetBase : public DatasetBase {
                            bool* end_of_sequence) override {
       mutex_lock l(mu_);
 
-      // Intialize and read first batch
-      if (current_batch_ == nullptr) {
+      // If in initial state, setup and read first batch
+      if (current_batch_ == nullptr && current_row_idx_ == 0) {
         TF_RETURN_IF_ERROR(SetupStreamsLocked(ctx->env()));
       }
 
-      // Try to go to next batch if consumed all rows
-      if (current_row_idx_ >= current_batch_->num_rows()) {
+      // Try to go to next batch if consumed all rows in current batch
+      if (current_batch_ != nullptr &&
+          current_row_idx_ >= current_batch_->num_rows()) {
         TF_RETURN_IF_ERROR(NextStreamLocked());
       }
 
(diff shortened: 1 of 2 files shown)


BryanCutler:
@yongtang please review. This was caused because the `map_and_batch_dataset_op` will make repeated calls to `GetNext` even after `end_of_sequence=true`.

BryanCutler:
Thanks @yongtang !

PR #180, opened 2019-04-12
Add zlib dependency to libarchive.BUILD

BryanCutler:
With the addition of #140, the libarchive build is dependent on the zlib library. This PR adds zlib as a dependency to libarchive.BUILD

### DIFF
--- a/third_party/libarchive.BUILD
+++ b/third_party/libarchive.BUILD
@@ -33,6 +33,9 @@ cc_library(
         "libarchive",
     ],
     visibility = ["//visibility:public"],
+    deps = [
+        "@zlib",
+    ],
 )
 
 genrule(


BryanCutler:
@yongtang , if this looks ok I can try to merge it :)

yongtang:
@BryanCutler Thanks! I think the issue could be resolved by adding the dependency of `"@zlib"` to `third_party/libarchive.BUILD` like
```
+    deps = [
+        "@zlib",
+    ],
```

We build most of the libraries into binary (not relying on system provided libraries), which pretty much aligns with Tensorflow's pattern.

There are some pros and cons to this approach. The positive side is that we could control the exact behavior of the program (not depending on different version of system provided libraries). The negative side is obviously the increased binary size (for the same reason tensorflow's pip package is about 100MB).

On a size note, zlib is less of an issue, because zlib API has been super stable over the years (if not decades). System provided one does not have any incompatible issues.

There are some discussions about building tensorflow with system provided libraries and distribute with Linux distributions. For example, CentOS 7 will have one binary exclusively build on it (with system library dependencies, thus smaller size of tensorflow), and Ubuntu 18.04 will have another one binary built with system gcc and python. I believe at one point there were some effort to bundle TensorFlow with Linux distributions like Debian. (don't know the current status or plan yet).

At the moment, I think we can just stick with the path of building all library into binary. When or once TensorFlow releases Linux distribution specific binaries (and smaller package size), we could follow as well to release rpm and deb for tensorflow-io.

Now for the zlib issue, the issue is not exposed to our Travis CI build, because Travis CI build uses buildpack-deps (instead of vanilla Ubuntu) which somehow already have zlib package installed. Unfortunately, even though bazel have a sandbox mode (and we use sandbox mode), it does not completely guard the unintended library headers I believe. So in buildpack-deps the zlib.h was in.

We should have used Ubuntu docker image to maintain consistency across. At one point I switched to buildpack-deps for the purpose of reduce the build time. Our build time have been very close to 50 minutes limit posed by Travis CI. Maybe we should start migrating to Kokorun or other services soon.

Anyway, I think adding `@zlib` to `libarchive.BUILD` will help resolve the issue for now. In the future, we will need to
- Update the CI docker image to use Ubuntu
- In case TensorFlow provides rpm or deb package, tensorflow-io will also provide rpm or deb and build the code with system libraries.

yongtang:
@BryanCutler can you update the PR with zlib dips on libarchive.BUILD? You can use the GitHub UI to merge. Let me know if you run into any issues.

BryanCutler:
Thanks @yongtang , and yes I agree that it would probably be best to add the dep to libarchive.BUILD. I'll change this up now. It would be nice to have CI use an Ubuntu image also, but we can work that out later after we get things better situated with the time limits :)

yongtang:
@BryanCutler Looks like all tests finally passed. You can give it a try to see you are able to merge from GitHub UI.

BryanCutler:
merged to master, woohoo! Thanks @yongtang

PR #181, opened 2019-04-12
Fix libsvm error

### LINKED ISSUE #179: global name 'core_readers' is not defined
When I try to use function `make_libsvm_dataset`, it notice me the error that
```
NameError: global name 'core_readers' is not defined
```
How could I fix this?

yongtang:
This PR fixes #179, where libsvm encountered some issues with Api changes. Also added a test case for it.


yongtang:
Let's merge this bug fix to be in time for the 0.5.0 release.

PR #184, opened 2019-04-13
Refactor MNIST implementation

yongtang:
This fix reuses the implementation in CIFAR, reduces code duplication, and also added a couple of test cases in eager mode


yongtang:
@terrytangyuan CIFAR dataset's C++ code has been reduced from 470+ lines to 130+ lines. Should help contributors to writer ops easier.

yongtang:
Added in graph optimization. Dataset's implementation seems to have some trouble with Variant, so converts inputs to accept either variant, or string types. In case of python interface, variant is still used. The internal graph optimization used serialized variant (string) now.

PR #190, opened 2019-04-15
Fix soname issue in tf-nightly (and 1.14.0)

yongtang:
Looks like tf-nightly (and potentially tf 1.14.0) changes the shared object extension, so update of the build script is needed.

### DIFF
--- a/config_helper.py
+++ b/config_helper.py
@@ -75,6 +75,13 @@ def write_config():
 
       bazel_rc.write('build --action_env TF_SHARED_LIBRARY_DIR="{}"\n'
                      .format(libdir_list[0][2:]))
+      library_name = library_list[0][2:]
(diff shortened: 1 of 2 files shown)


yongtang:
The soname issue with tf-nightly has been fixed. There is another issue about protobuf header file which has been fixed in https://github.com/tensorflow/tensorflow/commit/79d8779069a154a74786c28bf9d8af3d0cbd9905

So the build will pass after the https://github.com/tensorflow/tensorflow/commit/79d8779069a154a74786c28bf9d8af3d0cbd9905 is part of the tf-nightly. (tomorrow?)

yongtang:
Some plumbing changes as tensorflow changed the library .so file to include soname. Not impacting exiting one until 1.14.0+. Merging.

PR #192, opened 2019-04-16
Fix kafka dataset op save and restore, and add some log

zhjunqin:
I find some other problems during testing kafka dataset op.
1. The Kafka dataset OP cannot save and restore from checkpoint. If do restore, it would lead to core dump. The root cause is the `consumer_` not exists when function `ResetStreamsLocked` called during restore.
2. The `RestoreInternal` function cannot restore to latest offset. Seems `consumer_->assign` needs to be called during restore.
3.  The function `consumer_->poll` seems must not be used for `RdKafka::KafkaConsumer`, check document https://docs.confluent.io/2.0.0/clients/librdkafka/classRdKafka_1_1Handle.html#a98d3b4ee48457ff13e4d5155e3fc5ea4. It would also lead to core dump in some case.
4. I think it would be better to add some LOG info for Kafka dataset OP, otherwise, it would be hard to debug.
5. I add a test case `test_kafka_dataset_save_and_restore` to test save and restore.
Thanks!

### DIFF
--- a/tensorflow_io/kafka/kernels/kafka_dataset_ops.cc
+++ b/tensorflow_io/kafka/kernels/kafka_dataset_ops.cc
@@ -14,6 +14,7 @@ limitations under the License.
 ==============================================================================*/
 
 #include "tensorflow/core/framework/dataset.h"
+#include "tensorflow/core/platform/logging.h"
 
 #include "rdkafkacpp.h"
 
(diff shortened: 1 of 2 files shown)


yongtang:
@zhjunqin Graph has been removed in 2.0. I think you could add the following label so that the test could be skipped in 2.0.
```
@pytest.mark.skipif(
    (hasattr(tf, "version") and
         tf.version.VERSION.startswith("2.0.")), reason=None)
```

zhjunqin:
> @zhjunqin Graph has been removed in 2.0. I think you could add the following label so that the test could be skipped in 2.0.
> 
> ```
> @pytest.mark.skipif(
>     (hasattr(tf, "version") and
>          tf.version.VERSION.startswith("2.0.")), reason=None)
> ```

@yongtang Thanks for the reminder!  CI passed finally.  Could you help review code?

PR #193, opened 2019-04-16
Enable Python 3.7 build

yongtang:
Think I managed to make python 3.7 build working.

/cc @terrytangyuan @BryanCutler @dmitrievanthony 

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -62,6 +62,7 @@ jobs:
     before_script: *developer_build
     script:
     - docker run -i -t --rm -v $PWD:/v -v $PWD/.cache/pip/:/root/.cache/pip -w /v --net=host buildpack-deps:18.04 bash -x -e script.sh
+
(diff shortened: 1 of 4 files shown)


BryanCutler:
Nice work @yongtang !  The 50 min limit seems to be a real problem now..

yongtang:
@BryanCutler Yes it is concerning as we are gradually reaching 50min. Something may have to be done soon. Another factor to take into consideration is the upcoming tf 1.14 release. I think we may be able to do something once 1.14 is out, see if we could get help from Kokorun(Google C).

yongtang:
At the moment, the biggest chunk of the time is the C++ code compile which is close to 25-30 minutes. Other than that, we have Ignite/Kafka/Kinesis/PubSub setup for about 10 minutes. Kind of hard to optimize with Travis CI.

PR #203, opened 2019-04-19
Change dev Docker image to use tensorflow:custom-op as base

### LINKED ISSUE #173: Update dev/Dockerfile to use gcc 4.8 by default
The default gcc installed on the Docker image is gcc 7.0.3. Compiling custom ops using newer gcc and interop-ing with older gcc 4.8 used by official TF build prones to Segmentation Fault ABI issue: https://github.com/tensorflow/tensorflow/issues/13308.

some reference:
- https://gist.github.com/application2000/73fd6f4bf1be6600a2cf9f56315a2d91
- https://gist.github.com/vmaks/06ac9bcd859621aa589f

BryanCutler:
This changes the tensorflow-io developer Docker file to be based from tensorflow/tensorflow:custom-op. Using this base image gives us Ubuntu 14.04 and gcc 4.8, which is in-line with what TensorFlow is using to allow for better compatibility. The remaining dev image stays the same, installing a newer Bazel version and conda environment with required packages for building/testing.

Fixes #173

### DIFF
--- a/dev/Dockerfile
+++ b/dev/Dockerfile
@@ -1,14 +1,13 @@
-FROM ubuntu:18.04
+FROM tensorflow/tensorflow:custom-op
 
 RUN apt-get update && \
     apt-get install -y \
-    build-essential \
     curl \
     nano \
     unzip \
     ffmpeg
 
-ARG BAZEL_VERSION=0.20.0
+ARG BAZEL_VERSION=0.24.1
 ARG BAZEL_OS=linux
 
 RUN curl -sL https://github.com/bazelbuild/bazel/releases/download/${BAZEL_VERSION}/bazel-${BAZEL_VERSION}-installer-${BAZEL_OS}-x86_64.sh -o bazel-install.sh && \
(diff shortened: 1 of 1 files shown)


dev/Dockerfile
```diff
@@ -1,14 +1,13 @@
-FROM ubuntu:18.04
+FROM tensorflow/tensorflow:custom-op
 
 RUN apt-get update && \
     apt-get install -y \
-    build-essential \
```
BryanCutler: custom-op should set up the required C++ build tools

dev/Dockerfile
```diff
@@ -1,14 +1,13 @@
...
     curl \
     nano \
     unzip \
     ffmpeg
 
-ARG BAZEL_VERSION=0.20.0
+ARG BAZEL_VERSION=0.24.1
```
BryanCutler: @yongtang I think you said 0.24.1 is proffered to build TF, so updating that here

BryanCutler:
@yongtang could you please have a look and make sure I'm not missing any required package from the latest ops, thanks!

PR #205, opened 2019-04-19
Allow ArrowDataset to create an initializable iterator for input data

### LINKED ISSUE #204: Allow ArrowDataset to use an initializable iterator
The `ArrowDataset` serializes data in memory and could be made more efficient by creating an initializable iterator and using `feed_dict` to initialize the input.

BryanCutler:
This PR changes the input of the `ArrowDataset` to be a string Tensor that contains a serialized Arrow buffer. A placeholder Tensor can then be used to create the Dataset and then the iterator can be initialized with the serialized data when running the session. By doing this, it prevents the data from being stored in the graph as a constant, which leads to unnecessarily copies.

Fixes #204

### DIFF
--- a/tensorflow_io/arrow/kernels/arrow_dataset_ops.cc
+++ b/tensorflow_io/arrow/kernels/arrow_dataset_ops.cc
@@ -19,6 +19,7 @@ limitations under the License.
 #include "arrow/util/io-util.h"
 #include "tensorflow_io/arrow/kernels/arrow_stream_client.h"
 #include "tensorflow/core/framework/dataset.h"
+#include "tensorflow/core/graph/graph.h"
 
 #define CHECK_ARROW(arrow_status)             \
   do {                                        \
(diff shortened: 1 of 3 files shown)


BryanCutler:
TODO need to make test conditional on 2.0 because of placeholder

BryanCutler:
Ping @yongtang and @terrytangyuan if you could please take a look when you get the chance? I ran some tests and these changes still help quite a bit. Thanks!

yongtang:
Overall, the Dataset itself is a pull model and is capable of pull a chunk (or part) of the data at a time. But when the whole data is already in memory, it is hard to feed data into Dataset in piecemeal, as Dataset will require converting the whole data into tensor. We may want to address this issue, as was pointed by:
https://stackoverflow.com/questions/48956404/tensorflow-dataset-tf-estimator-inputs-numpy-input-fn

I am playing with the idea of exposing the data in memory (in python or other languages) as a "server" to allow Dataset to "pull" from "server", will see if I could come up with some PoC.

BryanCutler:
Thanks @yongtang !

>I  am playing with the idea of exposing the data in memory (in python or  other languages) as a "server" to allow Dataset to "pull" from "server",  will see if I could come up with some PoC.

Sounds good, I've been thinking about something very similar. I think it could be very useful..

yongtang:
@BryanCutler Created an initial PR #206.

PR #206, opened 2019-04-24
Add GRPCDataset to allow pulling data from a gRPC server such as numpy array

yongtang:
This fix tries to have a way to pull data from a gRPC server, as long as a gRPC server implement the protocol in endpoint.proto.

The main purpose is to allow reading partial data from numpy array in memory.

In the from_numpy method, the following is done:
- Create a gRPC server and exposes ReadRecord endpoint.
- Start the gRPC server on a random port in local host.
- Pass endpoint (<localhost:port>) to GRPCDataset
- GRPCDataset will pull the data from gRPC server.

Note from_numpy is just a facility method to setup a gRPC server. In theory, a GRPC server could be created by any other process and by any language. Only the endpoint information is needed to create GRPCDataset.

In case numpy array is huge, then this method could be helpful as it is not required to save numpy file, and load back by tf.data, or pass the whole numpy data into a tensor.

This could open up doors for other languages as well. For example, in case of R, we could setup a grpc server with R, and pass R dataframes in memory and allows GRPCDataset to pull a chunk of the data at a time.

Note we create gRPC server locally, but it is possible to expose gRPC server remotely so that the workload could be distributed.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -149,7 +149,9 @@ if datapath is not None:
       os.path.join(datapath, "tensorflow_io")):
     if (not fnmatch.fnmatch(rootname, "*test*") and
         not fnmatch.fnmatch(rootname, "*runfiles*")):
-      for filename in fnmatch.filter(filenames, "*.so"):
+      for filename in [
+          f for f in filenames if fnmatch.fnmatch(
+              f, "*.so") or fnmatch.fnmatch(f, "*.py")]:
         src = os.path.join(rootname, filename)
         dst = os.path.join(
             rootpath,
--- a/tensorflow_io/core/kernels/dataset_ops.h
+++ b/tensorflow_io/core/kernels/dataset_ops.h
@@ -147,35 +147,83 @@ class ArchiveInputStream : public io::InputStreamInterface {
   TF_DISALLOW_COPY_AND_ASSIGN(ArchiveInputStream);
 };
 
+// Note: Forward declaration for friend class.
+template<typename T> class FileInput;
+template<typename T> class StreamInput;
+
 template<typename T>
 class DataInput {
  public:
   DataInput() {}
   virtual ~DataInput() {}
-  virtual Status FromStream(io::InputStreamInterface& s) = 0;
-  virtual Status ReadRecord(io::InputStreamInterface& s, IteratorContext* ctx, std::unique_ptr<T>& state, int64 record_to_read, int64* record_read, std::vector<Tensor>* out_tensors) const = 0;
+ protected:
   virtual void EncodeAttributes(VariantTensorData* data) const = 0;
   virtual bool DecodeAttributes(const VariantTensorData& data) = 0;
-
-  Status ReadInputStream(io::InputStreamInterface& s, int64 chunk, int64 count, string* buffer, int64* returned) const {
-    int64 offset = s.Tell();
-    int64 bytes_to_read = count * chunk;
-    Status status = (buffer == nullptr) ? s.SkipNBytes(bytes_to_read) : s.ReadNBytes(bytes_to_read, buffer);
-    if (!(status.ok() || status == errors::OutOfRange("EOF reached"))) {
-      return status;
+  virtual Status ReadReferenceRecord(void* s, IteratorContext* ctx, std::unique_ptr<T>& state, int64 record_to_read, int64* record_read, std::vector<Tensor>* out_tensors) const = 0;
+  Status ReadReferenceBatchRecord(void* s, IteratorContext* ctx, std::unique_ptr<T>& state, int64 batch, int64 count, int64* returned, std::vector<Tensor>* out_tensors) const {
+    int64 record_read = 0;
+    int64 record_to_read = count - (*returned);
+    std::vector<Tensor> chunk_tensors;
+    TF_RETURN_IF_ERROR(ReadReferenceRecord(s, ctx, state, record_to_read, &record_read, &chunk_tensors));
+    if (record_read > 0) {
+      if (out_tensors->size() == 0) {
+        // Replace out_tensors with chunk_tensors
+        out_tensors->reserve(chunk_tensors.size());
+        // batch == 0 could only read at most one record
+        // so it only happens here.
+        if (batch == 0) {
+          for (size_t i = 0; i < chunk_tensors.size(); i++) {
+            TensorShape shape = chunk_tensors[i].shape();
+            shape.RemoveDim(0);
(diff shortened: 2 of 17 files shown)


setup.py
```diff
@@ -156,6 +156,13 @@ def has_ext_modules(self):
             os.path.relpath(os.path.join(rootname, filename), datapath))
         print("setup.py - copy {} to {}".format(src, dst))
         shutil.copyfile(src, dst)
+      for filename in fnmatch.filter(filenames, "*.py"):
+        src = os.path.join(rootname, filename)
+        dst = os.path.join(
+            rootpath,
+            os.path.relpath(os.path.join(rootname, filename), datapath))
+        print("setup.py - copy {} to {}".format(src, dst))
+        shutil.copyfile(src, dst)
```
BryanCutler: Just wondering why adding this now, was it grabbing some incorrect files?

Also could you just reuse the similar loop above and just nest under a list of extensions? like

```
for file_pattern in ["*.so", "*py"]:
  for filename in fnmatch.filter(filenames, file_pattern):
   ...
```

tensorflow_io/core/kernels/dataset_ops.h
```diff
@@ -152,26 +152,89 @@ class DataInput {
  public:
   DataInput() {}
   virtual ~DataInput() {}
-  virtual Status FromStream(io::InputStreamInterface& s) = 0;
-  virtual Status ReadRecord(io::InputStreamInterface& s, IteratorContext* ctx, std::unique_ptr<T>& state, int64 record_to_read, int64* record_read, std::vector<Tensor>* out_tensors) const = 0;
+  virtual Status ReadReferenceRecord(void* s, IteratorContext* ctx, std::unique_ptr<T>& state, int64 record_to_read, int64* record_read, std::vector<Tensor>* out_tensors) const = 0;
```
BryanCutler: Would you mind explaining why using `void *` here and the change to using pointers instead of references?

setup.py
yongtang: Thanks @BryanCutler. The PR has been updated. Previously, we pick up py files in the source tree, and pick up .so files in bazel-bin. However, for gRPC, it consists of generated code in python (.py) based on endpoint.proto file. Those files are generated as part of the build so they are placed in bazel-bin (by bazel). So setup.py has to be updated to capture the py files in bazel-bin as well.

tensorflow_io/core/kernels/dataset_ops.h
yongtang: @BryanCutler Previously, references works well as we are mostly dealing with file streams (`io::InputStreamInterface& s`). However, for gRPC it is not a file stream and is not returning raw bytes (no `ReadNBytes`). So `InputStreamInterface` is irrelevant here. Changed to pointers so that `nullptr` could be passed for gRPC.

The main purpose is to try to reuse the logic of batch where we need to piece together chunks of tensor.

yongtang:
Add a reference of https://github.com/tensorflow/tensorflow/issues/13530 , which has a need for pandas input. Could be a follow up PR I think.

tensorflow_io/core/kernels/dataset_ops.h
BryanCutler: Ok, thanks for the explanation. Having `void*` arguments on public apis make a little nervous though. Do you think putting this as a `protected` member makes sense?

tensorflow_io/grpc/BUILD
```diff
@@ -0,0 +1,65 @@
...
+    srcs = [
+        "endpoint.proto",
+    ],
+    outs = [
+        "python/ops/__init__.py",
+        "python/ops/endpoint_pb2.py",
+        "python/ops/endpoint_pb2_grpc.py",
+    ],
+    cmd = "python -m grpc_tools.protoc -Itensorflow_io/grpc --python_out=$(BINDIR)/tensorflow_io/grpc/python/ops/ --grpc_python_out=$(BINDIR)/tensorflow_io/grpc/python/ops/ $< ; touch $(BINDIR)/tensorflow_io/grpc/python/ops/__init__.py",
+    output_to_bindir = True,
```
BryanCutler: The flatbuffers build for Arrow had an option to put the generated files under a prefix directory, e.g. `out_prefix = "cpp/src/arrow/ipc/"`, is that possible to do here to just put the files in the right place initially?

yongtang: @BryanCutler Bazel used mount point to hide the source directory purposely, so that the generated code could not be exposed in source directory. So we have to copy the python file exposed in  `bazel-bin` (not the source directory). The flatbuffers case is different as the generated file are used for compile (bazel could see), not for final exposure (like python case we encounter here).

tensorflow_io/core/kernels/dataset_ops.h
yongtang: @BryanCutler Thanks for the review. The PR has been updated. Now FileInput (and subclass delcared) only need to implement:
```
ReadRecord(io::InputStreamInterface* s,  IteratorContext* ctx, ...)
```

and StreamInput only need to implement:
```
Status ReadRecord(IteratorContext* ctx, ...
```

So `void *` is hidden and will not be touched by any ops.

Please take a look.

yongtang:
@BryanCutler The PR has been updated. Please take a look.

tensorflow_io/grpc/python/ops/grpc_ops.py
```diff
@@ -0,0 +1,83 @@
...
+  @staticmethod
+  def from_numpy(a, batch=None):
+    """from_numpy"""
+    from tensorflow_io.grpc.python.ops import grpc_endpoint
+    grpc_server = grpc_endpoint.GRPCEndpoint(a)
+    grpc_server.start()
+    endpoint = grpc_server.endpoint()
+    dtype = a.dtype
+    shape = list(a.shape)
+    batch = batch
```
BryanCutler: Is this maybe a typo?

tensorflow_io/grpc/BUILD
BryanCutler: Ok, I see thanks for the explaination

tensorflow_io/core/kernels/dataset_ops.h
BryanCutler: Sounds good thanks!

tensorflow_io/grpc/python/ops/grpc_ops.py
yongtang: @BryanCutler Thanks! Just updated.

yongtang:
Let's merge this one. Saw some discussion about Java for which this PR should help.

PR #209, opened 2019-05-02
* Actually allow allowed symbols

zou000:
[`remove_undocumented` ](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/util/all_util.py#L86) removes undocumented symbols from the module, but allow exceptions to be passed via `allowed_exception_list` argument. These packages defined allowed symbols but forgot to pass them into `remove_undocumented` and thus has no effect. Actually, since all these symbols are documented so passing them to the function has no effect either, but this PR fixed the intention.

### DIFF
--- a/tensorflow_io/arrow/__init__.py
+++ b/tensorflow_io/arrow/__init__.py
@@ -35,4 +35,4 @@ _allowed_symbols = [
     "ArrowStreamDataset",
 ]
 
-remove_undocumented(__name__)
+remove_undocumented(__name__, allowed_exception_list=_allowed_symbols)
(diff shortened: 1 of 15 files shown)


yongtang:
@zou000 Not sure about this PR, could you add some description?

zou000:
> @zou000 Not sure about this PR, could you add some description?

Added more detailed description above.

yongtang:
@zou000 Thanks. The remove_undocumented was from tensorflow repo during the migration. Probably could have a better way of manage symbols in the future. But let's merge in anyway.

PR #211, opened 2019-05-02
Add initial Kokorun script

yongtang:
This is the initial setup of kokorun script.

The script is located in `.kokorun/io_cpu.sh`, so something like `bash -e .kokorun/io_cpu.sh` or `.kokorun/io.cpu.sh` will invoke the script.

### DIFF
--- a/.kokorun/io_cpu.sh
+++ b/.kokorun/io_cpu.sh
@@ -1,0 +1,54 @@
+#!/usr/bin/env bash
+# Copyright 2019 The TensorFlow Authors. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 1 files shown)


yongtang:
@jsimsa I created an initial version of kokorun script. It is located in `.kokorun/io_cpu.sh`. Don't know enough about kokorun but something like `bash -e .kokorun/io_cpu.sh` or `.kokorun/io.cpu.sh` on CI setup should be good enough I think.

The script might fail, but that should be ok. Once the CI flow with kokorun is up I will update the script to fix the build/test failures.

yongtang:
Also /cc @jsimsa  @dmitrievanthony @BryanCutler @terrytangyuan @ewilderj

Issue #212, opened 2019-05-04
Build shared object library into one


yongtang:
At the moment, tensorflow-io build shared object libraries for each module. The advantage is that it avoids name collision (if there is) but it also means much of the binary code are shipped multiple times. It would be good to combine shared object libraries into one to reduce the package size.

BryanCutler:
This would make sense for ops that share a lot of the same files/dependencies, but I don't know if it's a good idea to put everything into one shared object. That would end up having the user load a lot of things that might not be used right?

yongtang:
@BryanCutler It might be eventually an issue for us. At the moment Bigtable and pubsub are the biggest in code size (think they have duplications). Arrow and Parquet also are big as well.

Though compared with TensorFlow's code base itself, the tensorflow-io is still manageable even if we combine everything together (that by itself will reduce the total size).

I think we could try move into one first, and in case we see a slow down, we could do a better planning (e.g, have dependency .so). I do see have one libtensorflowio.so as an `entry point .so`, could be valuable. The libtensorflowio.so could depend on other `.so` as we move forward.

yongtang:
@BryanCutler Think further I think in the long run, we could distribute tensorflow-io with deb and rpm (in addition to the default whl). For deb and rpms we should build tensorflow-io directly with specific versions of Ubuntu/Debian/RHEL, and link against system installed libraries. This will greatly reduce the package size. (and is the standard way on those Linux distributions).

The deb/rpm may have to wait for tensorflow to release deb/rpm first, or wait for tensorflow modularization project complete, though.

BryanCutler:
Sounds good, I'm fine with trying an all-in-one for now. I like the idea of factoring out dependencies and common code to shared objects, but that would take some effort, so we can explore that another time if it seems necessary.

yongtang:
We pretty much have only one .so now (`libtensorflow_io.so`). The exceptions:
- libtensorflow_ffmpeg_.so (due to FFmpeg license limitation)
- libtensorflow_golang.so (due to issue #492)

Think this issue could be closed.

PR #214, opened 2019-05-04
Build shared object libraries into one

yongtang:
This PR is part of the effort to address #212. Text/MNIST/CIFAR has been moved into one libtensorflowio.so

Will gradually move other components as well.


yongtang:
@terrytangyuan PR has been updated with build failure fixed.

PR #218, opened 2019-05-05
Map zlib to com_github_madler_zlib, to avoid duplication of zlib binary

yongtang:


yongtang:
Plumbing change, let's merge this one.

Issue #131, opened 2019-03-03
Kafka dataset may exit in macOS


yongtang:
On macOS sometimes Kafka dataset may exit early:
https://travis-ci.org/tensorflow/io/jobs/499999592

This bug is likely caused by usage of librdkafka. The resources (e.g., messages) may have to be closed before close kafka consumer.

yongtang:
This issue is closed by PR #225.

PR #225, opened 2019-05-08
Enable Kafka test on macOS

yongtang:
I think issue #131 likely should have be fixed by PR #192,
so give it a try to enable Kafka on macOS.


yongtang:
First run successful: https://travis-ci.org/tensorflow/io/builds/529571850?utm_source=github_status&utm_medium=notification

Will try another run.

Issue #191, opened 2019-04-15
Add batch support for dataset at the creation


yongtang:
The current workflow for Dataset is to read one record at a time, then use a batch() call to aggregate the result. Since Dataset is processed sequentially, it really makes sense to add batch support when reading records. In other words, data will be copied in large memory chunks (batch) with multiple records at each process. Believe this will help improve the overall performance. The batch parameter could be passed as the output_shapes so it should be fairly easy. The only item remaining is that, in case of multiple files, batch will need to be pieced together in between two files (if the record count is not divided by the batch).

yongtang:
With PR #202 the batch support has been added for Text, MNIST, and CIFAR. Will update other ops as well.

BryanCutler:
I'm working on this for the Arrow Dataset ops, should have a PR ready soon.

yongtang:
Instead of add batch to creation directly, now we count on follow up op of `batch(n)` and `unbatch()` in tf.data.Dataset. Think this issue could be resolved.

PR #237, opened 2019-05-15
Add batch param to Arrow Datasets to batch directly on Arrow data

BryanCutler:
This adds a parameter to Arrow Datasets that allow the user to specify a batch size to be applied when converting Arrow data to Tensors. This is much more efficient that the alternative of batching the Dataset output after conversion because the batch can be copied directly from the arrow data buffer.

Partial batches are handled by loading the next chunk of Arrow data and then concatenating.  An option  to drop remainder is also available that will discard the remainder records that do not form a full batch.


BryanCutler:
There are a few more things to do before this is complete:

- [ ] Pass test with remaining data
- [ ] Add tests with multiple Arrow record batches
- [ ] Document the batch param

yongtang:
@BryanCutler For `drop_reminder`, I am wondering if it makes sense to change it to `string` (vs. `bool`) to allow future expansion? I haven't done that in other ops but I notices that batch could have different ways:
- leave as it is (the last shape of the elem will have different first dim)
- drop last one (all shapes of the elems are the same)
- pad last one (all shapes of the elems are the same)
- drop last one of the `file/natural boundary` (if there are two files, then each file behaves like drop last one, then combine two files, this is different from drop last one of the combined two files).
- pad last one of the `file/natural boundary` (similar but pad each file, then combine)

I think we don't need to implement every case. Though I think it makes sense to have an option of `batch_mode` (in string) to allow different mode of batching in the future?

BryanCutler:
@yongtang that sounds like a good idea for `drop_remainder`, I can see all those as being useful options. There is one more option for use with Arrow that I was thinking, that is to automatically use the batch size of the incoming Arrow record batch. I think that will often be the most efficient option for people that can control the Arrow batch size upfront. In that case, I would want to make the param `batch_mode` and then add the option `auto` to do this. WDYT?

yongtang:
@BryanCutler That looks good 👍. We could have batch_mode to have different ways of batching for different file types.

BryanCutler:
Sounds good. I'll change up the param and I could support leave as is, drop remainder and auto right now.

BryanCutler:
This implements #191 for Arrow datasets

BryanCutler:
@yongtang this is close to being done, but I hit a snag with https://github.com/tensorflow/tensorflow/issues/28775. I'll try to find a workaround, but if you want to start on the release, this doesn't need to block.

tensorflow_io/arrow/python/ops/arrow_dataset_ops.py
```diff
@@ -30,6 +30,7 @@
   from tensorflow import nest # pylint: disable=ungrouped-imports
 else:
   from tensorflow.python.data.util import nest # pylint: disable=ungrouped-imports
+from tensorflow.python.data.util import structure as structure_lib # pylint: disable=ungrouped-imports
```
BryanCutler: TODO: does this need a similar `hasattr` check above?

tensorflow_io/arrow/kernels/arrow_dataset_ops.cc
```diff
@@ -20,6 +20,7 @@ limitations under the License.
 #include "tensorflow_io/arrow/kernels/arrow_stream_client.h"
 #include "tensorflow/core/framework/dataset.h"
 #include "tensorflow/core/graph/graph.h"
+#include "tensorflow/core/util/batch_util.h"
```
BryanCutler: TODO: didn't end up using this

BryanCutler:
ok, I think this is ready for review

BryanCutler:
Gentle ping to @yongtang and @terrytangyuan to please review

tensorflow_io/arrow/python/ops/arrow_dataset_ops.py
```diff
@@ -312,8 +397,19 @@ def from_schema(cls, host, schema, columns=None):
             For a socket client, use "<HOST_IP>:<PORT>", for stdin use "STDIN".
       schema: Arrow schema defining the record batch data in the stream
       columns: A list of column indicies to use from the schema, None for all
+      batch_size: Batch size of output tensors
+      batch_mode: Mode of batching, supported strings:
+                  "keep_remainder" (default, keeps partial batch data),
+                  "drop_remainder" (discard partial batch data),
+                  "auto" (size to number of records in Arrow record batch)
```
terrytangyuan: Perhaps put this in an enum and then validate it?

tensorflow_io/arrow/python/ops/arrow_dataset_ops.py
```diff
@@ -134,19 +147,26 @@ def __init__(self,
       output_types: Tensor dtypes of the output tensors
       output_shapes: TensorShapes of the output tensors or None to
                      infer partial
+      batch_size: Batch size of output tensors
```
terrytangyuan: Maybe add a note on how this will play together with Dataset's own `batch()`

yongtang:
@BryanCutler Sorry for the late reply as I was out-of-office last week. The PR looks great. LGTM

PR #252, opened 2019-05-28
Release 0.6.0

BryanCutler:
Update `README.md` and `RELEASE.md` with recent features and contributor list and bump the version in `setup.py`.


BryanCutler:
Getting this prepped for when we are ready to make the release. @yongtang and @terrytangyuan please check if I've missed anything.

yongtang:
Think the setup.py also could be changed to 0.6.0.

BryanCutler:
>Think the setup.py also could be changed to 0.6.0

Oh right, I'll fix that now

PR #254, opened 2019-05-28
Add draw_text op for image

### LINKED ISSUE #232: Add draw_text support for tensorflow-io
We have plans to support both input and output of the images (and videos). Two really useful operations that will greatly help the usage of tensorflow-io, are draw_bbox(bounding box) and draw_text. The bounding box op is available, though text op is not available yet. It makes sense to add text support for images.

yongtang:
This PR is to add draw_text op for image, which could be really useful.

This PR fixes #232.

Here is a quick sample:
small-bb


yongtang:
To give some background on this PR:
Object detection and tracking is a very important area, and the last step could be the labeling and annotation of the tracked object.

In tensorflow, `tf.draw_bounding_box` allows you labeling the detected object based on the bounding box provided. It is possible to have a complete workflow in TF. The only missing piece is the text annotation.

It is possible to combine with other python tools to add the annotation such as:
https://towardsdatascience.com/object-detection-and-tracking-in-pytorch-b3cf1a696a98

Though it would be preferable to have a native TF support because it allows the action to be done in a graph and with other language bindings.

This PR extends `tf.draw_bounding_box` to optionally have a text tensor to specify the annotation for bounding box. If after some period of time the new version of bounding box is stable enough we could consider to merge it back in tensorflow main repo.

yongtang:
Note this PR does not using system font as systems font may be different and may not be available all the time. Instead this PR use Google's `Open Sans Regular` font which is Apache License 2.

yongtang:
@BryanCutler @terrytangyuan Here is an example of the usage (as is in tests/test_image.py):
```
with open(filename, "rb") as f:
    png_contents = f.read()

# decode png
image_tensor = tf.image.decode_png(png_contents, channels=channels)
# convert uint8 to float32
image_tensor = tf.image.convert_image_dtype(image_tensor, tensorflow.float32)
# append 1 dims as draw_bounding_boxes expect batches of images
image_tensor = tf.expand_dims(image_tensor, 0)
# draw bounding box and text
image_tensor = image_io.draw_bounding_boxes(image_tensor, [[[0.1, 0.2, 0.5, 0.9]]], ["hello world!"])
# convert float32 back to uint8
image_tensor = tf.image.convert_image_dtype(bb_image_p, tensorflow.uint8)
# remove added dims before
image_tensor = tf.squeeze(image_tensor, [0])
# encode back to png
png_contents = tf.image.encode_png(image_tensor)
# write to file if needed.
with open(filename, "wb") as f:
    f.read(png_contents.eval())
```

PR #255, opened 2019-05-29
Arrow batch_mode fail fast for unsupported options

BryanCutler:
This change checks if the `batch_mode` option is valid on Dataset creation, so that if an unsupported mode is entered it will fail fast. Previously, an unsupported `batch_mode` option would not fail until the op is serialized.  Also expanded documentation of batching with Arrow Datasets.

### DIFF
--- a/tensorflow_io/arrow/python/ops/arrow_dataset_ops.py
+++ b/tensorflow_io/arrow/python/ops/arrow_dataset_ops.py
@@ -153,7 +153,10 @@ class ArrowDataset(ArrowBaseDataset):
       output_types: Tensor dtypes of the output tensors
       output_shapes: TensorShapes of the output tensors or None to
                      infer partial
-      batch_size: Batch size of output tensors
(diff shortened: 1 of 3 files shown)


tensorflow_io/arrow/python/ops/arrow_dataset_ops.py
```diff
@@ -104,6 +106,10 @@ def __init__(self,
         batch_size or 0,
         dtype=dtypes.int64,
         name="batch_size")
+    if batch_mode not in self.batch_modes_supported:
```
BryanCutler: @terrytangyuan I generally like using strings for these kinds of options as opposed to enums, but if the other datasets in tfio end up doing something else for a similar `batch_mode` param, then we can change this so they are consistent. With this change, it will fail fast so the user will know right away if they entered an invalid string.

BryanCutler:
This is a followup to #237

BryanCutler:
Thanks @terrytangyuan !

PR #275, opened 2019-06-06
Change import tensorflow to import tensorflow as tf.

### LINKED ISSUE #274: Use `import tensorflow as tf` in tensorflow-io
In tensorflow-io for legacy reasons we used different ways of importing tensorflow. Some are through non-public api (to make TF 2.0 and 1.x work at the same time). Some used `import tensorflow`, some uses `import tensorflow as tf`.

It is desirable to use `import tensorflow as tf` if application. The reason is that this is the supported tf2_upgrade format. See https://github.com/tensorflow/tensorflow/issues/29380#issuecomment-499289446

This issue is to track the progress in changing `import tensorflow` to `import tensorflow as tf`.

Contributions welcome!

jiachengxu:
Fix #274

### DIFF
--- a/configure.sh
+++ b/configure.sh
@@ -15,7 +15,7 @@
 # ==============================================================================
 
 rm -f .bazelrc
-if python -c "import tensorflow" &> /dev/null; then
+if python -c "import tensorflow as tf" &> /dev/null; then
     echo 'using installed tensorflow'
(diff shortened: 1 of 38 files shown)


jiachengxu:
cc @yongtang

PR #276, opened 2019-06-07
Update Travis to upload each build to Bintray

yongtang:
So that build could be used as release candidate.


yongtang:
This is a plumbing change so that generated whl files could be saved to Bintray. Next time we do a release, we will pull the generated whl files and push it.

Note nightly build still pushed to pypi.org directly. The reason is that pypi.org does not allow override binary so we cannot apply nightly to formal release directly. (nightly build itself is less of a concern as it is only `nightly`).

Merging.

PR #249, opened 2019-05-27
[WIP] Add Tensorflow 1.14.0 support

yongtang:
This PR adds TensorFlow 1.14.0 support.

Signed-off-by: Yong Tang <yong.tang.github@outlook.com>

PR #279, opened 2019-06-07
Update dependency to 1.14.0rc0 and 2.0.0beta

yongtang:
This PR replaces #249 


yongtang:
With this PR, we build with TF 1.14.0rc and 2.0.0beta.

PR #281, opened 2019-06-09
Migrate TextDataset to 2.0 Dataset

yongtang:


yongtang:
All tests passed. Added a `BaseDataset` class as some of the implementations do not use `FileInput` variant (instead go direct to kernel ops of the DatasetOp). So `BaseDataset` (and `Dataset`) in python helps in both cases.

yongtang:
Rebased and all tests passed.

PR #285, opened 2019-06-09
Fix Bintray upload issues

yongtang:


yongtang:
Plumbing change to address the Bintray issue. Merging.

PR #295, opened 2019-06-12
Add batch support for VideoDataset and support reading from streams

### LINKED ISSUE #144: Support decoding video from byte stream for tensorflow-io
Video (ffmpeg) support (#30) is awesome. I was wondering if we can improve the video dataset API by having a counterpart of [`tf.contrib.ffmpeg.decode_video`](https://www.tensorflow.org/api_docs/python/tf/contrib/ffmpeg/decode_video), which decodes the video from the contents bytestream (`tf.string`) rather than a filename.

This can be useful in some use cases where videos are encoded and stored in a TFRecord dataset. A side note: it seems that a couple of [exemplar video datasets](https://github.com/tensorflow/datasets/blob/master/docs/datasets.md#video) stores a sequence of frame images (encoded as PNG) so I'm not 100% sure if this feature would be worthwhile.

I took a quick look at the C++ kernel implementation but the implementation seems quite tricky. One way is, as in `tf.contrib.ffmpeg.decode_video`, we could write the buffer content into a temporary file but I am doubtful this is the right way. Pipe or reading from stdin could work, but needs some study.

yongtang:
This fix adds batch support for VideoDataset and support reading from streams.

This fix also fixes #144.


tests/test_video_eager.py
```diff
@@ -30,7 +30,7 @@
 
 video_path = os.path.join(
     os.path.dirname(os.path.abspath(__file__)), "test_video", "small.mp4")
-
+video_path = "file://" + video_path
```
terrytangyuan: Consider combining with the above two lines?

tests/test_video.py
```diff
@@ -28,10 +28,10 @@
 
 video_path = os.path.join(
     os.path.dirname(os.path.abspath(__file__)), "test_video", "small.mp4")
-
+video_path = "file://" + video_path
```
terrytangyuan: Same here

tensorflow_io/video/python/ops/video_ops.py
```diff
@@ -68,10 +68,10 @@ def load_dependency_and_library(p):
     ],
 })
 
-class VideoDataset(data_ops.BaseDataset):
+class VideoDataset(data_ops.Dataset):
   """A Video File Dataset that reads the video file."""
 
-  def __init__(self, filename):
+  def __init__(self, filename, batch=None):
```
terrytangyuan: Add docstring for this and a note on the relation between dataset's own `batch()` since I think users would definitely be confused about this.

yongtang:
@terrytangyuan The PR has been updated, please take a look.

PR #296, opened 2019-06-12
Enhancement to TextDataset

yongtang:
This is a small enhancement to TextDataset to avoid allocating intermediate
vectors to store data (instead use Tensor).


yongtang:
Rebased and all tests passed. Merging.

PR #297, opened 2019-06-12
Reveting BigTable connector to use batch-bigtable.googleapis.com

vlasenkoalexey:
Most probably right fix would be to upgrade gRPC library from 1.19 to 1.21, unfortunately that breaks build and requires more work. Will look into that later.

### DIFF
--- a/tensorflow_io/bigtable/kernels/bigtable_kernels.cc
+++ b/tensorflow_io/bigtable/kernels/bigtable_kernels.cc
@@ -77,8 +77,7 @@ class BigtableClientOp : public OpKernel {
                   BigtableClientResource** ret) EXCLUSIVE_LOCKS_REQUIRED(mu_) {
                 auto client_options =
                     google::cloud::bigtable::ClientOptions()
-                        .set_connection_pool_size(connection_pool_size_)
-                        .set_data_endpoint("batch-bigtable.googleapis.com");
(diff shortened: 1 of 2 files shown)


yongtang:
@vlasenkoalexey I noticed that the following changes are not in place:
```
                       .set_connection_pool_size(connection_pool_size_)
                         .set_data_endpoint("batch-bigtable.googleapis.com");
```
as there are two commits cancels each other. Just wondering if this is intended?

vlasenkoalexey:
Yes, I'm reverting one of the changes. I got a pushback from BigTable team, they insist that connector should use batch API. As far as I understand I need to update gRPC and all related libraries in order for this to work. Will take a look at this today.

yongtang:
@vlasenkoalexey Thanks. I will merge this PR for now. Let me know if there are anything I can be of help.

vlasenkoalexey:
I see that gRPC version was bumped in google-cloud-cpp recently, but no new stable release was cut out yet: https://github.com/googleapis/google-cloud-cpp/commit/ec767162045961e59504978abdea8629e9299b2f#diff-c7f0cf822a57656a47eca9b634455b0d

Need to wait for release to happen before I update gRPC,  google-cloud-cpp and related libraries in IO.
@sshrdp FYI

PR #300, opened 2019-06-14
Update Arrow Datasets to 2.0 Dataset

BryanCutler:
Update all Arrow Datasets to use the tf.data.Dataset v2 API.

Updated Arrow tests to use eager mode only.

Fixed a bug with Arrow iterator when using batches that have a remainder, that cause an error after the iterator is reached the end.


BryanCutler:
@yongtang @terrytangyuan please take a look when you can, thanks!

Issue #264, opened 2019-06-02
Support for pcap offline network packet format for tensorflow-io


yongtang:
This issue is to track the effort to add offline network packet format (pcap) data for tensorflow-io.

Due to the lack of pcap format support, there are many cases where users are forced to converts pcap files to CSV, XML or JSON format back and force:
- [analyze anomaly in pcap file](https://www.h21lab.com/tools/anomaly-detection), 
- [deep learning for network packet forensics using TensorFlow](https://opensourceforu.com/2016/09/deep-learning-network-packet-forensics-using-tensorflow/)
- [intrusion detection with deep learning](https://towardsdatascience.com/building-an-intrusion-detection-system-using-deep-learning-b9488332b321).

See https://github.com/tensorflow/io/issues/50#issuecomment-496729552 for  additional details.

/cc @ivelin

ivelin:
Thank you, @yongtang . I will keep an eye on this and try to help as I build my confidence with the code base.

p1-martin:
Hi @ivelin, @yongtang 

I would suggest to clarify, if you would like:

- only to decode pcap and to be able to get the raw bytes from the packets
- or the decoding of different protocol layers is requested here 

As seen in the first [example](https://www.h21lab.com/tools/anomaly-detection), the wireshark/tshark is dissecting different protocol layers and only some decoded protocol fields are used to train neural  network. This is useful to train the network with relevant information only, but then it requires protocol decoder (like wireshark/tshark). 

So it is not only about pcap conversion into CSV, XML or JSON. But the pcap is "raw" format compared to XML, JSON which are already protocol decoded format.

Thanks 
Martin

p1-martin:
Hi @ivelin, @yongtang

I would suggest to clarify, if you would like:

only to decode pcap and to be able to get the raw bytes from the packets
or the decoding of different protocol layers is requested here
As seen in the first example, the wireshark/tshark is dissecting different protocol layers and only some decoded protocol fields are used to train neural network. This is useful to train the network with relevant information only, but then it requires protocol decoder (like wireshark/tshark).

So it is not only about pcap conversion into CSV, XML or JSON. But the pcap is "raw" format compared to XML, JSON which are already protocol decoded format.

Thanks
Martin

H21lab:
Hi @ivelin, @yongtang

I would suggest to clarify, if you would like:

only to decode pcap and to be able to get the raw bytes from the packets
or the decoding of different protocol layers is requested here
[...]

PR #303, opened 2019-06-17
Add PcapDataset

ivelin:
@yongtang this is a pull request for a PcapDataset. 
C++ code builds and python unit test passes. 
Hopefully the code is in compliance with best practices for TF IO. Looking forward to your comments.


tests/test_pcap.py
```diff
@@ -0,0 +1,60 @@
...
+        for packets_read in range(1, packets_total):
+            v = sess.run(get_next)
+            next_packet_timestamp = v[0][0]
+            next_packet_data = v[1][0]
+            assert next_packet_timestamp
+            assert next_packet_data
+
+
+if __name__ == "__main__":
+    test.main()
```
yongtang: Also as part of the effort to move to DatasetV2, we want the test to focus on TF 2.0 env. In TF 2.0 env, the biggest difference is that Eager execution is enabled by default. There is no need for `tf.Session` anymore.

Since we still want to support TF 1.x, we test TF 1.x in eager execution (non-default). The following will bring in eager execution in TF 1.x, and is compatible with TF 2.0 in any case:
```
if not (hasattr(tf, "version") and tf.version.VERSION.startswith("2.")):
```

With Eager execution enabled, the iteration of dataset could be simplified to
```
   for v in dataset:
      print(v)
```
And OutOfRange exception is handled automatically.

Overall, here is the test case:
```
"""
Test PcapDataset
"""
from __future__ import absolute_import
from __future__ import division
from __future__ import print_function

import os

import tensorflow as tf
if not (hasattr(tf, "version") and tf.version.VERSION.startswith("2.")):
import tensorflow_io.pcap as pcap_io # pylint: disable=wrong-import-position

def test_pcap_input():
  """test_pcap_input
  """
  print("Testing PcapDataset")
  pcap_filename = os.path.join(
      os.path.dirname(os.path.abspath(__file__)), "test_pcap", "http.pcap")
  file_url = "file://" + pcap_filename
  url_filenames = [file_url]
  dataset = pcap_io.PcapDataset(url_filenames, batch=1)

  packets_total = 0
  for v in dataset:
    (packet_timestamp, packet_data) = v
    if packets_total == 0:
      assert packet_timestamp.numpy()[0] == 1084443427.311224 # we know this is the correct value in the test pcap file
      assert len(packet_data.numpy()[0]) == 62 # we know this is the correct packet data buffer length in the test pcap file
    packets_total += 1
  assert packets_total == 43 # we know this is the correct number of packets in the test pcap file

if __name__ == "__main__":
  test.main()
```

Additional notes:
1: We use `.travis/lint.python.sh` to lint python code:
   ```
curl -o .pylint -sSL https://raw.githubusercontent.com/tensorflow/tensorflow/master/tensorflow/tools/ci_build/pylintrc

find . -name \*.py | xargs pylint --rcfile=.pylint
   ```

2. TF's eager execution does not play well with non-eager mode within the same process. For that reason, our CI will invoke twice. One for files ended with `test_xxx_eager.py` and another one for files not end with `_eager.py`.
    With pcap support DatasetV2, the file name could be changed to `tests/test_pcap_eager.py`.

ivelin:
@yongtang Thank you for the constructive and detailed feedback. I will try to work on the fixes over the weekend between flights.

----
Just a note that this PR is linked to Issue #264 .

ivelin:
Changes applied and committed. In my local environment there seems to be a v1 vs v2 python lib mixup. I am trying to trace the root of this new error in the test case:

```
========================================================================== ERRORS ==========================================================================
___________________________________________________________ ERROR collecting tests/test_pcap.py ____________________________________________________________
tests/test_pcap.py:24: in <module>
    import tensorflow_io.pcap as pcap_io # pylint: disable=wrong-import-position
tensorflow_io/pcap/__init__.py:25: in <module>
    from tensorflow_io.pcap.python.ops.pcap_ops import PcapDataset
tensorflow_io/pcap/python/ops/pcap_ops.py:17: in <module>
    from tensorflow_io.core.python.ops import data_ops as data_ops
tensorflow_io/core/python/ops/data_ops.py:24: in <module>
    class BaseDataset(tf.compat.v2.data.Dataset):
E   AttributeError: module 'tensorflow._api.v1.compat' has no attribute 'v2'
!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!! Interrupted: 1 errors during collection !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
================================================================= 1 error in 1.82 seconds ==================================================================

```

Updated tf to 2.0.beta and rebuilt tf io from source. Didn't help.

yongtang:
@ivelin I think your version of tensorflow is likely `1.13.1`? That is likely the case for `has no attribute 'v2'` error. TensorFlow's 1.13.1 was released too earlier and didn't pick up a change in API.

The DatasetV2 will require tensorflow 1.14.0+ or 2.0.0+.

You can install with pip install `tensorflow==1.14.0rc1`. Once installed, also need to re-invoke
```
./configure.sh
```

to allow bazel picking up the changes in configuration.

Also, you may want to change `tests/test_pcap.py` to `tests/test_pcap_eager.py`.

The reason is that TF's eager execution does not play well with non-eager mode within the same process. For that reason, our CI will invoke twice. One for files ended with test_xxx_eager.py and another one for files not end with _eager.py. Changing `tests/test_pcap.py` to `tests/test_pcap_eager.py` will resolve another issue in Travis CI.

yongtang:
The PR looks good to me and all tests passed. Thanks for the contribution! 👍 🎉

PR #306, opened 2019-06-19
Skip pubsub and prometheus setup on macOS, as they are not supported anyway

yongtang:
Skip pubsub and prometheus setup on macOS, as they are not supported anyway


yongtang:
@terrytangyuan Basically FFmpeg is not supported on macOS. For pubsub and prometheus it is not properly tested on macOS (may or may not work). We could also say they are not supported (yet).

yongtang:
Travis CI seems to have some trouble with rerun of the test. Let me close this PR and create a new PR.

Issue #49, opened 2019-01-12
Support audio format for tensorflow-io


yongtang:
With issue #11 and PR #30, tensorflow-io will have Video format support through FFmpeg. Since FFmpeg supports Audio as well, it is natural to add Audio format support through FFmpeg in tensorflow-io.

This issue captures the necessary effort to support Audio format in FFmpeg. Note that unlike video, many audio format need additional information feed from outside of the container. That means the API augments might be different from video support.

faroit:
@yongtang @terrytangyuan great to see that audio is getting some focus here in tf/io with the recent additions in 0.6.0. Regarding the `WAVDataset` could you please comment on this from the user perspective?

* comparing `WAVInput` and tf2.0 `decode_wav`, which one should one use to read wavs? 
* for a wav reading pipeline, is `WAVDataset` faster than manually mapping `tf.audio.decode_wav`?
* do you plan to add ffmpeg operators?

yongtang:
@faroit `WAVDataset` and `decode_wav` have different use cases but the biggest difference is that, `WAVDataset` is a data pipeline that directly support tf.keras for training and inference. `WAVDataset` will be similar in performance, except `WAVDataset` will performs better with really huge WAV files (e.g., GBs) as `decode_wav` requires disk IO to read the entire file into memory before doing any decoding.

In essence, `WAVDataset` is like `decode_wav` as both are part of the TF's graph. while `decode_wav` is a node (op) in the graph, `WAVDataset` optimize and is really a small subgraph in the graph. `WAVDataset` also simply the interaction with tf.keras.

For FFmpeg operators, we already have support for ffmpeg video operators. The audio operators with ffmpeg should not be very difficult to add (planed to add but haven't had enough time yet with many changes of 1.x -> 2.0).

faroit:
@yongtang thanks for the clarification. 

> WAVDataset will performs better with really huge WAV files (e.g., GBs) as decode_wav requires disk IO to read the entire file into memory before doing any decoding.

does this mean that WAVDataset supports seeking/memory mapping?
Maybe I can elaborate the use case: Imagine you have few but rather long audio recordings (like in [dcase](http://dcase.community/) or music datasets, a `WAVDataset` could give you access to the full decoded file. However, in practice, you would only train with chunked data of a few seconds per sample. The chunking could be done on the fly buy supporting seeking in the actual data loading. If this is not what you meant, I would propose to open an issue, if you agree that this could be a useful addition.

[...]

PR #307, opened 2019-06-19
Add FFmpeg support on Audio streams

yongtang:
This PR adds FFmpeg support on Audio streams. It is not completely finished yet as:
1. Only AV_SAMPLE_FMT_S16 format is supported
2. Additional optimization is needed as currently samples are not read through batchs (one at a time instead).

This PR is part of GitHub issue #49.


yongtang:
To explain the changes:
1. Initially I was thinking we could have to package one for video and one for audio. The video package initially uses FFmpeg. Note FFmpeg is not necessarily the only possible library we use. For audio, WAV was implemented from scratch as it is pretty simple, and does not require FFmpeg.
2. FFmpeg is kind of tricky as it is LGPL. We could not bundle FFmpeg source code directly. Instead we can only link against the pre-installed library by system. For that reason, we only support FFmpeg on Ubuntu 14.04, 16.04, and 18.04. We don't support FFmpeg on macOS as we could not fix a version. We also have to create additional .so files (libtensorflowio_ffmpeg.xxx.so) as otherwise it will not work on systems without FFmpeg installed.
3. When adding AudioDataset with FFmpeg as backend, we have to merge audio and video. That creates issues as we will have multiple copies of FFmpeg.
4. For the above reason, we create a ffpmeg subpackages to only link against Ffmpeg lib  (libtensorflowio_ffmpeg.xxx.so), but includes audio and video.
5. We leave audio package with WAVDataset alone, since some user may not want to use FFMpeg for just WAC files.

PR #314, opened 2019-06-20
Disable pubsub and prometheus on macOS

yongtang:


yongtang:
Very strange. Somehow `sudo -H python -m pip install --upgrade pip setuptools` always give us trouble when we create a PR. But it seems to be fine if running from master...

Will close this PR for now and do some further investigation.

PR #317, opened 2019-06-21
GGFS: GridGain CE file system

dmitrievanthony:
As you might know, `tensorflow-io` currently supports [IGFS](https://ignite.apache.org/features/igfs.html) (Ignite File System, see [here](https://github.com/tensorflow/io/tree/master/tensorflow_io/ignite#igfs)).
As a result of internal discussions, [GridGain CE](https://github.com/gridgain/gridgain) (another open source distribution of Apache Ignite with slightly different licenses) won't support IGFS anymore. To allow GridGain CE users to make checkpointing and store TensorFow training metainformation in it I've prepared another simple file system specifically for TensorFlow. This filesystem is called GGFS (GridGain File System) and repeats API of the IGFS.

This PR is in progress so far. I appreciate any comments.


yongtang:
Thanks @dmitrievanthony. One thing I am thinking is that we could move the .so file into libtensorflowio.so (build from `tensorflow_io/core`) so that we have one place to load the C++ functions. The packages are still in the existing place, only .so files are merged. I already moved some of the libraries into `libtensorflowio.so` (`tensorflow_io/core`), wondering if it makes sense to merge `_ignite_ops.so` with `libtensorflowio.so` as well?

yongtang:
Following are the example:
https://github.com/tensorflow/io/blob/490dfd59df66cabbc54d1c06b207b258d94fca79/tensorflow_io/core/BUILD#L111-L125

tensorflow_io/ignite/kernels/ggfs/ggfs_client.cc
```diff
@@ -0,0 +1,377 @@
...
+          "Method GGFSClient::ListFiles expects strings in response");
+
+    int32_t str_length;
+    TF_RETURN_IF_ERROR(client_->ReadInt(&str_length));
+
+    uint8_t str[str_length];
+    TF_RETURN_IF_ERROR(client_->ReadData(str, str_length));
+
+    out_files->push_back(MakeRelative(
+        string(reinterpret_cast<char *>(str), str_length), path + "/"));
```
avplatonov: are you sure that string c-tor will copy data from "str" to internal array? It don't remember...
because if it won't then data will be corrupted after method end

tensorflow_io/ignite/kernels/ggfs/ggfs_client.cc
```diff
@@ -0,0 +1,377 @@
...
+    uint8_t err_msg_header;
+    TF_RETURN_IF_ERROR(client_->ReadByte(&err_msg_header));
+    if (err_msg_header == kStringVal) {
+      int32_t err_msg_length;
+      TF_RETURN_IF_ERROR(client_->ReadInt(&err_msg_length));
+
+      uint8_t *err_msg_c = new uint8_t[err_msg_length];
+      auto clean = gtl::MakeCleanup([err_msg_c] { delete[] err_msg_c; });
+      TF_RETURN_IF_ERROR(client_->ReadData(err_msg_c, err_msg_length));
+      string err_msg(reinterpret_cast<char *>(err_msg_c), err_msg_length);
```
avplatonov: same as in list files

tensorflow_io/ignite/kernels/ggfs/ggfs.h
```diff
@@ -0,0 +1,62 @@
...
+  Status GetChildren(const string &dir, std::vector<string> *result) override;
+  Status GetMatchingPaths(const string &pattern,
+                          std::vector<string> *results) override;
+  Status DeleteFile(const string &fname) override;
+  Status CreateDir(const string &name) override;
+  Status DeleteDir(const string &name) override;
+  Status GetFileSize(const string &fname, uint64 *size) override;
+  Status RenameFile(const string &src, const string &target) override;
+  Status Stat(const string &fname, FileStatistics *stat) override;
+  string TranslateName(const string &name) const override;
```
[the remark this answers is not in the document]
dmitrievanthony: It's public in the parent class (see [https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/platform/file_system.h#L211](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/platform/file_system.h#L211)).

tensorflow_io/ignite/kernels/ggfs/ggfs_writable_file.h
```diff
@@ -0,0 +1,41 @@
...
+
+class GGFSWritableFile : public WritableFile {
+public:
+  GGFSWritableFile(const string &file_name,
+                   std::unique_ptr<GGFSClient> &&client);
+  ~GGFSWritableFile() override;
+  Status Append(StringPiece data) override;
+  Status Close() override;
+  Status Flush() override;
+  Status Sync() override;
```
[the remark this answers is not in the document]
dmitrievanthony: And again, it's a method from the parent class (see [https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/platform/file_system.h#L332](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/platform/file_system.h#L332)).

tensorflow_io/ignite/kernels/ggfs/ggfs.cc
```diff
@@ -0,0 +1,250 @@
...
+#include "tensorflow_io/ignite/kernels/client/ignite_plain_client.h"
+#include "tensorflow_io/ignite/kernels/client/ignite_ssl_wrapper.h"
+#include "tensorflow_io/ignite/kernels/ggfs/ggfs_random_access_file.h"
+#include "tensorflow_io/ignite/kernels/ggfs/ggfs_writable_file.h"
+#include <queue>
+
+namespace tensorflow {
+
+GGFS::GGFS() {
+  LOG(INFO) << "Call GGFS::GGFS";
```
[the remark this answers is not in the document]
dmitrievanthony: Changed to `LOG(DEBUG)`.

tensorflow_io/ignite/kernels/ggfs/ggfs.cc
```diff
@@ -0,0 +1,250 @@
...
+GGFS::~GGFS() { LOG(INFO) << "Call GGFS::~GGFS"; }
+
+Status GGFS::NewRandomAccessFile(const string &file_name,
+                                 std::unique_ptr<RandomAccessFile> *result) {
+  LOG(INFO) << "Call GGFS::NewRandomAccessFile [file_name = " << file_name
+            << "]";
+
+  result->reset(new GGFSRandomAccessFile(
+      TranslateName(file_name),
+      std::unique_ptr<GGFSClient>(new GGFSClient(
```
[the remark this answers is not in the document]
dmitrievanthony: `GGFSClient` opens a new connection for each request-response anyway, so we won't win anything sharing it. At the same time from the perspective of simplicity, it's easier to create a new client for each operation.

tensorflow_io/ignite/kernels/ggfs/ggfs.cc
```diff
@@ -0,0 +1,250 @@
...
+}
+
+GGFS::~GGFS() { LOG(INFO) << "Call GGFS::~GGFS"; }
+
+Status GGFS::NewRandomAccessFile(const string &file_name,
+                                 std::unique_ptr<RandomAccessFile> *result) {
+  LOG(INFO) << "Call GGFS::NewRandomAccessFile [file_name = " << file_name
+            << "]";
+
+  result->reset(new GGFSRandomAccessFile(
```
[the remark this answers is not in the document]
dmitrievanthony: It would be weird, wouldn't it? I guess the code that calls this method should be responsible for providing a not-null pointer. At least, in other places in such cases, we don't check it.

tensorflow_io/ignite/kernels/ggfs/ggfs_client.h
```diff
@@ -0,0 +1,79 @@
...
+  const string password_;
+
+  Status EstablishConnection();
+  Status Handshake();
+
+  Status SendCommonRequestHeader(uint8_t method_id, int32_t length);
+  Status ReceiveCommonResponseHeader();
+};
+
+constexpr uint8_t kWriteFileMethodId = 0;
```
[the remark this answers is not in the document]
dmitrievanthony: It means we use Google code style, see [https://google.github.io/styleguide/cppguide.html#Constant_Names](https://google.github.io/styleguide/cppguide.html#Constant_Names).

tensorflow_io/ignite/kernels/ggfs/ggfs_client.cc
dmitrievanthony: Yep, I checked (see [here](http://www.cplusplus.com/reference/string/string/string/)).

```
(5) from buffer
    Copies the first n characters from the array of characters pointed by s.
```

tensorflow_io/ignite/kernels/ggfs/ggfs.cc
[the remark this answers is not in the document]
dmitrievanthony: Changed back to `LOG(INFO)`, looks like there is no `DEBUG` level available.

dmitrievanthony:
> Thanks @dmitrievanthony. One thing I am thinking is that we could move the .so file into libtensorflowio.so (build from `tensorflow_io/core`) so that we have one place to load the C++ functions. The packages are still in the existing place, only .so files are merged. I already moved some of the libraries into `libtensorflowio.so` (`tensorflow_io/core`), wondering if it makes sense to merge `_ignite_ops.so` with `libtensorflowio.so` as well?

Hi, @yongtang. I updated the config and moved libraries into `libtensorflow.so`. I hope I did it correctly, please have a look.

tensorflow_io/ignite/kernels/ggfs/ggfs_client.cc
dmitrievanthony: See my reply above.

tensorflow_io/ignite/kernels/ggfs/ggfs_client.cc
```diff
@@ -0,0 +1,377 @@
...
+  TF_RETURN_IF_ERROR(client_->WriteByte(append));
+  TF_RETURN_IF_ERROR(client_->WriteByte(kByteArrayVal));
+  TF_RETURN_IF_ERROR(client_->WriteInt(length));
+  TF_RETURN_IF_ERROR(client_->WriteData(data, length));
+  TF_RETURN_IF_ERROR(ReceiveCommonResponseHeader());
+
+  return Status::OK();
+}
+
+Status GGFSClient::ReadFile(string path, uint8_t **out_data,
```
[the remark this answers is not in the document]
dmitrievanthony: Yep, great idea. I rewrote this part of the code and use `shared_ptr` now. Please have a look.

dmitrievanthony:
Hi, @yongtang. Looks like all tests pass now so could you please have a look?

dmitrievanthony:
BTW, what do you think about an idea to create base tests for all file systems? It seems all tests should be the same for all file system implementations, the only difference in a configuration and file prefixes.

tensorflow_io/ignite/kernels/ggfs/ggfs_client.cc
```diff
@@ -0,0 +1,381 @@
...
+
+  if (status != 0) {
+    uint8_t err_msg_header;
+    TF_RETURN_IF_ERROR(client_->ReadByte(&err_msg_header));
+    if (err_msg_header == kStringVal) {
+      int32_t err_msg_length;
+      TF_RETURN_IF_ERROR(client_->ReadInt(&err_msg_length));
+
+      uint8_t *err_msg_c = new uint8_t[err_msg_length];
+      auto clean = gtl::MakeCleanup([err_msg_c] { delete[] err_msg_c; });
```
yongtang: It might be OK though I am wondering if it is possible to convert
```
uint8_t *err_msg_c = new uint8_t[err_msg_length];
```

with smart pointers? `auto clean` releases the memory so not concern here.

But in the future if someone else touches this part of the code, having allocation and auto-release in one line with smart pointer, might be "safer" than having two lines of explicit allocate then auto-release.

tensorflow_io/ignite/kernels/ggfs/ggfs_client.cc
```diff
@@ -0,0 +1,381 @@
...
+    TF_RETURN_IF_ERROR(client_->ReadShort(&serv_ver_patch));
+    uint8_t header;
+    TF_RETURN_IF_ERROR(client_->ReadByte(&header));
+
+    if (header == kStringVal) {
+      int32_t length;
+      TF_RETURN_IF_ERROR(client_->ReadInt(&length));
+
+      uint8_t *err_msg_c = new uint8_t[length];
+      auto clean = gtl::MakeCleanup([err_msg_c] { delete[] err_msg_c; });
```
yongtang: Same here, possible with smart pointer?

yongtang:
Thanks @dmitrievanthony for the PR update. Added a couple of comments about alloc/release, but otherwise the PR looks nice to me.

I think creating base tests for all file systems would be an excellent idea. We have igfs/ggfs and css for alicloud for now, and azure storage file system is upcoming. So a common base test will definite help.

Maybe we could place igfs/ggfs base tests in `tensorflow_io/core`? In this way future file system additions could follow the same test pattern and cases.

tensorflow_io/ignite/kernels/ggfs/ggfs_client.cc
dmitrievanthony: Yes, it's a good point. I updated the code, now it uses smart pointers.

tensorflow_io/ignite/kernels/ggfs/ggfs_client.cc
dmitrievanthony: Done.

dmitrievanthony:
I'm going to reorganize tests in a separate task, so speaking about this PR, do you have any other concerns?

tensorflow_io/ignite/kernels/ggfs/ggfs_client.cc
```diff
@@ -267,10 +266,9 @@ Status GGFSClient::ReceiveCommonResponseHeader() {
       int32_t err_msg_length;
       TF_RETURN_IF_ERROR(client_->ReadInt(&err_msg_length));
 
-      uint8_t *err_msg_c = new uint8_t[err_msg_length];
-      auto clean = gtl::MakeCleanup([err_msg_c] { delete[] err_msg_c; });
-      TF_RETURN_IF_ERROR(client_->ReadData(err_msg_c, err_msg_length));
-      string err_msg(reinterpret_cast<char *>(err_msg_c), err_msg_length);
+      std::unique_ptr<uint8_t> err_msg_c = std::unique_ptr<uint8_t>(new uint8_t[err_msg_length]);
```
yongtang: @dmitrievanthony wondering if this should be `std::unique_ptr<uint8_t> err_msg_c(new uint8_t[err_msg_length])`, or a customized deleter?

I know some compilers will automatically call `delete []` even if `delete` is called for array, e.g., `uint_8 arr=new uint8[10]; delete arr;`. However, we are still using Ubuntu 14.04's default GCC (4.8) so it may or may not be able to call `delete []` automatically. Think we could explicitly calling the deleter for array?

dmitrievanthony:
Hi, @yongtang. Speaking about smart pointers I checked that `std::unique_ptr<uint8_t[]>` should work fine (see https://stackoverflow.com/questions/21377360/proper-way-to-create-unique-ptr-that-holds-an-allocated-array/21377424). So, I think it's fine, I merged this PR.

yongtang:
@dmitrievanthony Thanks for the update :+1

PR #321, opened 2019-06-24
Improvements to ArrowStreamDataset

BryanCutler:
This makes the following improvements to the `ArrowStreamDataset`:

* Adds the option to use a Unix Domain Socket, with the benefit of using simple file permission to create a secure local socket

* Allow the `ArrowStreamDataset` to accept multiple hostnames for input

* Adds the ability to create a `ArrowStreamDataset` from a Pandas DataFrame where slices are streamed to the dataset over a local socket

### DIFF
--- a/tensorflow_io/arrow/README.md
+++ b/tensorflow_io/arrow/README.md
@@ -76,20 +76,24 @@ a given `pyarrow.Schema`, e.g. `dataset = ArrowFeatherDataset.from_schema(filena
 
 ## From a Stream of Arrow Record Batches
 
-The `ArrowStreamDataset` provides a Dataset that will connect to a host over
-a socket that is serving Arrow record batches in the Arrow stream format. See
-[here](https://arrow.apache.org/docs/python/ipc.html#writing-and-reading-streams)
-for more on the stream format. The following example will create an
-`ArrowStreamDataset` that will connect to a host that is serving an Arrow
-stream of record batches with 2 columns of dtypes=(int32, float32):
+The `ArrowStreamDataset` provides a Dataset that will connect to one or more
+hosts over a socket that is serving Arrow record batches in the Arrow stream
+format. See [here](https://arrow.apache.org/docs/python/ipc.html#writing-and-reading-streams)
+for more on the stream format. Supported socket families are `AF_INET` and
+`AF_UNIX`, specified by the `host_type` parameter (default is `AF_INET`). The
+following example will create an `ArrowStreamDataset` that will connect to a
+host that is serving an Arrow stream of record batches with 2 columns of
+dtypes=(int32, float32):
 
 ```python
 import tensorflow as tf
 from tensorflow_io.arrow import ArrowStreamDataset
 
-# The str `host` should be in the format '<HOSTNAME>:<PORT>'
+# The parameter `hosts` can be a Python string or a list of strings and should
+# be in the format '<HOSTNAME>:<PORT>' for `host_type` of `AF_INET` or path
+# for `host_type` of `AF_UNIX`
 dataset = ArrowStreamDataset(
-    host,
+    hosts,
     columns=(0, 1),
     output_types=(tf.int32, tf.float32),
     output_shapes=([], []))
--- a/tensorflow_io/arrow/README.md
+++ b/tensorflow_io/arrow/README.md
@@ -77,23 +77,26 @@ a given `pyarrow.Schema`, e.g. `dataset = ArrowFeatherDataset.from_schema(filena
 ## From a Stream of Arrow Record Batches
 
 The `ArrowStreamDataset` provides a Dataset that will connect to one or more
-hosts over a socket that is serving Arrow record batches in the Arrow stream
+endpoints that are serving Arrow record batches in the Arrow stream
 format. See [here](https://arrow.apache.org/docs/python/ipc.html#writing-and-reading-streams)
-for more on the stream format. Supported socket families are `AF_INET` and
-`AF_UNIX`, specified by the `host_type` parameter (default is `AF_INET`). The
-following example will create an `ArrowStreamDataset` that will connect to a
-host that is serving an Arrow stream of record batches with 2 columns of
-dtypes=(int32, float32):
+for more on the stream format. Currently supported endpoints are a POSIX IPv4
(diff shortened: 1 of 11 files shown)


tensorflow_io/arrow/kernels/arrow_dataset_ops.cc
```diff
@@ -246,9 +246,9 @@ class ArrowConvertTensor : public arrow::ArrayVisitor {
  private:
   int64_t curr_index_;
   int64_t curr_batch_size_;
+  IteratorContext* curr_ctx_;
   int32_t curr_array_length_;
   DataType curr_type_;
-  IteratorContext* curr_ctx_;
```
BryanCutler: reorder to fix warning

tensorflow_io/arrow/kernels/arrow_dataset_ops.cc
```diff
@@ -293,7 +293,7 @@ class ArrowDatasetBase : public DatasetBase {
 
       // If in initial state, setup and read first batch
       if (current_batch_ == nullptr && current_row_idx_ == 0) {
-        TF_RETURN_IF_ERROR(SetupStreamsLocked(ctx->env()));
```
BryanCutler: `Env` wasn't being used, so removed it in the overrides

tensorflow_io/arrow/kernels/arrow_dataset_ops.cc
```diff
@@ -871,13 +864,51 @@ class ArrowFeatherDatasetOp : public ArrowOpKernelBase {
...
+    *host_type = ArrowHostType::AF_INET;
+  } else if (host_type_str == "AF_UNIX") {
+    *host_type = ArrowHostType::AF_UNIX;
+  } else if (host_type_str == "STDIN") {
+    *host_type = ArrowHostType::STDIN;
+  } else {
+    return errors::Internal("Unsupported host type: " + host_type_str);
+  }
+  return Status::OK();
+}
```
BryanCutler: This enum conversion is kind of messy, it could be improved in the future I think.

tensorflow_io/arrow/kernels/arrow_stream_client.h
```diff
@@ -20,10 +20,16 @@ limitations under the License.
 
 namespace tensorflow {
 
+// Supported Address Families
+enum ArrowStreamFamily {
+  AF_INET_SOCKET,
+  AF_UNIX_SOCKET,  // For UNIX systems only
```
BryanCutler: This is a little different than the `host_type` enum because we are specifically talking about sockets here

BryanCutler:
cc @yongtang @terrytangyuan to please review, thanks!

tensorflow_io/arrow/README.md
```diff
@@ -76,20 +76,24 @@ a given `pyarrow.Schema`, e.g. `dataset = ArrowFeatherDataset.from_schema(filena
...
 import tensorflow as tf
 from tensorflow_io.arrow import ArrowStreamDataset
 
-# The str `host` should be in the format '<HOSTNAME>:<PORT>'
+# The parameter `hosts` can be a Python string or a list of strings and should
+# be in the format '<HOSTNAME>:<PORT>' for `host_type` of `AF_INET` or path
+# for `host_type` of `AF_UNIX`
 dataset = ArrowStreamDataset(
-    host,
+    hosts,
```
yongtang: Wondering if it makes sense to change it from `host` -> `endpoints` (or `endponit`)? as other types like stdin are not truly host.

BryanCutler: Yeah, that does make more sense. I'll update it, thanks!

tensorflow_io/arrow/python/ops/arrow_dataset_ops.py
```diff
@@ -357,19 +364,21 @@ class ArrowStreamDataset(ArrowBaseDataset):
...
   def __init__(self,
-               host,
+               hosts,
                columns,
                output_types,
                output_shapes=None,
                batch_size=None,
-               batch_mode='keep_remainder'):
+               batch_mode='keep_remainder',
+               host_type='AF_INET'):
```
yongtang: For python I think we want to avoid introduce args if possible. For unix sockets normally they are specified with scheme prefix `unix://<socket_address>`. For inet socket they are specified with `tcp://<host>:<port>`.

For standard input, I vaguely remember it used to be `fd://<number>`? (and pipe used to be `pipe://<address>`)

To make it easier (and compatible) with previous definitions, I think we also could default the scheme as `tcp://` (so that `host:port` will be interpreted as `tcp://host:port`.

For that I think maybe we could have the following API for hosts (or `endpoints` as was mentioned in another comment):
```
endpoints: A `tf.string` tensor, Python list or scalar string defining the  input stream.
    endpoints` could have the following format:,
      - `host:port`: IPv4 address (default)
      - `tcp://host:port`: IPv4 address,
      - `unix://<unix socket address>`: local path as unix socket address,
      - `fd://<number>`: STDIN or file descriptor number.
```
```

BryanCutler: great idea, it's best to use a standard and not introduce unnecessary args. +1 on the above API, I saw you used `fd://-` for stdin on another PR.

yongtang: I am a little unsure about stdin, as there are many ways to specify it.

In some software, `-` are used in `command line` to specify `stdin`. e.g
- `docker import -`
- `kubectl apply -f -`
- `tshark -r -`.

In some other software, `/dev/stdin` are used as stdin is essentially still a file on Linux. The issue with `file://` is that, in TF `file://` stands for posix file system, and the file specified by `file://` is `openable` and `closable`. But this is not the case for stdin. So we have to use another sheme for stdin than `file://`

I think we could either treat stdin as a generic case of `fd` (so `fd://0`) or a special case of `fd` (`fd://-`). Or support both?

BryanCutler: >I think we could either treat stdin as a generic case of fd (so fd://0) or a special case of fd (fd://-). Or support both?

It shouldn't be too hard to support both, I'll give it a try

BryanCutler:
@yongtang I fixed up the `endpoints` parameter like we discussed, please take another look when you can. Thanks!

BryanCutler:
Thanks @yongtang !

PR #322, opened 2019-06-26
Ignore font for github linguist

yongtang:
At the moment the language of the repo was incorrectly
labeled as C. That was because Font binary code was
generated. It should not be part of the source code.

This fix excludes font binary code.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+font_*.h linguist-generated=true


yongtang:
@terrytangyuan Language has been correctly changed back to C++, which is much better for contributors. Still would prefer to see Python as the default language (currently C++ 59.6% vs. Python 31.6%).

terrytangyuan:
Good to know! Yea currently C++ code is still the majority.

PR #324, opened 2019-06-28
Add configuration for kafka dataset

zhjunqin:
Add configuration for kafka dataset, including 'Global configuration properties' and 'Topic configuration properties'

Configuration refer to https://github.com/edenhill/librdkafka/blob/v0.11.6/CONFIGURATION.md .

### DIFF
--- a/tensorflow_io/kafka/python/ops/kafka_dataset_ops.py
+++ b/tensorflow_io/kafka/python/ops/kafka_dataset_ops.py
@@ -32,7 +32,9 @@ class KafkaDataset(data.Dataset):
                servers="localhost",
                group="",
                eof=False,
-               timeout=1000):
+               timeout=1000,
+               config_global=None,
+               config_topic=None):
     """Create a KafkaReader.
 
     Args:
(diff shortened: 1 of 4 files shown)


tensorflow_io/kafka/python/ops/kafka_dataset_ops.py
```diff
@@ -44,6 +46,17 @@ def __init__(self,
...
+                  properties in [Key=Value] format,
+                  eg. ["enable.auto.commit=false",
+                       "heartbeat.interval.ms=2000"],
+                  please refer to 'Global configuration properties'
+                  in librdkafka doc.
+      topic_cfg: A `tf.string` tensor containing topic configuration
+                 properties in [Key=Value] format,
+                 eg. ["auto.offset.reset=earliest"],
+                 please refer to 'Topic configuration properties'
+                 in librdkafka doc.
```
yongtang: @zhjunqin Maybe it makes sense to change the arg names to `config_global` and `config_topic`, so that they could be separated from the configuration parameters that belongs to generic Dataset?

yongtang:
@zhjunqin Thanks for the PR! Overall it looks good, maybe the name of the parameters could be changed?

tensorflow_io/kafka/python/ops/kafka_dataset_ops.py
zhjunqin: OK, I will change the name. : )

Issue #315, opened 2019-06-20
Standardize columnized dataset?


yongtang:
With the upcoming DatasetV2 a lot of the APIs are getting simplified. That also opens up some additional possibilities than just passing the dataset to tf.keras.

One area of interest, is that we already have support for many columnized dataset, e.g, Arrow, Avro, Parquet, Json, HDF5, etc. Those dataset may potentially be standardized with the same API so that we could treat them homogeneously. For example, ArrowDataset already exposes a `columns()` property method. We could apply the same to Avro, Parquet, Json, HDF5 etc. Thought?

Since those columnized dataset are largely numeric values, I think one area we also could have a common base class for those dataset, and support additional operations. For example, dataset_1 + dataset_2 => dataset_3 (add) where dataset_3 could be passed to tf.keras.  The implementation could start with zip + map in python (not even needed in C++). Maybe this could be one use case that will help users?

/cc @terrytangyuan @BryanCutler

BryanCutler:
Sounds like a good idea to me. I think it would be pretty useful to support operations and making composite columnar datasets, like merging 2 different datasets into 1.

terrytangyuan:
Yes, agreed that this will be helpful to users. "The implementation could start with zip + map in python (not even needed in C++)" - I am worried about the performance and the efforts spent on the implementation of each additional operations though. Is there a way to reuse what's available in pandas/pyspark, etc.?

yongtang:
In order to make dataset more useful we also need to support write (save) so that a dataset could be saved in certain format (e.g., Parquet, Arrow, HDF5, CSV, JSON, etc).

I have created an experimental PR #326 to add TextDataset's save support. (Will expand to other dataset format later).

yongtang:
@BryanCutler @terrytangyuan With recent PRs I am writing down some notes about **Data Formats**:

There are many data formats available and it is hard to fit them all into one types. From data science point of view pandas could be a good reference though I noticed in some edge cases even pandas is not handling very well. Here is the list I come up with, I am also seeking some suggestions on certain cases.

1) **Standalone n-dimensional array**:

    This is the type that is commonly seen with `numpy`'s `array`, it is also referred as a `tensor` in tensorflow, and `dataset` in HDF5.

    Note in pandas, HDF5 is considered as columnar/tabular format. But that is largely an "overuse" and may caused issues as HDF5 datasets could be totally unrelated.

    While this format normally fits into memory, there are cases where we really want to just take a slice of it from the imported file.
[...]

PR #326, opened 2019-07-02
Add save method for TextDataset

yongtang:
This is an experimental implementation of the `save` (write) method for TextDataset in order to allow saving the content of Dataset into files.

TextDataset itself is straightforward so it may not be that useful. However, it setup an example for saving to other future format (e.g., Parquet/Arrow/etc).

This PR is related to #315

### DIFF
--- a/tests/test_text_eager.py
+++ b/tests/test_text_eager.py
@@ -110,22 +110,20 @@ def test_text_output():
     lines = [line.strip() for line in f]
   text_filename = "file://" + text_filename
 
-  text_dataset = text_io.TextDataset([text_filename, text_filename], batch=2)
-
-  text_dataset = text_dataset.map(lambda x: tf.strings.join([x[0], x[1]]))
-
   f, filename = tempfile.mkstemp()
   os.close(f)
 
-  text_io.TextDataset.save(text_dataset, filename)
+  df = text_io.TextDataset(text_filename)
+  df = df.take(5)
+  text_io.save_text(df, filename)
 
   with open(filename, 'rb') as f:
     saved_lines = [line.strip() for line in f]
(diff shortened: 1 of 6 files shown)


tests/test_text_eager.py
```diff
@@ -102,5 +102,30 @@ def on_predict_batch_end(self, batch, logs=None):
...
+  text_io.TextDataset.save(text_dataset, filename)
+
+  with open(filename, 'rb') as f:
+    saved_lines = [line.strip() for line in f]
+  for i, _ in enumerate(saved_lines):
+    assert saved_lines[i] == lines[
+        (2 * i) % len(lines)] + lines[
+            (2 * i + 1) % len(lines)]
+  assert len(saved_lines) == len(lines)
+
```
BryanCutler: does the file need to be deleted after the test?

yongtang:
@BryanCutler @terrytangyuan @feihugis Overall the code is ready for review. The usage is similar with pandas (except pandas does not have read_text):
```
import tensorflow_io.frame as frame_io

# Write first 5 (head) entries to dest_path

df = frame_io.read_text(source_path)
df.head().to_text(dest_path)
```

If this is OK I am planning to expand the similar usage to more interesting formats such as csv/parquet.

(The file lock issue has not been resolved though this will be done in follow up PRs).

jiachengxu:
This is a really good example of implementing `save` method, I think I can use the same idea for the JsonDataset

yongtang:
@jiachengxu In PR  #351 I made some enhancement to utilize C++ template. It should be even easier to implement (only need to write a class with `Status Write(WritableFile *file, std::vector<Tensor>& components)` method.

PR #329, opened 2019-07-08
fixing read-write test on linux

damienpontifex:
`std::ostringstream::rdbuf::pubsetbuf` doesn't seem to be implemented on macOS (my dev machine) so I had rembed the `pubsetbuf` but it seems to fix things with linux tests.

Also removed the azure cpp lib patch which isn't needed anymore


damienpontifex:
@yongtang this fix seems to have all tests working on linux. One of the macOS builds is timeing out, but the macOS 2.0 passes. I'm not sure what we could do there to resolve?

yongtang:
@damienpontifex I just restarted the test and all passed. LGTM and thanks for the fix!

PR #331, opened 2019-07-09
specify smaller cc_library targets as link_static

damienpontifex:
I noticed the results of cc_library targets in azfs all produce so libraries that end up getting copied on release. Simplifying this so there is only one file in azure products copied to make it clearer this is the only one that's needed


yongtang:
All tests pass except macOS which hits the execution timeout (50 min) on Travis CI. merging now. A follow up PR #333 will try to reduce the Travis CI execution time on macOS.

PR #332, opened 2019-07-09
Change from libtensorflowio.so to libtensorflow_io.so

yongtang:
This fix change the .so name from libtensorflowio.so to libtensorflow_io.so,
this is to match tensorflow's setup of libtensorflow_core.so and
libtensorflow_framework.so


yongtang:
On macOS, Travis CI hits the max execution time (50 min) which will be taken care of in PR #333. Merging now.

PR #334, opened 2019-07-09
Move _azfs.so to libtensorflow_io.so to reduce package size

yongtang:
The change is inline with other file system plugins (e.g., ignite)

also /cc @damienpontifex 


damienpontifex:
@yongtang does this change anything for the usage guide here https://github.com/tensorflow/io/blob/master/tensorflow_io/azure/README.md ?

yongtang:
@damienpontifex nothing changes from usage. As you could see the tests are not touched so they remains the same.

PR #336, opened 2019-07-10
Azure blob storage guide

damienpontifex:
- Fixed python import path
- Added information about environment variables for configuration


tensorflow_io/azure/README.md
```diff
@@ -1,14 +1,19 @@
...
+- `TF_AZURE_STORAGE_USE_HTTP`  
+  Set to any value if you don't want to use https transfer. `unset` to use default of https 
+- `TF_AZURE_STORAGE_BLOB_ENDPOINT`  
+  Set to the endpoint of blob storage - default is .core.windows.net
 
 ```python
 >>> import os
 >>> import tensorflow as tf
->>> import tensorflow_io.azure.python.ops.azfs_ops
+>>> import tensorflow_io.azure.azfs_ops
```
yongtang: @damienpontifex I think we could move one step further, by moving `azfs_ops.py` to `__init__.py` (under the same directory). Then we could use:
```
import tensorflow_io.azure
```

yongtang:
@damienpontifex Think test_azure.py should also change `import tensorflow_io.azure.azfs_ops` to `import tensorflow_io.azure`

Issue #262, opened 2019-05-31
[GSoC] JSON Reader for tensorflow-io


jiachengxu:
Hi folks,
I am Jiacheng Xu, and I am working on this feature during this summer as my Google Summer of Code project, here is a description from my mentor @yongtang: 

With tensorflow and tensorflow-io, a lot of file formats have been well covered. One format which is used everywhere but has not been supported yet, is the json format file. In tensorflow, it is certainly possible to read json files with tf.py_func and pass it through graph or tf.data. There are some limitations with this approach as was mentioned in the tensorflow docs.

We could actually build the json format parser support in, and selectively pass the values to tf.data.

A C++ library of json parser might be needed. The jsoncpp could be a choice, other options are fine as long as the library is in good quality. 

Refs:
- [Project website](https://summerofcode.withgoogle.com/projects/?sp-search=jiacheng#5823672818860032)
- [Proposal of the project](https://docs.google.com/document/d/1yCB41MLIDS0OUgKhv9xP8XXHHctFGTkppaQKsHm0r9g/edit?usp=sharing)
- [Benchmark about different json libraries](https://github.com/mloskot/json_benchmark)

jiachengxu:
Hi @yongtang, can you try to assign this issue to me?

BryanCutler:
Hi @jiachengxu , welcome!  This will be a nice addition, I'm looking forward to the PR. Thanks!

jiachengxu:
Hi @BryanCutler, thanks for your warm welcome, I am also looking forward to contributing to the project!

yongtang:
@jiachengxu The C++ code base in tensorflow-io has a pretty deep learning curve.

To get started quickly, I would suggest you to:
1. Start with a test for JSON format in python. You can look into tests/test_text.py (or other test for examples).
   In tensorflow-io we typically have two types of tests. The first is the integration with tf.keras so that dataset could be passed for training and inference directly. The second is the iteration of tf.data.Dataset itself.
2. Implement in python a JSON parser (no need for jsoncpp for now), and pass it through `tf.data.Dataset.from_tensor_slices`. This will help you pass the test implemented in 1). `from_tensor_slices` has limitations but that could be an easy step to make progress.
3. Finally we could work on get jsoncpp integrated. The tests in 1) could be reused.

Let us know if you run into any issues.

jiachengxu:
Hi @yongtang, thanks for your information, and I will take a look at the `test` part and try to add some tests first.
[...]

PR #337, opened 2019-07-10
[GSoC-Phase2] Add JSON Support

jiachengxu:
This PR contains the work of Phase 2 of my GSoC project #262:
- Implement the JSON Dataset support in C++ based on `jsoncpp`.
- Rewrite the test in `test_json.py` to make it more generic.
- Add a test of using `JSONDataset` with `tf.keras` based on [TensorFlow Custom training: walkthrough](https://www.tensorflow.org/tutorials/eager/custom_training_walkthrough).

Limition:
- In the current implementation, The `filename()` is called to get the filenames, and this will not work 
in file systems with scheme (e.g., `s3://`, `gcs://`, `azfs://` etc).

Next Step:
-  Get scheme file system (e..g, s3:// or gcs://) working.
- Add documentation/tutorial of JSON Dataset support.
- Implement `write_to_json` method.

/cc @yongtang @terrytangyuan


yongtang:
@jiachengxu to help speed up your development, inside 
```c++
Status ReadRecord(io::InputStreamInterface* s, ..
```

you could call `filename()` to get the filename. Then you can try just process like a regular file with jsoncpp. 

The above will not work in file systems with scheme (e.g., `s3://`, `gcs://`, `azfs://` etc) but let's start getting basic workflow done before next step.

jiachengxu:
Hi @yongtang, I have updated the PR with `ReadRecord` implemented, PTAL, and I am wondering if there is a way to only compile the JSON dataset source code for testing and debugging instead of compiling all the dataset support?

Thanks,
Jiacheng

tensorflow_io/json/kernels/json_input.cc
```diff
@@ -0,0 +1,130 @@
...
+ public:
+  Status ReadRecord(io::InputStreamInterface* s, IteratorContext* ctx, std::unique_ptr<io::BufferedInputStream>& state, int64 record_to_read, int64* record_read, std::vector<Tensor>* out_tensors) const override {
+    if (state.get() == nullptr) {
+      state.reset(new io::BufferedInputStream(s, 4096));
+    }
+
+    // Read the json file
+    ifstream ifs(filename());
+    Json::Reader reader;
+    Json::Value records;
```
yongtang: @jiachengxu In this case, your `state` consists of ifstream, Json::Reader, and Json::Value, you should wrap them into one class, and use the newly defined class as part of the template.

`class JSONInput: public FileInput<io:BufferedInputStream> ` => `class JSONInput: public FileInput<YourNewClassWithJsonReaderValueAndIfstream> `

yongtang:
@jiachengxu Also take a look at pandas' read_json with `records` type:
https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.read_json.html

This will be the one we try to match.

tensorflow_io/json/kernels/json_input.cc
jiachengxu: Ok, I see, will update the PR now

yongtang:
@jiachengxu Can you try to fix the test failures? Let me know if you have any questions. I am hoping this PR could be merged first. Then we can move to get scheme file system (e..g, `s3://` or `gcs://`) working in the follow up PR.

jiachengxu:
PTAL, @yongtang @terrytangyuan, I wanted to request you guys for a review, but I don't think I have the permission of doing that :).

jiachengxu:
Hi @yongtang, as we discussed in the weekly gsoc meeting today, a test of `JSONDataset` with `tf.keras` is necessary. 

I added a test of that in the latest commits. The dataset and the `tf.keras` model are quite similar to this [TensorFlow Custom training: walkthrough](https://www.tensorflow.org/tutorials/eager/custom_training_walkthrough) on the official TensorFlow website. I also use the same dataset but in JSON format for testing.

PR #345, opened 2019-07-12
Fix error with ArrowStreamDataset.from_pandas with batch_size None

BryanCutler:
This fixes an error when using the ArrowStreamDataset.from_pandas with a batch_size of None. The error raised was `TypeError: 'NoneType' object cannot be interpreted as an integer` due to using the batch_size of None as the step argument in range.


BryanCutler:
cc @yongtang I missed this case with the last patch but hopefully we can get this in before the release

BryanCutler:
All tests passed, merging

PR #348, opened 2019-07-13
Move dropbox to a subfolder so that it could be shared

yongtang:


yongtang:
This is a plumbing change so that dropbox folders could be shared among maintainers.

PR #350, opened 2019-07-13
Add rebatch method for Dataset

yongtang:
This PR adds rebatch method for Dataset where `rebatch(n) = unbatch().batch(n)`

The motivation for rebatch is that there are situations we read the data in
big batches but then we want to adjust the batch size to fit differnet
scenarios.

### DIFF
--- a/tensorflow_io/core/BUILD
+++ b/tensorflow_io/core/BUILD
@@ -42,6 +42,22 @@ cc_library(
 )
 
 cc_library(
+    name = "core_ops",
+    srcs = [
+        "kernels/rebatch_dataset_op.cc",
+        "ops/core_ops.cc",
+    ],
+    copts = tf_io_copts(),
+    includes = [
+        ".",
+    ],
+    linkstatic = True,
+    deps = [
+        ":dataset_ops",
+    ],
+)
+
+cc_library(
     name = "ffmpeg_3.4",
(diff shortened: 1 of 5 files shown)


yongtang:
@BryanCutler @terrytangyuan some additional discussion about the `batch` field.

While batch is essential in tf.keras, it does not necessarily has to be fixed at the beginning. In tf.data.Dataset, it expect each elements is read one by one which severely limits the performance.

We try to improve the performance by reading a chunk (or batch) to cache the I/O. However, later on at the time a tf.data.Dataset is passed to tf.keras, it may have different preferences on batch size.

In `tf.data.Dataset`, we could use `dataset.unbatch().batch(n)` but this again drags down performance as it becomes element wise in graph pipeline.

Also, we discussed different batch mode in the past. Thinking again I think it makes more sense to have it the other way:
- During the dataset read: Use hidden parameter or "natural boundary" to read dataset with a natural batch.
- Do some pre-processing on batches (not necessarily even equal)
- During the tf.keras training: calling rebatch(batch_size, batch_mode) to adjust the batch and pass it to tf.keras.

@terrytangyuan @BryanCutler any ideas or suggestions?

yongtang:
Close this PR for now. I think I might have some other ways to get around the batch issue.

PR #356, opened 2019-07-16
[PoC] Add experimental pandas like API

yongtang:
This PR adds experimental pandas like API so that it is possible to manipulate features in column data way


yongtang:
Following is an example
```python
import tensorflow as tf
import tensorflow_io.frame as frame_io
from sklearn.model_selection import train_test_split

"""
NOTE: converted from tutorial:
test from https://www.tensorflow.org/beta/tutorials/keras/feature_columns
"""
sample = "heart.csv"

df = frame_io.DataFrame.from_csv(sample)

train, test = df.split(lambda x: train_test_split(x, test_size=0.2))
train, val = train.split(lambda x: train_test_split(x, test_size=0.2))

print(len(train), 'train examples')
print(len(val), 'validation examples')
print(len(test), 'test examples')

def df_to_dataset(dataframe, shuffle=True, batch_size=32):
  # dataframe = dataframe.copy()
  labels = dataframe.pop('target')
  ds = tf.data.Dataset.from_tensor_slices((dict(dataframe), labels))
  if shuffle:
    ds = ds.shuffle(buffer_size=len(dataframe))
  ds = ds.batch(batch_size)
  return ds

batch_size = 5 # A small batch sized is used for demonstration purposes
train_ds = df_to_dataset(train, batch_size=batch_size)
val_ds = df_to_dataset(val, shuffle=False, batch_size=batch_size)
test_ds = df_to_dataset(test, shuffle=False, batch_size=batch_size)

...

model.fit(train_ds, validation_data=val_ds, epochs=5)
```

yongtang:
Added additional ops, and noticed that for this to be really useful (combined with sklearn and other tools), it is more of an implementation about indexing and slicing so that any numpy or pandas compatible tools could utilize tf. More work is needed, but looks promising.

tensorflow_io/frame/python/ops/frame_ops.py
```diff
@@ -0,0 +1,168 @@
...
+
+  @property
+  def iloc(self):
+    return self._iloc
+
+  def __len__(self):
+    return len(self._index)
+
+  def __iter__(self):
+    return iter(self._index)
```
BryanCutler: I think you would want to iterate over `_data` not the index right?

tensorflow_io/core/kernels/dataset_to_tensors_ops.cc
```diff
@@ -0,0 +1,129 @@
...
+        components.reserve(dataset->output_dtypes().size());
+        components_list.emplace_back(components);
+        Status s =
+            raw_iterator->GetNext(&iter_ctx, &components_list.back(), &end_of_sequence);
+        if (!s.ok()) {
+          ctx->SetStatus(s);
+          return;
+        }
+        if (end_of_sequence) {
+          components_list.pop_back();
```
BryanCutler: could you just put this outside the loop?

tensorflow_io/frame/python/ops/frame_ops.py
```diff
@@ -0,0 +1,168 @@
...
+
+import tensorflow as tf
+from tensorflow_io.text.python.ops import text_ops
+from tensorflow_io.core.python.ops import data_ops
+
+class _ILocIndexer(object):
+  """_ILocIndexer"""
+  def __init__(self, df):
+    self._df = df
+    return
```
BryanCutler: nit: don't need the return here

tensorflow_io/frame/python/ops/frame_ops.py
```diff
@@ -0,0 +1,168 @@
...
+  def head(self, n=5):
+    return self.iloc[:n]
+
+  def tail(self, n=5):
+    if n == 0:
+      return self.iloc[0:0]
+    return self.iloc[-n:]
+
+  def split(self, func):
+    indices_x, indices_y = func(self._index)
```
terrytangyuan: What if users want to use one of the [splitter classes](https://scikit-learn.org/stable/modules/classes.html#splitter-classes)? Can we support that?

tensorflow_io/frame/python/ops/frame_ops.py
yongtang: Yes. Updated

tensorflow_io/frame/python/ops/frame_ops.py
yongtang: @terrytangyuan I think it should be possible, likely by checking the input to see if it is a splitter class or a function. The split could call accordingly.

yongtang:
@BryanCutler Yes the idea is to have a pandas API that use tensorflow implementation. At the moment everything is pretty rough with all data read into a tensor (then in a tf.Variable for mutable). But this could be extended into data lazy read, and data could be split into chunks where each chunk fits into gpu.

Also, the dataset_to_tensor implantation was a shortcut to read the whole data into one tensor so that we could reuse the tf.data pipeline. I have changed to use iterator then concat `tf.concat([v for v in dataset], axis=0)`, which does not requires new C++ implementation of `dataset_to_tensor` anymore.

The PR is still a proof-of-concept at the moment. Will spend more time on this PR.

yongtang:
Let me close this one for now. I think we have other ways to achieve complex feature engineering.

PR #360, opened 2019-07-18
Remove python 2.7 and 3.4 mentions since we no longer support them

terrytangyuan:


yongtang:
@terrytangyuan I think we still support 2.7. We don't support 3.4 on macOS (though 3.4 for Linux is still available). We could consider drop 3.4 support for Linux.

terrytangyuan:
Why are we still supporting 2.7? We should drop it. I don't think it's useful to keep 3.4 support fo Linux either.

BryanCutler:
I think it's because TensorFlow is still supporting these, not sure what is planned for tf 2.0 though

yongtang:
@terrytangyuan I think there is still a big community using python 2.7 (as Ubuntu 16.04 and Ubuntu 18.04's default), and this aligns with tensorflow's whl package release.

Officially python 2.7 will EOL next year. We could consider drop it then I think.

It should be safe to drop python 3.4, as python 3.4 is the default python3 package on Ubuntu 14.04 and Ubuntu 14.04 standard support has been dropped this march, and python 3.4 has been EOL'ed this march as well.

Ironically, we still need to use Ubuntu 14.04 to build tensorflow and tensorflow-io for python 2, 3.5, 3.6 to make it binary compatible with tensorflow. (until tensorflow switch to CentOS 6 or 7).

yongtang:
@terrytangyuan @BryanCutler On a side note, I think we could `promote` python 3 by changing dev docker image to use python 3.7 + Ubuntu 16.04? (Note python 3.7 can only be built with Ubuntu 16.04)

In this way, contributors will be feeling more comfortable with python 3 with tensorflow-io.

terrytangyuan:
Yes, let's do that. Should I close this PR then?

yongtang:
@terrytangyuan I think we can drop 3.4, but leave 2.7 alone.

BryanCutler:
>On a side note, I think we could promote python 3 by changing dev docker image to use python 3.7 + Ubuntu 16.04? (Note python 3.7 can only be built with Ubuntu 16.04)

@yongtang the dev image starts up with Python 3.6, but I think we can make it 3.7 if you think that's better.

yongtang:
@BryanCutler If python 3.6 I think that should be fine. That is recent enough I believe.

terrytangyuan:
@yongtang Done. Updated.

PR #362, opened 2019-07-19
Fix ArrowStreamDataset batch slicing and run thread as daemon

BryanCutler:
The `ArrowStreamDataset.from_pandas` was incorrectly slicing the DataFrame into `batch_size` batches, instead of N slices of size `batch_size`. This fix also runs the background thread as daemon.


BryanCutler:
Thanks @yongtang !

PR #363, opened 2019-07-19
Add an entry in README.md for dicom support

yongtang:
Add an entry in README.md for dicom support


yongtang:
Added an entry of dicom to the top level README.md. @marcelolerendegui @Ouwen Maybe you could create a PR to add a README.md in dicom directory and add the link from the top level README.md?

Ouwen:
@yongtang, sounds good we can do that.

PR #365, opened 2019-07-20
Dicom

marcelolerendegui:
Fixed error:
- Images were not being loaded because of: "Warning: no data dictionary loaded"
- Changed dcmtk.BUILD to define ENABLE_BUILTIN_DICTIONARY

Fixed style:
- In decode_dicom_image.cc, class member var names without underscore at the end.

Added dicom tests:
- test_dicom_eager.py
- test_dicom.py
- dicom_samples.sh: script to download and extract the samples to test
- dicom_sample_source.txt: download urls for the samples

Added readme file created by @Ouwen:
- Added dicom/README.md


Ouwen:
@googlebot I consent.

tests/test_dicom_eager.py
```diff
@@ -13,14 +13,32 @@
...
+import pytest
+
 import tensorflow as tf
+import tensorflow_io.dicom as dicom_io  # pylint: disable=wrong-import-position
+
+
 if not (hasattr(tf, "version") and tf.version.VERSION.startswith("2.")):
   tf.compat.v1.enable_eager_execution()
-import tensorflow_io.dicom as dicom_io  # pylint: disable=wrong-import-position
+
```
yongtang: @marcelolerendegui The line of enable_eager_execution() ideally should be placed immediately after `import tensorflow as tf`:
```
import tensorflow as tf
if not (hasattr(tf, "version") and tf.version.VERSION.startswith("2.")):
import tensorflow_io.dicom as dicom_io  # pylint: disable=wrong-import-position
```

The reason is that TensorFlow can only work in either eager mode or non eager mode. However, this mode has to be done immediately after `import tensorflow as tf` as otherwise there will be unintended consequences.

tests/test_dicom.py
```diff
@@ -0,0 +1,181 @@
...
+from __future__ import print_function
+
+import os
+import pytest
+
+import tensorflow as tf
+import tensorflow_io.dicom as dicom_io  # pylint: disable=wrong-import-position
+
+
+tf.compat.v1.disable_eager_execution()
```
yongtang: @marcelolerendegui Also, disable_eager_execution() need to be placed immediately after `import tensorflow as tf` as well for the same reason.

tests/test_dicom.py
```diff
@@ -0,0 +1,179 @@
...
+  """test_decode_dicom_image
+  """
+
+  dcm_path = os.path.join(
+      os.path.dirname(os.path.abspath(__file__)),
+      "test_dicom",
+      fname
+  )
+
+  g1 = tf.Graph()
```
yongtang: @marcelolerendegui For non-eager tests (tests without name of `test_xxx_eager.py`, our tests runs with both TF 1.x and 2.0. Since in 2.0 graph mode has been deprecated, we have to use v1 API of:

```
tf.compat.v1.Graph()
```

tests/test_dicom.py
```diff
@@ -0,0 +1,179 @@
...
+  g1 = tf.compat.v1.Graph()
+
+  with g1.as_default():
+    file_contents = tf.io.read_file(filename=dcm_path)
+    dcm_data = dicom_io.decode_dicom_data(
+        contents=file_contents,
+        tags=tag
+    )
+
+  sess = tf.Session(graph=g1)
```
yongtang: @marcelolerendegui `tf.Session` also needs to be in v1 `tf.compat.v1.Session`

marcelolerendegui:
@yongtang Is there any way to run the bash script that downloads the DICOM sample files? should I run it from test_dicom.py?

yongtang:
@marcelolerendegui There are two places to add:
- one is in .travis.yml

https://github.com/tensorflow/io/blob/master/.travis.yml#L27

- another is in .kokorun/io_cpu.sh

https://github.com/tensorflow/io/blob/master/.kokorun/io_cpu.sh#L54

You can place the script after other setup scripts.

yongtang:
All tests finally pass so I will merge. Thanks for the great contribution 👍 🎉

PR #368, opened 2019-07-23
Add re2_full_match support

yongtang:
This PR adds re2_full_match support
so that it is possible to run regex on string tensor,
and return match (bool) and captured groups.


yongtang:
@terrytangyuan This PR is mostly used in situations where you have a big log file and you want to extract certain fields (thus using regular expression to capture groups). It is to be used in combination of TextDataset.

BryanCutler:
Very nice @yongtang !

PR #371, opened 2019-07-23
Support repeat in ArrowStreamDataset

BryanCutler:
This change adds support for repeating the `ArrowStreamDataset` when using `from_pandas` and `from_record_batches`.  Also changed `ArrowDataset.from_record_batches` column argument to have a default value of None, to match `ArrowStreamDataset`.

### DIFF
--- a/tests/test_arrow_eager.py
+++ b/tests/test_arrow_eager.py
@@ -545,6 +545,34 @@ class ArrowDatasetTest(test.TestCase):
         preserve_index=False)
     self.run_test_case(dataset, truth_data)
 
+  def test_stream_from_pandas_repeat(self):
+    """test_stream_from_pandas_repeat"""
+
+    batch_data = TruthData(
+        self.scalar_data,
+        self.scalar_dtypes,
+        self.scalar_shapes)
+
+    batch = self.make_record_batch(batch_data)
+    df = batch.to_pandas()
+
+    num_repeat = 10
+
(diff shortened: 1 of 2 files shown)


BryanCutler:
cc @yongtang @feihugis

tests/test_arrow_eager.py
```diff
@@ -545,6 +545,27 @@ def test_stream_from_pandas_not_batched(self):
...
+        self.scalar_dtypes,
+        self.scalar_shapes)
+
+    batch = self.make_record_batch(truth_data)
+    df = batch.to_pandas()
+
+    dataset = arrow_io.ArrowStreamDataset.from_pandas(
+        df,
+        batch_size=2,
+        preserve_index=False).repeat(1)
```
feihugis: Could we try a larger repeat number?

tests/test_arrow_eager.py
```diff
@@ -150,6 +150,7 @@ def evaluate_result(value):
         # Check the result then increment the row counter
         evaluate_result(value)
         row += 1
+        print("*** row")
```
feihugis: Is this for debugging?

BryanCutler: oops, yes I will remove

tests/test_arrow_eager.py
BryanCutler: Sure, I can increase it

BryanCutler:
Thanks for the review @feihugis , I updated the test

PR #378, opened 2019-07-25
Authenticate with device code to azure

damienpontifex:
Unsure on whether this is the best place for this, but it eases retrieval of authentication information for azure storage access.

Idea being you could call `authenticate_with_device_code` inside your interactive code to automatically set up auth information for az:// storage use. 

Another investigation is how to use azure CLI auth to retrieve this information if a user is already logged in on their machine

### DIFF
--- a/tensorflow_io/azure/__init__.py
+++ b/tensorflow_io/azure/__init__.py
@@ -70,6 +70,7 @@ def authenticate_with_device_code(account_name):
     subscriptions = json.load(f)
   subscriptions = subscriptions['value']
 
+  storage_account = None
   for subscription in subscriptions:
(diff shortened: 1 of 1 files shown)


yongtang:
Thanks @damienpontifex, can you take a look at lint issue?
```
************* Module tensorflow_io.azure
W: 83,27: map/filter on lambda could be replaced by comprehension (deprecated-lambda)
```

damienpontifex:
@yongtang seems like test failures from Travis. I can't see any connection to this change though

yongtang:
@damienpontifex Can you try to rebase with the master? There might be some changes recently that may has some side effects.

yongtang:
@damienpontifex All tests passed now.

PR #380, opened 2019-07-26
Expose cache for macOS as well

yongtang:


yongtang:
Plumbing change to enable macOS bazel cache (missed on last checkin due to `sudo -H`)

Issue #382, opened 2019-07-26
Discuss Batch Standards in TFIO with Keras


BryanCutler:
Following the discussion on #366 batching can serve different purposes and optimizing for each is not always done the same way.

>Previously, we thought we will reuse the same batch concept for serving two purposes:
Batch is a way to improve performance for record based files such as parquet, feature, etc, as we really don't want to "read one integer or one float32 at a time". This is a caching issue.
>
>We also want to directly feed the dataset to tf.keras so batch concept in tf.keras is reused (number of samples populated in neural network). This is a model parameter.
But those two batch concepts are different. @BryanCutler I am wondering if it makes sense to split those things out. In other words, for 1) we should "read as much record as possible, as long as it fits the memory", and for 2) we should do a rebatch() to adjust the batch_size that has been feeded to tf.keras?

yongtang:
To adds additional information, tf.data actually has a `tf.data.Dataset.concatenate` API which effectively solves the issue of multiple files:
```
a = Dataset.range(1, 4)  # ==> [ 1, 2, 3 ]
b = Dataset.range(4, 8)  # ==> [ 4, 5, 6, 7 ]
a.concatenate(b)  # ==> [ 1, 2, 3, 4, 5, 6, 7 ]
```

I think we only need to focus on single filename case (as multiple files could be processed at higher level with concatenate). After Dataset has been concatenated batch mode could reapply:
```
a.concatenate(b).batch(batch_size, drop_reminder)

# or

a.batch(batch_size, drop_reminder).concatenate(b.batch(batch_size, drop_reminder)
```

I do think `rebatch` could be a useful feature which is not available in tf.data.Dataset. `rebatch` could be emulated with:
```
d.unbatch().batch(batch_size) # <= d.rebatch(batch_size)
```

but I suspect `d.unbatch().batch(batch_size)` is not going to have a good performance.

yongtang:
Added a PR #393 to introduce `rebatch(batch_size) ~= unbatch().batch(batch_size)`.

@terrytangyuan @BryanCutler Here is my thinking on tf.keras batch vs cache batch issue:
1) We want to max out the cache batch (unless not fit into memory)
2) the tf.keras batch_size could always be placed as the last step before feeding too tf.keras (so that caching in the middle are not split into small chunks (of several float32 numbers).

So overall I think we could:
1) All Dataset API will not expose a batch_size, but internally Dataset always batched (up to the cache size).
[...]

Issue #342, opened 2019-07-12
output_shape doesn't make sense, the output remains no structure


CaptainDuke:
As mentioned before, I tried to read hdf5 data using tf-io, but the output_shape doesn't make sense.  ([1,8]) expected but returns ( ).

For comparing, I worte dataset2 form tf.data.Dataset.from_tensor_slices(). Also I did some tf.reshape ops, where dataset read via tf-io failed. 

I thinks it's a kind of serious problem since we can not continue tf.layers.conv2d or other ops without output_shapes.

Hope this issue would be fixed. Thanks.

![image](https://user-images.githubusercontent.com/24515303/61100918-897b7880-a49a-11e9-8bc2-a3c75db282d7.png)
![image](https://user-images.githubusercontent.com/24515303/61100992-c47dac00-a49a-11e9-905a-94619b9216c5.png)

CaptainDuke:
Code above 
[issue_test.py.zip](https://github.com/tensorflow/io/files/3384898/issue_test.py.zip)

CaptainDuke:
temporary solution:
add dataset.batch() and then tf.reshape
`
dataset = hdf5_io.HDF5Dataset(
       [filename], columns, output_types, output_shapes)
dataset = dataset.batch(10)
iterator = dataset.make_one_shot_iterator()
get_next = iterator.get_next()

 input_x = tf.reshape(get_next, output_shape)
sess = tf.Session()
vv = sess.run(tf.shape(input_x))
print(vv)
`

Besides, if we want to use dataset.shuffle and dataset.prefetch then, dataset.make_one_shot_iterator() works.

yongtang:
@CaptainDuke in 2.0 the output shape and dtype are implicitly probed. I believe the issue has been resolved. Will close the issue for now but feel free to reopen if issue still persists.

PR #384, opened 2019-07-27
Rework on ParquetDataset for easy access and better cache size in eager mode

yongtang:
This fix is part of the effort to improve overall Dataset for easy access and better cache size in eager mode. See #382 and #366 for related discussions.

In order to be able to read file either in filename or in mmeory, this PR adds an SizedRandomAccessFile which allows to provide an optional memory buffer as file content. This could be useful in process compression or archives where we could just read the uncompressed file content into memory.

The preivous limitation in Dataset was that Dataset was a iterable so sequence length is unknown until graph runtime. In this PR, we provide an helper function to read the columns of parquet file and lenth is know.

This also could open other avenues such as map parquet file with __getitem__ and __len__.
Further, parquet file could be read into a Tensor and processed easily (such as pandas like API).

The list_parquet_columns could be similarly applied to HDF5 which is more important: HDF5 could have dataset with different sizes.

Summary:
1) Two basic C++ kernel ops are implemnted: list_parquet_columns and read_parquet
2) One ParquetDataset that is python implementation only (no C++ anymore)
3) ParquetDataset support eager and graph mode, in graph mode, dtype and shape
   are provided by user explicitly. In eager mode, only column name is needed.
4) read_parquet works in eager and graph mode, can read records either in full, or in slices
5) list_parquet_columns works in eager mode only (limitation).

For cache batch vs. batch in tf.keras
1) Added a hidden `capacity` to adjust the cache batch size
2) batch to be passed in tf.keras is unrelated to `capacity`, but we could use `rebatch`
   to change at the end of the pipeline.
3) `capacity` could be padded to allow `rebatch` to only cut a slice over one chunk.
   If not padded to `batch_size` in tf.keras, then `rebatch` likely will copy over boundary.


CaptainDuke:
Many thanks to Yongtang.

Yes, actually contents in HDF5 files do not need to decode. Also I'm working on HDF5 files with diffierent size. For example.
```
# h5ls test_data_level_6/10.hdf5 
atk_diff            Dataset {5120, 1}
emy_vec_5           Dataset {5120, 429}
frame                    Dataset {5120, 1}
global_info              Dataset {5120, 68}
hot_label                Dataset {5120, 1}
hot_weight               Dataset {5120, 1}
img_data                 Dataset {5120, 5, 31, 31}
...
```
I believe such enhancement would be helpful. 
BTW, is [#342 issue](https://github.com/tensorflow/io/issues/342) bug related to this problem? 

> The preivous limitation in Dataset was that Dataset was a iterable so sequence length is unknown until graph runtime. In this PR, we provide an helper function to read the specs of parquet file and lenth is know.

yongtang:
@CaptainDuke the issue #342 you are referring to, might not be directly related to this problem. However, the recent changes in upstream tf.data: https://github.com/tensorflow/tensorflow/commit/c5c183971b3cb42b123474e1e8a50a99c930b8db

might make things complicated as we likely will need to update API pretty soon. With the ongoing rework of cache size and tf.io pipeline to interact with tf.data, it might make sense to fix that together with the PR here.

PR #392, opened 2019-07-28
Rework on HDF5: add list_hdf5_datasets and read_hdf5 ops

yongtang:
This PR is part of the effort in enhancing performance and ease of use for tf.data pipeline, as was discussed in #382 and #366.

Previously, HDF5Dataset is relatively manual and user has to find out the dataset (columns) in hdf5 file.

In this PR, the idea is to allow user to use list_hdf5_datasets to probe the shape, dtype, and name of the datasets within HDF5. A subsequent call to read_hdf5 will bring content to a shaped Tensor so that it could be used later in TensorFlow.

The read_hdf5 has the option to specify a slice (or a subblock) of the dataset. This should open up possibility in the future to allow binding a class with a hdf5 file by implement `__len__` and `__getitem__`.

With list_hdf5_datasets and read_hdf5 ops, it is also possible to ease the HDF5Dataset in eager mode. In eager mode, HDF5Dataset could juse call list_hdf5_datasets to find out all the necessary information, then calling read_hdf5 in pieces to maintain the `batch_size` to be fed in tf.keras.

The limitation would be in graph mode as in graph mode user still has to specify almost everything dtype, shape, name for HDF5Dataset to work.

This PR has not changed HDF5Dataset implementation to use list_hdf5_datasets and read_hdf5 ops. But this could be easily done and see #384 for similar changes.


yongtang:
@CaptainDuke I changed the `start:count` to `start:stop`, to match the suggestion in another PR:
https://github.com/tensorflow/io/pull/406#issuecomment-517687583

Plan to merge this PR shortly.

terrytangyuan:
@yongtang Thanks!

CaptainDuke:
@yongtang Thanks!

PR #397, opened 2019-07-30
Add read_text to read lines from splittable text file

yongtang:
This PR is part of the effort to rework on Dataset with large files reading into Tensors first to speed up performance. See #382 and #366 for related discussions.

Summary:
1) read_text is able to read a text file with in the range of [offset, offset+length]
2) that gives us the Splittable text file where we could read file in chunks (similar to hadoop)
3) the plan is to read a text file in big chunks and then wire up with tf.data.Dataset
4) read_text is a primitive C++ op so it could be used in tf.data, and it could be used in other places.


yongtang:
Will merge this PR soon as well. It exposes a primitive kernel op `read_text` which allows reading text in slices. This is more useful to tf.data API in most of the cases.

PR #402, opened 2019-08-01
Rework on PrometheusDataset

yongtang:
This is part of the effort to reduce the dedicated C++
implementation of Dataset and replace with primitive ops
that could be used both with tf.data, and with Tensor.

See #382 and #366 for related discussions.

There are some room for enhancement, for example,
a timestamp could be passed to read_prometheus
and each call will only read a small slice of the data
to tensor. Will have follow up PRs later to implement that.


yongtang:
Plan to merge this PR shortly.

PR #406, opened 2019-08-02
Add read_wav and list_wav_info support

yongtang:
This PR adds read_wav and list_wav_info support so that it is possible to read wav file into a tensor.
Since WAV file is splittable, read_wav could take a start and count parameter so that only a slice of the wav file is read. 

Read a chunk of the wav file is related to https://github.com/tensorflow/io/issues/49#issuecomment-498703460

This is also part of the effort to rework on Dataset to move to primitive ops. See See #382 and #366 for related discussions.


yongtang:
/cc @faroit for reading a chunk of WAV file. You can use `read_wav` and pass `start=start, count=count` to only read a chunk of the samples.

Note start stand for `the starting index of the sample`. 

The total number of samples in WAV could be obtained through:
`spec, rate = list_wav_info(filename)` where spec is a `TensorSpec` which gives you the shape, and dtype of the tensor. Shape is [n, c] where n is the total number of samples and c is the channels.

The dtype could be `int8` or `int16` for 8-bit and 16-bit.

I tried to see if I could also add `24bit`, though I need a sample 24bit WAV file to see the actual memory layout (to figure out how to fit 24bit into an `int32`)

faroit:
@yongtang This is great!

`count` sounds a bit wobbly, I would propose to use either 

* `frames` or `num_frames` (samples*channel)  
* `length` (duration in samples)

or to set the end sample in samples

* `stop`
* `end`

faroit:
@yongtang before I try it: is the wavdataset form this PR compatible to the tf 2.x beta?

faroit:
Looks great, thanks

PR #407, opened 2019-08-03
Add NumpyDataset support

yongtang:
This PR adds support for reading npy and npz files, as well as reading numpy array from python process, related to #68.

### DIFF
--- a/tensorflow_io/core/BUILD
+++ b/tensorflow_io/core/BUILD
@@ -415,6 +415,20 @@ cc_library(
     alwayslink = 1,
 )
 
+cc_library(
+    name = "numpy_ops",
+    srcs = [
+        "kernels/numpy_kernels.cc",
+        "ops/numpy_ops.cc",
+    ],
+    copts = tf_io_copts(),
+    linkstatic = True,
+    deps = [
+        "//tensorflow_io/core:dataset_ops",
+    ],
+    alwayslink = 1,
+)
+
 cc_binary(
     name = "python/ops/libtensorflow_io.so",
     copts = tf_io_copts(),
@@ -435,6 +449,7 @@ cc_binary(
         "//tensorflow_io/core:json_ops",
         "//tensorflow_io/core:kinesis_ops",
         "//tensorflow_io/core:lmdb_ops",
+        "//tensorflow_io/core:numpy_ops",
         "//tensorflow_io/core:parquet_ops",
         "//tensorflow_io/core:pcap_ops",
         "//tensorflow_io/core:pubsub_ops",
--- a/tensorflow_io/core/kernels/numpy_kernels.cc
+++ b/tensorflow_io/core/kernels/numpy_kernels.cc
@@ -1,0 +1,613 @@
+/* Copyright 2019 The TensorFlow Authors. All Rights Reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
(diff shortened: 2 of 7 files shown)


yongtang:
Actually reading numpy array from python process seems to be straightforward, as numpy's memory layout is consistent. Will update shortly.

yongtang:
@terrytangyuan @BryanCutler I just realized that in case tensorflow 's kernel is local, then kernel and python will be in the same process. That means memory address space is accessible.

That will save a lot of effort for us I think. Maybe this could be applied to Arrow Batch/Stream as well?

In case tensorflow's session is not local, if they are in the same machine then we still could access the memory from another process.

BryanCutler:
>I just realized that in case tensorflow 's kernel is local, then kernel and python will be in the same process. That means memory address space is accessible.
>
>That will save a lot of effort for us I think. Maybe this could be applied to Arrow Batch/Stream as well?

I have actually been working on this, will post a WIP PR to discuss more

tensorflow_io/numpy/python/ops/numpy_ops.py
```diff
@@ -0,0 +1,169 @@
...
+      dtype = array.dtype
+      shape = tf.TensorShape(
+          [dim if i != 0 else None for i, dim in enumerate(
+              array.shape.as_list())])
+      start = 0
+      stop = array.shape[0]
+    else:
+      filename = ""
+      arrayname = ""
+      address, _ = array.__array_interface__['data']
```
BryanCutler: One concern I had with sharing memory addresses between Python and C++ is that I think it's possible for the numpy array to go out of scope and get cleaned up, but the Dataset could still exists and hold an address that is no longer valid. If you store a reference to the data in the dataset, I think that should be enough. WDYT @yongtang ?

yongtang:
@BryanCutler Thanks! Updated the PR with a reference to data added.

tensorflow_io/numpy/python/ops/numpy_ops.py
```diff
@@ -0,0 +1,170 @@
...
+      address = 0
+      arrayshape = array.shape.as_list()
+      dtype = array.dtype
+      shape = tf.TensorShape(
+          [dim if i != 0 else None for i, dim in enumerate(
+              array.shape.as_list())])
+      start = 0
+      stop = array.shape[0]
+    else:
+      self._array_data = array.data
```
BryanCutler: I believe `array.data` is a memoryview, does that increment the refcount of the object?

yongtang:
Thanks @BryanCutler. I updated with `np.array(array, copy=False)`. Think this should solve the issue.

tensorflow_io/numpy/python/ops/numpy_ops.py
```diff
@@ -0,0 +1,170 @@
...
+      address = 0
+      arrayshape = array.shape.as_list()
+      dtype = array.dtype
+      shape = tf.TensorShape(
+          [dim if i != 0 else None for i, dim in enumerate(
+              array.shape.as_list())])
+      start = 0
+      stop = array.shape[0]
+    else:
+      self._array_holder = np.array(array, copy=False)
```
BryanCutler: This looks fine, but could you just do `self._array_holder = array`?

yongtang:
Change to Working-in-Progress, as I think there might be some further improvement if we could come up with a better alignment /memory pool that fits into TF.

terrytangyuan:
@yongtang Any update on this? Perhaps we could get this in first and then improve later in separate PRs?

lxiongh:
Is there any plan to support non-eager mode?

yongtang:
@lxiongh Non-eager mode should not be very difficult to support, as the data type is visible from python side.

yongtang:
@lxiongh @terrytangyuan It may take several days to update this PR, as quite a few things have been updated in tensorflow-io. I will see if I can find some time to get this one in this year. Otherwise it will likely be updated in January.

yongtang:
The PR has been updated and all should work. The graph mode for NumpyDataset is not supported yet. Once this PR is merged I will have a follow up PR to add graph mode support. (graph mode will require user to provide dtype before hand).

/cc @terrytangyuan

terrytangyuan:
Great! Thanks for the efforts!

PR #408, opened 2019-08-05
JSONDataset: support for different file schemes.

jiachengxu:
This PR provides support for JSONDataset to read JSON files in different file schemes.
Ref: #262

### DIFF
--- a/tensorflow_io/json/kernels/json_input.cc
+++ b/tensorflow_io/json/kernels/json_input.cc
@@ -25,7 +25,7 @@ namespace data {
 
 class JSONInputStream {
 public:
-  explicit JSONInputStream(io::InputStreamInterface* s, const string& filename) {
+  explicit JSONInputStream(const string& filename) {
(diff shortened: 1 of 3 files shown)


tensorflow_io/json/kernels/json_input.cc
```diff
@@ -17,22 +17,32 @@ limitations under the License.
...
 #include "include/json/json.h"
 
 namespace tensorflow {
 namespace data {
 
 class JSONInputStream {
 public:
-  explicit JSONInputStream(io::InputStreamInterface* s, const string& filename)
-    : ifs_(filename){}
+  explicit JSONInputStream(io::InputStreamInterface* s, const string& filename) {
```
yongtang: Since `io::InputStreamInterface* s, ` is not going to be used any more, can you remove this parameter?

jiachengxu: Yes, PR updated

PR #413, opened 2019-08-07
Enable zero-copy transfer  to ArrowDataset

BryanCutler:
This change enables zero-copy transfer of pyarrow buffer to ArrowDataset kernel when in eager mode and if the kernel process is local. Instead of serializing the entire buffer as input into the ArrowDataset, this uses the buffer address and size so that in C++, Arrow can create a zero-copy buffer to read record batches from. A reference to the pyarrow buffer is stored in the Dataset to prevent cleanup during the lifetime of the Dataset.

### DIFF
--- a/tensorflow_io/arrow/kernels/arrow_dataset_ops.cc
+++ b/tensorflow_io/arrow/kernels/arrow_dataset_ops.cc
@@ -583,11 +583,155 @@ class ArrowOpKernelBase : public DatasetOpKernel {
   std::vector<PartialTensorShape> output_shapes_;
 };
 
-// Op to create an ArrowDataset that consumes Arrow record batches from
-// memory in a Python process, or a Pandas DataFrame.
-class ArrowDatasetOp : public ArrowOpKernelBase {
+// Op to create an ArrowZeroCopyDataset that consumes Arrow record batches
+// from a memory buffer address owned in Python.
+class ArrowZeroCopyDatasetOp : public ArrowOpKernelBase {
  public:
-  explicit ArrowDatasetOp(OpKernelConstruction* ctx) : ArrowOpKernelBase(ctx) {}
+  explicit ArrowZeroCopyDatasetOp(OpKernelConstruction* ctx)
+      : ArrowOpKernelBase(ctx) {}
+
+  virtual void MakeArrowDataset(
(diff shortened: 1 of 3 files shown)


BryanCutler:
cc @yongtang @terrytangyuan

BryanCutler:
There are a couple improvement that could be made here:

1. The `ArrowZeroCopyDataset` is better suited to be a primitive op that can be wrapped in a Python tf.data.Dataset, similar to `NumpyDataset` in #407 

2. Arrow Columns can be made up of multiple chunks, each in a separate buffer. Ideally, all buffers used in all columns should be input to the Dataset, which could then reconstruct the columns into a Table. Currently, this takes a simple approach of creating a single buffer and writing all columns into it in Python. Then the single buffer is reconstructed in C++.

yongtang:
@BryanCutler I will take a closer look during the weekend. Several things I am thinking of:
1) We actually could use RT_RESOURCE and implement custom primitive ops as well, since RT_RESOURCE will persist in memory.
2) For same process we can certainly pass address directly. But even if they are different processes, it is still possible to rely on OS system calls to tap into another process's memory address and read memory. (in Linux that could be `process_vm_readv` and in Windows that used to be `ReadProcessMemory` if I remember correctly : haven't touched WIN32 for years).
3) The only case we need blocking/synchronous serialization is the case where the Tensor is transferred to another server/node. This is very rare and could be fallback.

The above are just some general discussions, not necessarily to be related to this PR.

BryanCutler:
Thanks @yongtang , those sounds like great ideas. Let's definitely follow up and look into those. The Mac test failures look unrelated to this change.

BryanCutler:
I going to go ahead and merge this to get another PR up. I can address comments on a followup, as there will be some more improvements to this as @yongtang pointed out.

jsimsa:
Drive by comment: it would be great to add tests for this functionality that, besides testing the correctness, illustrate how to use it.

BryanCutler:
> Drive by comment: it would be great to add tests for this functionality that, besides testing the correctness, illustrate how to use it.

Thanks for the comment @jsimsa ! I'm working on some improvements to this and then will definitely follow up with more tests and docs for usage.

PR #416, opened 2019-08-09
[WIP] Expose top level tfio.Dataset as a superset of tf.data.Dataset

yongtang:
This PR expose top level tfio.Dataset as a superset of tf.data.Dataset.

In the past, we have been trying to place different Dataset implementations
into individual modules. For example, we have tensorflow_io.arrow.ArrowFeatherDataset
under `tensorflow_io.arrow`. As the project grow we already have too many modules and
it is getting harder to find the features we want.

This PR tries to expose a top level tfio.Dataset as a superset of tf.data.Dataset,
so that users could use tfio.Dataset the same way as tf.data.Dataset and
with additional functionalities.

To achive that first we have to expose all functions in tf.data.Dataset to tfio.Dataset.
Since tfio.Dataset is a subclass we also have to fix the return types so that
`tfio.Dataset.range(5)` will returns a subclass of `tfio.Dataset` (even though
we redirect tfio.Dataset.range(5) to tf.data.Dataset.range(5) which by nature
will return a subclass of `tf.data.Dataset`.)

This PR creates an wrapper of `_DatasetAdapter` and `dataset_decorator` so that
return type could be adjusted.

Also, all functions/methods of `tf.data.Dataset` are routed with `tfio.Dataset`
automatically. (We could decide to manually define those functions/methods instead,
if we feel it is much better).

In addition to create a class of tfio.Dataset, this PR also adds the `rebatch()`
as part of `tfio.Dataset`, so that user could use it in an natural way.
```
tfio.Dataset.range(5).batch(3).rebatch(5)
```

Further, I think we could also start exposing individual datasets as part of
the static method in `tfio.Dataset`. For example, we could imagine:
```
tfio.Dataset.from_arrow(filename)
```
which would be very similiar to:
```
tf.data.Dataset.from_tensors(...)
```

Note those static methods of `from_arrow` has not been added yet, we could discuss
what would be the best API to expose those.

Finally the way of using `inspect.getmembers` with `_DatasetAdapter` and `dataset_decorator`
in this PR may not be the best way. A more pythonic way might be to use metaclass.
However, as tf.data.Dataset is also a metaclass so things are complicated. This PR is
what I can think of so far but ideas or contributions to make the mechanism more elegant
are welcomed.

### DIFF
--- a/tensorflow_io/__init__.py
+++ b/tensorflow_io/__init__.py
@@ -57,3 +57,7 @@ def _load_library(filename, lib="op"):
   raise NotImplementedError(
       "unable to open file: " +
       "{}, from paths: {}\ncaused by: {}".format(filename, filenames, errs))
+
+from tensorflow_io.core.python.ops import dataset_ops # pylint: disable=wrong-import-position
+
+Dataset = dataset_ops.create_dataste_class()
--- a/tensorflow_io/core/python/ops/__init__.py
+++ b/tensorflow_io/core/python/ops/__init__.py
@@ -17,5 +17,5 @@ from __future__ import absolute_import
 from __future__ import division
 from __future__ import print_function
 
-from tensorflow_io import _load_library
+from tensorflow_io import _load_library # pylint: disable=cyclic-import
 core_ops = _load_library('libtensorflow_io.so')
--- a/tensorflow_io/core/python/ops/data_ops.py
+++ b/tensorflow_io/core/python/ops/data_ops.py
@@ -18,40 +18,6 @@ from __future__ import division
 from __future__ import print_function
 
 import tensorflow as tf
-from tensorflow_io.core.python.ops import core_ops
-
-class _AdjustBatchDataset(tf.compat.v2.data.Dataset):
-  """AdjustBatchDataset"""
-
-  def __init__(self, input_dataset, batch_size, batch_mode=""):
-    """Create a AdjustBatchDataset."""
-    self._input_dataset = input_dataset
-    self._batch_size = batch_size
-    self._batch_mode = batch_mode
-
-    self._structure = input_dataset._element_structure._unbatch()._batch(None) # pylint: disable=protected-access
-
-    variant_tensor = core_ops.adjust_batch_dataset(
-        self._input_dataset._variant_tensor,  # pylint: disable=protected-access
-        batch_size=self._batch_size,
-        batch_mode=self._batch_mode,
-        output_types=self._structure._flat_types, # pylint: disable=protected-access
-        output_shapes=self._structure._flat_shapes) # pylint: disable=protected-access
-
-    super(_AdjustBatchDataset, self).__init__(variant_tensor)
-
-  def _inputs(self):
-    return [self._input_dataset]
-
-  @property
-  def _element_structure(self):
-    return self._structure
-
-def rebatch(batch_size, batch_mode=""):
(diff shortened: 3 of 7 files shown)


yongtang:
/cc @BryanCutler @terrytangyuan  to take a look. I am pushing API to top level `tfio.Dataset`, so any suggestions and opinions are welcomed.

Also, if we move forward with top level API, we will need to be really careful as it will be difficult to pull back once it is out (to maintain backward-compatibility for users).

tensorflow_io/__init__.py
```diff
@@ -57,3 +57,7 @@ def _load_library(filename, lib="op"):
   raise NotImplementedError(
       "unable to open file: " +
       "{}, from paths: {}\ncaused by: {}".format(filename, filenames, errs))
+
+from tensorflow_io.core.python.ops import dataset_ops # pylint: disable=wrong-import-position
+
+Dataset = dataset_ops.create_dataste_class()
```
terrytangyuan: typo: `create_dataste_class` -> create_dataset_class

terrytangyuan:
I will need to take a closer look at this soon. In the meantime, could you briefly describe what `rebatch()` does to an end user?

yongtang:
@terrytangyuan here is an example of `rebatch`, which essentially changes the batch size (dim[0]) of each tensor when iterating the dataset:

```python
# batch of size 0 (scalar),
# since total is 4, so we have [1, 1, 1, 1] elements
dataset = tfio.Dataset.range(1, 4) # <=[1, 2, 3, 4]

# batch of size 3,
# since total is 4, so we have [3, 1] elements
dataset = dataset.batch(3)  # <= [[1, 2, 3], [4]] 

# batch of size 2
# since total is 4, so we have [2, 2] elements
dataset = dataset.rebatch(2) # <= [[1, 2], [3, 4]] 
```

yongtang:
@terrytangyuan also, 
```
rebatch(n) = unbatch().batch(n)
```

The reason to have a dedicated `rebatch` is that `unbatch().batch(n)` will need to unroll elements into each single element then piece together again. That will have a performance hit.

yongtang:
With the discussion in https://github.com/tensorflow/tensorflow/issues/31548#issuecomment-520507145 , I think exposing instance method is not needed anymore (.e.g, `rebatch`).

Though it still makes plenty of sense to have an alias of tf.data.Dataset => trio.Dataset which gives us ability to have factory method like from_arrow, from_parquet.

Will update the PR soon.

BryanCutler:
@yongtang it sounds like a great idea to have static methods in `tfio.Dataset` so users have easy access to all functionality in tfio. I'll try to think to think about how to handle the different constructors for Arrow to keep things simple.

yongtang:
@BryanCutler I think `classmethod` could be a good choice for us.

One additional thing is that, with the upcoming TF 2.0 and eager mode, most of the information could already be probed automatically so there is no need for many user inputs that was necessarily in graph mode. I would like to reduce user knobs as much as possible.

BryanCutler:
>One additional thing is that, with the upcoming TF 2.0 and eager mode, most of the information could already be probed automatically so there is no need for many user inputs that was necessarily in graph mode. I would like to reduce user knobs as much as possible.

Yup, I totally agree. The APIs can be simplified a lot for eager mode.

yongtang:
Close the PR as this PR is now covered by #420.

PR #417, opened 2019-08-10
[WIP] Use resource type for LMDB (and iterable data format types)

yongtang:
This is part of the effort to bring tf.io into more primitive TF kernel ops than handcrafted tf.data subgraphs.

In the meeting we discussed several data format types:
- Indexable (support `__len__` and `__getitem__` natively). Example: Fixed length records, parquet).
- Splittable (a file could be splited into multiple definite non-overlapping chunks). Example: Text files (split through LF), Avro (sync position)
- Iterable (only support `__next__` and not knowing length). Example: PCAP file (variable-length record), Kafka streams.
- Block (the whole file has to be read into memory entirely). Example, WebP file, Image files, etc.

With the recently PRs that was meerged, we have converted the following formats:
- Indexable: We allow read with record index of [start, stop) - See `read_parquet`
- Splittable: We allow read with file position of [offset, offset + length) - See `read_avro`
- Block: We read the whole file anyway, as no point to split. - See `read_json`, `decode_webp`.

The remaining one is the Iterable format which still very much requires a handcrafted Dataset C++ subgraph.

The challange is that, when an Iterable format is opened, the state (kafka stream handle) has to stay within the memory and we can not serialize/deserialize to pass around (we could in theory, but that will be a huge cost to always open a Kafka stream to read every record).

It is actually possible to persist a state in memory (without serialization), that is to construct a `DT_RESOURCE` type. A `DT_RESOURCE` type will persist in memory until the reference count reduced to zero.

For that reason, it is possible to construct Iterable format type and build a Dataset on top of primitive TF kernel ops:
- Create a `DT_RESOURCE` type and associate stream/file handle to it in memory
- Itearte through with `__next__`(DT_RESOURCE) until the end (zero record returned)

In tf.data.Dataset's high level API there is a `take_while` which could be used to stop (size(__next__()) == 0) the sequence of dataset.

This PR takes the above approach, and converted LMDB into two kernel ops: `InitLMDB` (return`DT_RESOURCE`), and `NextLMDB` (return key, value pair for a chunk). For the background, LMDB will always return an ordered sequence of key, value pairs. And there is no way to `seek` or `skip` to a position, or knowing the length before hand.

As could be seen in the C++ code, the code base is reduced substantially and quite easy to follow.

To make any future Iterable formats implementation easier, the following C++ interface has been defined:
```c++
class IOIterableInterface : public ResourceBase {
 public:
  virtual Status Init(const string& input, const void* memory_data, const int64 memory_size, const string& metadata) = 0;
  virtual Status Next(const int64 record_to_read, std::vector<Tensor>& tensors, int64* record_read) = 0;

  virtual Status GetShape(std::vector<TensorShape>& shapes) = 0;
};
```

As is easily seen, as long as `Init` and `Next` virtual functions are implemented, the kernel ops could be constructred automatically through:

```c++
REGISTER_KERNEL_BUILDER(Name("InitLMDB").Device(DEVICE_CPU),
                        IOIterableInitOp<LMDBIterable>);
REGISTER_KERNEL_BUILDER(Name("NextLMDB").Device(DEVICE_CPU),
                        IOIterableNextOp<LMDBIterable>);
```

Hopefully this concludes our transition to primitive ops, and attract more contributors to our community.

Also, on why primitive ops are preferred:
As our project grow we will need to maintain backward-compatibility and primitive ops will be a lot easy to work on and interact with upstream TF repo. Think this is very important for the community.

Update:

Also added `IOIndexableInterface` to help construct:
```c++
class IOIndexableInterface : public IOIterableInterface {
 public:
  virtual Status GetItem(const int64 start, const int64 stop, std::vector<Tensor>& tensors) = 0;
  virtual Status Len(int64 *len) = 0;
}
```

Here is an example of usage:
```c++
REGISTER_KERNEL_BUILDER(Name("ReadMNISTLabel").Device(DEVICE_CPU),
                        IOIndexableReadOp<MNISTLabelIndexable>);
```


yongtang:
I take another look, ideally, we could easily bind __getitem__ and __next__ and expose a high level class.

I have cleared most of the issues so far so it is almost ready. The only glitch I encounter is that LMDB is a tuple (or key-value) pair which could be easily handled with tensorflow's CompositeTensor.

Unfortunately CompositeTensor API is not public and has been changed a lot recently, meaning we have to wait for TF 2.0 RC (release candidate).

In the meantime, I can apply the binding of `__getitem__` and `__next__` to an individual data type.

Will put this PR on hold. I will try to see if I could apply `__getitem__` and `__next__` WAV file which does not have the requirement of CompositeTensor.

@terrytangyuan @BryanCutler  Sorry guys, I move back and forth a lot. Though eventually the goal is to have a library that will be much more useful than its current state for the community.

yongtang:
This PR is now superseded by RP #450.

PR #418, opened 2019-08-13
Use TFIO ArrowRandomAccessFile to read Arrow Feather files

BryanCutler:
This PR makes use of the `ArrowRandomAccessFile` from TFIO that wraps a TF `RandomAccessFile` to read Arrow Feather files in the `ArrowFeatherDataset`. By using this interface, it makes it possible to load cloud storage files, as discussed in https://github.com/tensorflow/io/pull/404#issuecomment-518013165.


BryanCutler:
@yongtang please have a look when you can. This is an initial step to use the `ArrowRandomAccessFile` interface in the current ArrowFeatherDataset. This can be refactored later into a primitive op, and then support loading from memory also.

PR #422, opened 2019-08-14
Upgrade Apache Arrow to v0.14.1

BryanCutler:
This Upgrades Apache Arrow to version 0.14.1, also with Parquet-Cpp. The Thrift dependency is upgraded to 0.12.0 to match, and new dependency on Double-Conversion is added.


third_party/arrow.BUILD
```diff
@@ -25,19 +26,29 @@ flatbuffer_cc_library(
     out_prefix = "cpp/src/arrow/ipc/",
 )
 
+genrule(
+  name = "arrow_util_config",
+  srcs = ["cpp/src/arrow/util/config.h.cmake"],
+  outs = ["cpp/src/arrow/util/config.h"],
+  cmd =  "sed 's/@ARROW_VERSION_MAJOR@/0/g ; s/@ARROW_VERSION_MINOR@/14/g ; s/@ARROW_VERSION_PATCH@/1/g ; s/#cmakedefine/#undef/g' $< > $@",
```
BryanCutler: This is pretty hacky, but it's the only way I could figure out how to do it. Any other ideas @yongtang ?

yongtang: @BryanCutler The Bazel allows using Starlark for complex logic. Starlark essentially is a dialect of python. However, I am afraid using Starlark to achieve `cmake` functionality will be more complicated than `sed`.

yongtang:
@BryanCutler You can use the following command to fix the bazel format issue (assuming docker installed):
```sh
docker run -i -t --rm -v $PWD:/v -w /v --net=host golang:1.12 bash -x -e -c 'go get github.com/bazelbuild/buildtools/buildifier && buildifier $(find . -type f \( -name WORKSPACE -or -name BUILD -or -name *.BUILD \))'

```

third_party/arrow.BUILD
BryanCutler: I see, well since this is a simple replace. It's probably not worth doing anything complicated. Is there an easy way to parametrize the version here?

BryanCutler:
>You can use the following command to fix the bazel format issue (assuming docker installed):

Thanks for the reminder, I forgot that command. Do you think this is worth putting in the readme?

BryanCutler:
Strange, I don't see this error when building locally with same bazel version:
```
external/arrow/cpp/src/arrow/util/string_view.h:23:66: fatal error: arrow/vendored/string_view.hpp: No such file or directory

 #include "arrow/vendored/string_view.hpp"  // IWYU pragma: export
```
I guess that file should be declared in arrow.BUILD

yongtang:
> Thanks for the reminder, I forgot that command. Do you think this is worth putting in the readme?

@BryanCutler Adding the command in README.md would be great!

yongtang:
@BryanCutler I think the following patch should fix the build issue:
```diff
diff --git a/third_party/arrow.BUILD b/third_party/arrow.BUILD
index 788f613..4e2a19e 100644
--- a/third_party/arrow.BUILD
+++ b/third_party/arrow.BUILD
@@ -72,6 +72,7 @@ cc_library(
         ],
     ) + [
         "cpp/src/parquet/api/io.h",
+        "cpp/src/parquet/windows_compatibility.h",
         "cpp/src/parquet/api/reader.h",
         "cpp/src/parquet/api/schema.h",
         "cpp/src/parquet/deprecated_io.cc",
diff --git a/third_party/boost.BUILD b/third_party/boost.BUILD
index a0649cb..6290c6e 100644
--- a/third_party/boost.BUILD
+++ b/third_party/boost.BUILD
@@ -11,6 +11,7 @@ cc_library(
     name = "boost",
     srcs = glob([
         "boost/**/*.hpp",
+        "boost/**/*.ipp",
     ]) + glob([
         "boost/predef/**/*.h",
     ]) + [

```

Also, the pyarrow version in `.travis/wheel.test.sh` and `.travis/wheel.r.test.sh` may have to be updated as well.

yongtang:
@BryanCutler Also, looks like both Kafka and Arrow have a copy of xxhash. After you rebase against the master I think the following patch will be needed as well (in addition to the previous patch):
```diff
diff --git a/third_party/arrow.BUILD b/third_party/arrow.BUILD
index 4e2a19e..1e2f3e6 100644
--- a/third_party/arrow.BUILD
+++ b/third_party/arrow.BUILD
@@ -115,6 +115,7 @@ cc_library(
     ],
     includes = [
         "cpp/src",
+        "cpp/src/arrow/vendored/xxhash",
     ],
     deps = [
         ":arrow_format",
diff --git a/third_party/kafka.BUILD b/third_party/kafka.BUILD
index 2106a8d..823b599 100644
--- a/third_party/kafka.BUILD
+++ b/third_party/kafka.BUILD
@@ -137,8 +137,6 @@ cc_library(
         "src/tinycthread.h",
         "src/tinycthread_extra.c",
         "src/tinycthread_extra.h",
-        "src/xxhash.c",
-        "src/xxhash.h",
     ],
     hdrs = [
         "config.h",
@@ -155,6 +153,7 @@ cc_library(
     linkopts = ["-lpthread"],
     visibility = ["//visibility:public"],
     deps = [
+        "@arrow",
         "@boringssl//:ssl",
         "@com_github_madler_zlib//:z",
     ],
```

BryanCutler:
Thanks @yongtang , I'll rebase and try the patches

BryanCutler:
Thanks @yongtang , looks like most of the builds have passed! I'm retrying mac that failed due to timeout.

BryanCutler:
Thanks for the help!

PR #426, opened 2019-08-18
Manylinux2010 and Travis with latest update

yongtang:
Manylinux2010 and Travis with latest update


yongtang:
@BryanCutler @terrytangyuan @dmitrievanthony We are all set for manylinux2010. I have to patch the auditwheel itself, to get around the libtensorflow_frameworks.so.1 issue tough

The test is based on tf-nightly which actually incur quite some changes in tf.data. Since 1.15 that will be RC'ed tomorrow ,will likely use tf-nightly as the base, we will need to fix the test breaks caused by the API change in tf.data.

So far the tests all passed except:

- Ignite tests has been temporarily disabled  (9c8dce4)
- Arrow tests has been temporarily disabled (b704dee)
- Arrow has to comment out the optimization_only usage (4e6718c)

@BryanCutler @dmitrievanthony could you take a look at those failed tests due to the API changes in tf.data?

yongtang:
The original plan was that we wait for 1.15RC in 20th out to merge this PR. Though 1.15RC is not there yet. I will merge this PR shortly anyway, as master build is broken at the moment. We need to clean up Travis CI cache or tensorflow-io-nightly will not be generated anyway.

yongtang:
I will also split out a couple of follow up PR to clean and retriever the Travis CI cache.

PR #427, opened 2019-08-19
[WIP] Switch macOS to xcode8 to see if whl works with macOS 10.11

yongtang:
Switch macOS to xcode8 to see if whl works with macOS 10.11

Experimetnal only


yongtang:
Closing as it looks like pyenv is not available on Travis CI with osx 8.

PR #428, opened 2019-08-20
Add a user guide for JSON file support

jiachengxu:
- Add a user guide for JSON file support.
- Add an entry in the README.md for JSON file support.


jiachengxu:
cc @yongtang @terrytangyuan @BryanCutler

yongtang:
Overall looks good, I think probably could add a couple of more:
1) JSON Dataset is a subclass of `tf.data`, so it fully conforms to TensorFlow's data pipeline (without any intermediate conversion), especially for TF 2.0.
2) JSON Dataset could be used directly with tf.keras which is the recommended high level API for TF 2.0.

jiachengxu:
@yongtang , all good points
> JSON Dataset is a subclass of tf.data, so it fully conforms to TensorFlow's data pipeline (without any intermediate conversion), especially for TF 2.0.

For this one, I think I can have an example like the following:
```python
import tensorflow as tf
import tensorflow_io.json as json_io

dataset = json_io.JSONDataset(filename, "floatfeature", dtype=tf.float64)

iterator = dataset.make_one_shot_iterator()
next_element = iterator.get_next()

with tf.Session() as sess:
  for i in range(len(df)):
    print(sess.run(next_element))
```
> JSON Dataset could be used directly with tf.keras which is the recommended high level API for TF 2.0.
I haven't found a good example about this, the main reason is that the `JSONDataset` can only read one column to one tensor currently, and it is quite hard to find an example which build a keras model by using only one tensor. That's why I used some `zip` operations in my examples, 
any ideas about this?

yongtang:
@jiachengxu You can piece together a JSONDataset and do a zip internally, then use the zipped dataset._varient_tensor, and element_spec (or element_structure for earlier versions) to construct a Dataset easily I think.

yongtang:
@jiachengxu the following could be an example:
```python
import tensorflow as tf

d1 = tf.compat.v2.data.Dataset.from_tensor_slices([1, 2, 3, 4, 5])
d2 = tf.compat.v2.data.Dataset.from_tensor_slices([6, 7, 8, 9, 0])

class JSONDataset(tf.compat.v2.data.Dataset):
  def __init__(self, value):
    if isinstance(value, tuple):
      x, y = value

      dataset_x = x
      dataset_y = y
      self._input_dataset = tf.compat.v2.data.Dataset.zip((dataset_x, dataset_y))
      super(JSONDataset, self).__init__(self._input_dataset._variant_tensor)
    else:
      # for normal construction
      pass

  def _inputs(self):
    return []

  @property
  def _element_structure(self):
    if self._input_dataset is not None:
      return self._input_dataset._element_structure
    else:
      # for normal construction
      pass

d = JSONDataset((d1, d2))
for v in d:
  print(v)

```

tensorflow_io/json/README.md
```diff
@@ -0,0 +1,207 @@
+# TensorFlow i/o JSON Datasets
+
+[JSON (JavaScript Object Notation)](https://www.json.org/) is an open-standard
+file format that uses human-readable text to transmit data objects consisting
+of key–value pairs and array data types (or any other serializable value). JSON
+format is widely used in machine learning and data science.
+
+TensorFlow i/o JSON support mainly focuses on the **record** array in JSON, here
```
BryanCutler: nit: We normally capitalize I/O

tensorflow_io/json/README.md
```diff
@@ -0,0 +1,207 @@
...
+is an example:
```
BryanCutler: maybe describe the example like "...is an example with 3 columns and 2 records."

tensorflow_io/json/README.md
BryanCutler: It's a little vague what you mean by "mainly focuses". Are any other formats supported or does it need to be in record arrays?

tensorflow_io/json/README.md
```diff
@@ -0,0 +1,207 @@
+# TensorFlow i/o JSON Datasets
```
BryanCutler: nit: also I/O here

tensorflow_io/json/README.md
```diff
@@ -0,0 +1,207 @@
...
+
+import tensorflow as tf
+# Enable Eager Execution.
+tf.enable_eager_execution()
+import tensorflow_io.json as json_io
+
+# Assume `example.json` is a JSON format file that contains the content of the
+# example above.
+filename = "example.json"
+cols = json_io.read_json(filename)
```
BryanCutler: is this meant to be `cols = json_io.list_json_columns(filename)`?

tensorflow_io/json/README.md
```diff
@@ -0,0 +1,207 @@
...
+
+# Assume `example.json` is a JSON format file that contains the content of the
+# example above.
+filename = "example.json"
+cols = json_io.list_json_columns(filename)
+```
+
+### `read_json`
+
+`read_json` can be used to read a specified column of the JSON file into a
```
BryanCutler: does it only read one single column or one or more columns?

tensorflow_io/json/README.md
```diff
@@ -0,0 +1,207 @@
...
+
+  dataset = tf.data.Dataset.zip((
+      feature_dataset,
+      label_dataset
+  ))
+
+  dataset = dataset.map(pack_features_vector)
+
+  def pack_features_vector(features, labels):
+  """Pack the features into a single array."""
```
BryanCutler: indentation is off here

yongtang:
@jiachengxu the example I gave only works with TF 1.14. On tf-nightly or 1.15 (soon), you have to change `_element_structure` to `element_spec`. (and some small fix-ups).

jiachengxu:
> Thanks so much for the contribution @jiachengxu and for writing a usage guide, great job! I just had a few minor comments.

@BryanCutler No problem :), also thank you so much for reviewing this PR 👍

tensorflow_io/json/README.md
jiachengxu: Yeah, that's right, currently, it only supports the record arrays. I will update the PR

tensorflow_io/json/README.md
jiachengxu: Right :)

tensorflow_io/json/README.md
jiachengxu: Good catch :) 
Currently, it only read one single column, the reason for that is that I just followed the Python API of avro dataset, but now I am thinking about maybe it will be better to `zip` the tensors internally, in this way, user can read more columns and also can specify columns order.

cc @yongtang , Do you know the reason that the python API of avro dataset only allows the user to read a single column?

yongtang: @jiachengxu @BryanCutler Avro was done with single column for no particular reason, other than I was hoping tf.io could be implemented in a more column-friendly way.

As was mentioned in other threads, at the moment dataset is not truly column-oriented (more row-oriented) so it is not easy to split columns once all columns has been read into a dataset.

On the other hand, it is fairly easy to go the other route: combine multiple single-column dataset into one dataset. It could be done with `zip`, or we could easily expand an API to allow doing that (see my example in the comments of this PR).

As mentioned it is fine if the dataset is closer to destination (tf.keras), but for many data sources, additional feature engineering is needed and it is ready hard to use dataset if it is row-oriented.

jiachengxu:
@yongtang, thanks for your example, I will add it by following commits :)

jiachengxu:
@yongtang, I update the JSONDataset to put the `zip` operation internally, now JSONDataset can be built by using multiple columns:
```python
import tensorflow as tf
# Enable Eager Execution.
tf.enable_eager_execution()
import tensorflow_io.json as json_io
feature_filename = os.path.join(
    os.path.dirname(os.path.abspath(__file__)),
     "test_json",
      "iris.json")
feature_cols = ["sepalLength", "sepalWidth", "petalLength", "petalWidth"]
feature_dataset = json_io.JSONDataset(feature_filename, feature_cols)
```
which meets the **JSON Dataset could be used directly with tf.keras which is the recommended high level API for TF 2.0** I think.

Also, the example to demonstrate
**JSON Dataset is a subclass of tf.data, so it fully conforms to TensorFlow's data pipeline (without any intermediate conversion), especially for TF 2.0.**
is added.

PR #430, opened 2019-08-20
Split out the bazel build to trigger Travis CI cache.

yongtang:
Split out the bazel build to trigger Travis CI cache.


yongtang:
Cleared all Travis CI cache, try with the split.

PR #436, opened 2019-08-21
Enable all tests and builds

yongtang:


yongtang:
All tests and builds passed finally!

PR #438, opened 2019-08-23
 Add tfio.IOTensor.from_prometheus support

yongtang:
~~Note: This PR depends on PR #437~~ PR #437 has been merged.

In the past couple of months I have been trying to come up with a good example of using Prometheus with TensorFlow for infrastructure/compute usage prediction and alerting in case of anomaly. My plan was to use LSTM to combine with prometheus observation data for that.

Though there is a PrometheusDataset support already, the format of dataset is not very intuitive and hard to do it in a smooth way.

The biggest challenge is that, normally for time series data you have to setup a look back window for training, and you normalize the data (with total). Both are not easily available.

Further more, to process data in normal TF operations such as tf.roll, it requires reading the whole data into one Tenor. But this is not very straightforward as well with iterable dataset.

This PR adds tfio.IOTensor.from_prometheus which allows easily conversion the prometheus observation data into a tuple of Tensors, and do additional feature engineering.


yongtang:
@jiachengxu This PR adds prometheus support for IOTensor, though the last commit 6844b4f I added `records` mode for `from_json` that matches `records` in pandas's `read_json`. It may not cover all cases though could be a start. Please take a look.

PR #439, opened 2019-08-23
Add tfio.IOTensor.from_parquet support

yongtang:
Parquet columnar file format that naturally fits into a table/column data. Since Parquet file itself is indexable, degenerating parquet into an iterable dataset is not desirable as it loses convenience and flexibility.

This PR adds tfio.IOTensor.from_parquet support so that it is possible to acess parquet data through natual `__getitem__` operations.


yongtang:
Intend to merge this PR soon as it has been open for some time. This PR basically follow the same pattern as other PRs, and adds the ability of partitioned read.

PR #440, opened 2019-08-23
Add tfio.IOTensor.from_avro support

yongtang:
Avro is a columnar file format that naturally fits into a table/column data. Avro file itself is not directly indexable. However, it is pseudo-indexable as it consists of blocks with each blocks specifying file offset/size, and count of items. So indexing coulb be done by small range iteration.

It would be desirable to make Avro indexable as it will be much more convenient with increased flexibility.

This PR adds tfio.IOTensor.from_avro support so that it is possible to access `avro` data through natual `__getitem__` operations.


yongtang:
@BryanCutler There is a need to update Avro so I will just merge this PR for now. This PR also updated the method of using partitions to handle cached tensors in segments. This should open up the possibility of having multiple filenames (or repeat of the same dataset). I will create a follow up PR to add the multiple filenames support.

PR #441, opened 2019-08-24
Add tfio.IOTensor.from_hdf5 support

yongtang:
HDF5 file is a widely used format. It normally stores data into each named `dataset` which is a block of array with shape. It is not exactly columnar as different `dataset` in HDF5 could have different shapes unrelated to each other. From that standpoint it is more like a storage for collections of tensors (where each `dataset` represent one `tensor`).

HDF5 does allow slicing and indexing. In fact, the slicing and indexing in HDF5 are much more powerful than many other formats.

This PR adds tfio.IOTensor.from_hdf5. It treats HDF5 as a collection of BaseIOTensor which could be further used for slicing and indexing.

Note the `collection` here essentially is just a dictionary of key with BaseIOTensor as the value. It is different from Columnar IOTensor's case like Parquet or Avro.

### DIFF
--- a/tensorflow_io/core/python/ops/hdf5_io_tensor_ops.py
+++ b/tensorflow_io/core/python/ops/hdf5_io_tensor_ops.py
@@ -1,0 +1,51 @@
+# Copyright 2018 The TensorFlow Authors. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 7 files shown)


yongtang:
Plan to merge this PR soon, as all tests passed now, and this PR is less impactful to other column dataset like parquet/avro.

PR #443, opened 2019-08-24
Add tfio.IOTensor.from_csv support (experimental with Apache Arrow's CSV parser)

yongtang:
CSV file is one of the most widely used format and in TensorFlow's main repo there is already a CsvDataset which could be conviniently used for iteration, either feed into tf.keras directly, or access through a for loop.

There are, still some reasons to have a CSV input processor that give indexing and slicing access. The most notable reason is that while CSV file itself technically is only splittable (not truly indexable), in reality especially in data science by default CSV file is almost always loaded into memory. And because of its wide usage, it is really more convenient and more flexible to have a CSV processor that allows indexing and slicing.

This PR takes the indexing and slicing approach and built the parser on top of Arrow. One advantage of Arrow is that Arrow's CSV's parser options are closer to widely used pandas. This will allow easy usage of importing csv files created by pandas.

### DIFF
--- a/tensorflow_io/core/python/ops/csv_io_tensor_ops.py
+++ b/tensorflow_io/core/python/ops/csv_io_tensor_ops.py
@@ -1,0 +1,51 @@
+# Copyright 2018 The TensorFlow Authors. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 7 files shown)


yongtang:
Plan to merge this PR soon as well, it largely follows existing patterns. Could address in follow up PRs if any issue surface.

Issue #411, opened 2019-08-06
tensorflow 2.0 strategy for tensorflow-io


yongtang:
There are lots of changes in tensorflow for 2.0 recently and that effectively makes tensorflow-io build break against tf-nightly and tf-nightly-2.0-preview (see #410 for an example).

As 2.0 release is near (see https://github.com/tensorflow/io/issues/403), we have to come up with a 2.0 strategy that is feasible.

Here are the list of issue we encounter, once we bump the version from tensorflow==2.0.0beta1 to tensorflow==nightly:

1) protobuf has to be updated to 3.8.0 for TF 2.0 and stay 3.7.1 for TF 1.14.x.
2) grpc has to be updated to 1.22.0+ for TF 2.0 and stay with 1.21.1 for TF 1.14.x
3) tf.data.Dataset's base class now requires `element_spec` for TF 2.0 and stays with `element_strucutre` for 1.14.x
4) we also have to drop `output_types`, `output_shapes` pair and replace them with `TensorSpec`.
5) tf.data.Dataset's input method may need some adjustment in TF 2.0, haven't figure out exact change yet.

@terrytangyuan @BryanCutler @dmitrievanthony I am opening up this issue, as it looks like we will be forced to either stay with 1.14 or go completely to TF 2.0 very soon (in a couple of weeks).

My thinking at the moment, is that we release another version tensorflow-io 0.8.0 to stay with 1.14.x.

Then next week after tensorflow's nightly switching to 2.0 by default (https://github.com/tensorflow/io/issues/403#issue-475662297), we will switch to 2.0 by default as well, fix the above mentioned issues, and drop 1.14.x support for tensorflow-io. WDYT?

yongtang:
I am also thinking we should stick with primitive tensor ops as the upstream API is quite stable than dataset. @BryanCutler @terrytangyuan

terrytangyuan:
+1 to switch to 2.0 by default and use of primitive tensor ops.

yongtang:
@BryanCutler @terrytangyuan Release candidate of TF 2.0.0 is imminent:
https://groups.google.com/a/tensorflow.org/forum/#!topic/developers/SIfaft4wuXY

We will need to get manylinux2010 working soon.

I have been trying to build manylinux2010. I believe it is very close to get it done.

BryanCutler:
I think having a final tf-io 0.8.0 for 1.14.x would be good. After that, I agree we should switch gears to concentrate on TF 2.0.

[...]

PR #444, opened 2019-08-24
Bump version to tensorflow==2.0.0

yongtang:
Note: This PR will have to :
- Wait for 1.15RC0 is released.
- Then we can create a branch of R0.8 for 1.x support.
- Then this PR is merged to master.

### DIFF
--- a/.kokorun/io_cpu.sh
+++ b/.kokorun/io_cpu.sh
@@ -71,7 +71,8 @@ if [[ $PYTHON_VERSION == "2" ]]; then
     docker run -i --rm -v $PWD:/v -w /v --net=host buildpack-deps:18.04 bash -x -e .travis/wheel.test.sh python
 
     ## R
-    docker run -i --rm -v $PWD:/v -w /v --net=host -e GITHUB_PAT=9eecea9200150af1ec29f70bb067575eb2e56fc7 buildpack-deps:18.04 bash -x -e .travis/wheel.r.test.sh
+    # TODO: R is not compatible with eager mode.
+    echo docker run -i --rm -v $PWD:/v -w /v --net=host -e GITHUB_PAT=9eecea9200150af1ec29f70bb067575eb2e56fc7 buildpack-deps:18.04 bash -x -e .travis/wheel.r.test.sh
 
 else
     ## Python 3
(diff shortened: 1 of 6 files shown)


AlexisBRENON:
Hi,

I am starting a new ML project and hence start with TF 2.0.
I would like to be able to read Parquet data and thus would rely on tensorflow-io.

Installing it with: `pip install tensorflow-io` complains about TF incompatibility (this is normal) and I cannot import the parquet package with an error about `libtensorflow_framework.so.1` not found. In fact, I have a `libtensorflow_framework.so.2` file in the `tensorflow_core` package.

Thus, I fetch your branch `yongtang:2.0-release`, rebase it onto tensorflow/io:master and try to build from source. Build succeed, I can install the package in my virtualenv (with no warning or error). However, I still cannot import the sub-packages:
```
NotImplementedError: unable to open file: libtensorflow_io.so, from paths: ['.../venv/lib/python3.7/site-packages/tensorflow_io/core/python/ops/libtensorflow_io.so']
caused by: ['libtensorflow_framework.so.2: cannot open shared object file: No such file or directory']
```
However `libtensorflow_framework.so.2` is available at the right place (returned by `tf.sysconfig.get_lib()`).

Can I help to port tensorflow-io to TF2.0?

Thanks.

yongtang:
@AlexisBRENON The PR has been updated and it should work for 2.0 now. To build wheel packages on linux you can run download the PR and run:
```
bash -x -e .travis/python.release.sh
```

That will give you wheel files in `wheelhouse` directory (with 2.7, 3.5, 3.6, 3.7 supported).

We are looking into switch 2.0.0rc0 as the default (as opposed to stick with tf-nightly 1.x) in our nightly build. Related discussion is in https://github.com/tensorflow/io/issues/411#issuecomment-529146101

Also /cc @terrytangyuan @BryanCutler see if switching to 2.0.0rc0 in our nightly build makes sense. We can create a special branch for upcoming TF 1.15 compatible build, and release tensorflow-io 0.8.0 when TF 1.15 is available.

terrytangyuan:
I think it’s okay to switch now at least for nightly builds.

yongtang:
@terrytangyuan @BryanCutler I will merge this PR, and create a special branch R0.8 for v0.8.0 release (match tensorflow==1.15.0).

PR #446, opened 2019-08-24
Make sure ./configure.sh installs the same verion of tensorflow as setup.py

yongtang:
With multiple version of TensorFlow (1.14, 1.15RC0, 2.0.0Beta, 2.0.0RC0, and our current build of tf-nightly) in the mix, it is necessary to pin the version in `./configure.sh`.


yongtang:
It is a mechanical change so I will just merge.

PR #448, opened 2019-08-24
Update README.md with Arrow Dataset blog

yongtang:
Update README.md with Arrow Dataset blog

/cc @BryanCutler 


yongtang:
This is a README.md update with link added, I will just merge.

PR #451, opened 2019-08-26
[DICOM] Fix memory leak

suyashkumar:
This change fixes a memory leak by allocating the `DcmFileFormat` object on the stack. Previously, this was allocated on the heap and never deleted.

Note: even though `DcmFileFormat` object exists on the stack, the DICOM data read into `DcmFileFormat`'s `DcmDataset` [is on the heap](https://github.com/DCMTK/dcmtk/blob/6c8799a470804d5c46fea61b2a1f7607f5606066/dcmdata/libsrc/dcfilefo.cc#L65-L66) (so we should not have to worry about filling up the stack on reading in large images). 

The variables in the scope of this change were also renamed to adhere to Google's C++ styleguide [guidance on naming](https://google.github.io/styleguide/cppguide.html#Variable_Names), though the rest of the file does not yet adhere to this.

### DIFF
--- a/tensorflow_io/dicom/kernels/decode_dicom_image.cc
+++ b/tensorflow_io/dicom/kernels/decode_dicom_image.cc
@@ -82,19 +82,19 @@ class DecodeDICOMImageOp : public OpKernel {
     const auto in_contents_scalar = in_contents.scalar<string>()();
 
     // Load Dicom Image
-    DcmInputBufferStream dataBuf;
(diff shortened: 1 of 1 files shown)


suyashkumar:
@yongtang generally, are we aiming to adhere to [Google's C++ styleguide](https://google.github.io/styleguide/cppguide.html) in tf-io? If you think it makes sense, I can make some other readability, maintainability, and small efficiency improvements in a future PR for this dicom op as well. :)

PR #460, opened 2019-09-03
Updates for developer Dockerfile

### LINKED ISSUE #377: Adding gdb to the dev dockerfile for debugging purpose.
gdb is really necessary for debugging both python and cpp code, so it would be really useful to have gdb installed when building the dev docker image. If it is ok to add an entry of gdb, I can take care of that by following a PR.

BryanCutler:
This changes the dev image to be based on tensorflow custom-op-ubuntu16 that is manylinux2010 compatible and adds the gdb package for debugging.

Fixes #377

### DIFF
--- a/tools/dev/Dockerfile
+++ b/tools/dev/Dockerfile
@@ -1,10 +1,11 @@
-FROM ubuntu:16.04
+FROM tensorflow/tensorflow:custom-op-ubuntu16
 
 RUN apt-get update && \
     apt-get install -y \
     git \
     gcc \
     g++ \
+    gdb \
     make \
     patch \
     curl \


BryanCutler:
@yongtang I saw the tf custom-op-ubuntu16 image is available now and thought it would be good to keep our dev image based off of this. I was also able to create a manylinux2010 wheel with this, using your patch. You're more knowledgeable of the build though, so if you think we should stick to the base Ubuntu-16.04 image, that is fine too.

yongtang:
@BryanCutler The base image is ok. The only potential glitch is that custom op image is from ubuntu 16.04 but the python is actually 3.6 (different from 3.5 system default on ubuntu 16.04). That was why I initially used ubuntu 16.04 base image. If there is no additional dependency with python 3.5 then we should be fine.

Issue #215, opened 2019-05-04
Support genome data formats in tensorflow-io


yongtang:
Come across this:
https://medium.com/tensorflow/the-power-of-building-on-an-accelerating-platform-how-deepvariant-uses-intels-avx-optimizations-c8f0acb62344

Think it makes sense for tensorflow-io to support genome data formats like FASTA, FASTQ, BAM formats.

Note HDF5 seems to be used in biological data frequently as well, which is being covered in #174.

suyashkumar:
I think this could be pretty useful! The Google Brain Genomics team has released [Nucleus](https://github.com/google/nucleus): a C++/Python toolkit for reading and writing common genomics file formats. They used Nucleus to prepare data for their [DeepVariant model](https://github.com/google/deepvariant).

I very loosely [hacked around a little with integrating the two in a sandbox branch](https://github.com/tensorflow/io/compare/master...suyashkumar:s/fastq) which results in a passing bazel build, however issues arise when importing the `.so` into python (`free(): invalid pointer`, even with an empty Op body, indicative of linking/build issues, perhaps different versions of malloc). Most likely this comes from issues reconciling the build targets of Nucleus and tensorflow-io since both depend on separate instances of tensorflow core libraries (nightly vs. 2.0 rc0). Nucleus typically pulls in tensorflow source directly (which I did in my playground branch too, effectively "duplicating" tensorflow libraries for nucleus which is not ideal). Most likely we need to get Nucleus libs to build from the same @local_config_tf repository that tensorflow-io uses instead of from a separate copy of the TF source. Likely some custom `third_party` BUILD files will be needed to build the Nucleus targets. Note that I’m not at all an expert on bazel, so let me know what you think. I don’t have much more time to play around with it this weekend, but just wanted to share these thoughts. 

Ensuring version compatibility of linked tensorflow libraries needed in Nucleus vs. tensorflow-io might also be an issue longer term. 

It might make sense to consider using other lightweight libraries without additional dependencies on tensorflow that might be easier to integrate:

https://github.com/seqan/seqan3
https://github.com/OpenGene/fastp

yongtang:
@suyashkumar I played with Bazel build file and it seems with some patching around, it is possible to build nucleus (See #463).

I only have a simple example just to see if the build works or not. So #463 may not be complete.

[...]

PR #466, opened 2019-09-06
[Genomics] Add Initial FASTQ Reader Op

suyashkumar:
This change adds an initial `ReadFastq` op that reads [FASTQ nucleotide sequence files](https://support.illumina.com/bulletins/2016/04/fastq-files-explained.html). This op depends on Google Brain's [Nucleus](https://github.com/google/nucleus) library. 

This begins to address #215. 

Many thanks to @yongtang for helping with the Nucleus build and integration.


suyashkumar:
This change is still a work in progress with the following items pending (some of which can possibly come in future PRs):

- [ ] Add some additional parametrized tests
- [ ]  Look into RaggedTensor support for returning the float calculated [phred quality](http://maq.sourceforge.net/qual.shtml).
- [ ] Look into implementation one-hot encoding of nucleotides

----
@yongtang, couple questions for you:

1. Right now the `raw_quality` output is the raw ASCII representation of the quality score for each nucleotide in each sequence (there can be multiple differently sized nucleotide sequences per FASTQ file). Right now, each string in the output string tensor can be different lengths which works well. However, if we choose to return the float probabilities of the calculated quality score, it seems like it might be difficult to do that without something like a [RaggedTensor](https://www.tensorflow.org/guide/ragged_tensors), which I'm going to look more into tomorrow. Any thoughts?
Example:
```
sequences_in_file = ["ATC", "AT"]
quality_score_raw = ["!AZ", "9H"]  # not exactly, but some ascii values like this
quality_score_calculated = [[0.1, 0.2, 0.4], [0.1, 0.4]]  # how to represent outputs like this
```

2. Similar to (1), many folks use one-hot encoding representations of nucleotides in their models. We could create another op that users use by passing in their sequence string tensors into to yield the one-hot encoded version, or perform the conversion directly in this op (but we'd need to figure out how we want to return the output of differently sized `int` tensors [one per sequence in the FASTQ file], also possibly using RaggedTensors). We can also just let users handle this encoding.

One hot for nucleotides: 
```
A -> [1, 0, 0, 0]
C -> [0, 1, 0, 0]
G -> [0, 0, 1, 0]
T -> [0, 0, 0, 1]
```

An advantage of creating a separate ops for these could be useful decoupling of raw data reading and subsequent processing (which may be left to the user or not belong in tf-io).

yongtang:
@suyashkumar The PR looks great! 👍 We prefer to have C++ implementation with only the most primitive operations, and add composite operations in python. (Composite operations in python could be optimized with `@tf.function` in 2.0). I think that will provide a balance and allow user to have different needs. Different python APIs could also be exposed if necessarily.

Utilizing RaggedTensor would be great as well, as RaggedTensor has great support from TensorFlow Core repo. It could be in follow up PRs I think.

suyashkumar:
@yongtang thanks! Sounds good on putting additional transformation logic in python. Will aim to use RaggedTensors for those since they seem to be well supported with other ops too.

It seems like the build is failing during the `h5py` wheel build, which I do not think has anything to do with this PR (we might be missing a library in the container or in the bazel build?), let me know if you'd like me to look into it! Seems like the latest commit on master has a similar build failure. 

I'd say we're in a good place to merge this PR when you're ready, and I'll follow up on the items above in future PRs if that's okay since this one is getting a bit large as is. Let me know if you'd prefer me to add anything else here! :) 

Also: 
>We need to confirm that all authors are ok with their commits being contributed to this project. Please have them confirm that by leaving a comment that contains only `@googlebot I consent.` in this pull request.

I think that it might be asking for your consent since I included some of your commits in this branch (thanks again)!

yongtang:
@googlebot I consent

yongtang:
Thanks @suyashkumar  for the PR! 👍 🎉 🎆 We could always add follow up PRs to cover additional items. 

The h5py is a known issue. h5py 2.10.0 was released yesterday but it is not working with macOS, so pin to 2.9.0 is needed.

I will go ahead and merge this PR, and open a follow up PR (or have PR #465) to have the h5py fixed.

yongtang:
@suyashkumar The PR #465 has been merged into the master and the h5py issue has been resolved. The bazel version was bumped to 0.29.0 so you may need to update your bazel version as well.

suyashkumar:
@yongtang sounds great, thanks! I’ll upgrade Bazel and start working on some of the next PRs tonight or sometime this week!

yongtang:
@suyashkumar Forget to mention the default dependency to tensorflow has be switched to `tensorflow==2.0.0rc0`. (we used to pin to `tf-nightly` in the past).

Issue #453, opened 2019-08-26
Add support for FFMPEG (audio) substreams


faroit:
#49 added support for audio decoding using the ffmpeg ops. [`tf.contrib.ffmpeg.decode_audio`](https://www.tensorflow.org/api_docs/python/tf/contrib/ffmpeg/decode_audio) supported the ability to select the audio substream, which is really some applications like music separation. I would propose to also add this to tf.io.

faroit:
back then this was added by @yongtang and @carlthome

yongtang:
Added PR #494 as the first step to obtain the shape and type info of the FFmpeg audio/video info

yongtang:
PR #494 adds the initial track information parsing for ffmpeg, which is the fist step to get sub stream support. There are more issues arise while looking into ffmpeg:
1) ffmpeg is timebase, not frame based in video/audio. While it makes sense in situations where video is processed sequentially, it may not be very helpful to training where you need indexed frames accurately.
2) ffmpeg tries to fit everything into one pipeline (video/audio/subtitles) and which actually makes parsing for certain containers non-initiative. For example, it is very hard to list the keyframe which is present in some format. 

Overall, I think more work will need to be done:

- **Split the container parsing from the decoding**
   Container parsing will parse mp4/wav/etc into raw data that could be fit into different decoders. There are several advantages:
   a) Decoder ops could be reused, and combined with other container parser that are natural to system. For example, Windows/macOS or mobile devices provide manual frequently used video/audio processing in system API with appropriate license. It is really not necessarily to always go with Ffmpeg whose license have limitations.
   b) Some container formats are so frequently used, such that it is easy to just find a library to processing it in quality. And some formats such as WAV are straightforward so it is just easy to implement one.
   c) While Ffmpeg incorporates many hardware accelerated decoders (e.g., from Nvidia), they are not necessarily installed properly (or enabled properly) on all systems. Supporting some frequently used decoders with native APIs are actually much more easy. Ffmpeg has the strength of "support all format", which is actually cumbersome for certain frequently used formats.

- **Add time-based indexing**
   Instead of one indexing, believe multi-indexing (both frame-index and time-index) will greatly help the processing of video/audio/subtitle.

yongtang:
/cc @ivelin here as audio might be related to your work.

[...]

PR #474, opened 2019-09-11
Change component to takes either int64 or string

yongtang:
Change component to takes either int64 or string, which is needed to allow PR #473 to take a UUID as key.

This PR also moved around the properties and now the property is tied to an individual steam (like audio stream out of multiple streams). This should be helpful for #453 as well.


yongtang:
Also /cc @ivelin as it is related to #473.

PR #478, opened 2019-09-14
[Discussion] Add wrapper to use tf.keras.util.Sequence

yongtang:
There are multiple ways to pass data to tf.keras. While
dataset could be used, it assumes that dataset is always repeatable
which means multiple runs produce the same type of data.
But that is not necesssarily the case for streaming
where data is not repeatable.

For example, prometheus largely depends on the time to
generate the data so unless the timestamp is passes, multiple
runs will have different data produced.

To avoid confusion, this PR adds a wrapper to tf.keras.util.Sequence
so that data is guaranteed to be repeatable when data is passed to tf.keras.

### DIFF
--- a/tensorflow_io/core/python/ops/io_sequence.py
+++ b/tensorflow_io/core/python/ops/io_sequence.py
@@ -1,0 +1,72 @@
+# Copyright 2018 The TensorFlow Authors. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ==============================================================================
+"""IOSequence is a subclass of tf.keras.util.Sequence"""
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+
+import numpy as np
+
+import tensorflow as tf
+
+class IOSequence(tf.keras.utils.Sequence):
+  """IOSequence
+
+  An `IOSequence` is a subclass of tf.keras.util.Sequence
+  that could be used to pass collections of IOTensor to tf.keras
+  for training purposes.
+
+  Since IOTensor is always repeatable, multiple runs will not change
+  the underlying data, it could avoid the cases where steaming dataset
+  are not able to be used in tf.keras due to unreliable re-run.
+
+  Example:
+
+  ```python
+
+  x = tfio.IOTensor.from_tensor(tf.reshape(
+      tf.image.convert_image_dtype(x_train, tf.float32), [-1, 28, 28]))
+  y = tfio.IOTensor.from_tensor(y_train)
+
+  d_train = tfio.IOSequence(batch_size=1, x=x, y=y)
+
+  model = tf.keras.models.Sequential([
+      tf.keras.layers.Flatten(input_shape=(28, 28)),
+      tf.keras.layers.Dense(512, activation=tf.nn.relu),
+      tf.keras.layers.Dropout(0.2),
+      tf.keras.layers.Dense(10, activation=tf.nn.softmax)
+  ])
+  model.compile(optimizer='adam',
+                loss='sparse_categorical_crossentropy',
+                metrics=['accuracy'])
+
+  model.fit(d_train, epochs=5)
+  ```
+  """
+
+  def __init__(self, batch_size, x, y):
+    self.batch_size = batch_size
+    self.x = x
+    self.y = y
+
+  def __len__(self):
+    return int(np.ceil(len(self.x) / float(self.batch_size)))
+
+  def __getitem__(self, idx):
+    batch_x = self.x[idx * self.batch_size:(idx + 1) * self.batch_size]
+    batch_y = self.y[idx * self.batch_size:(idx + 1) * self.batch_size]
+    if self.batch_size == 1:
+      return tf.expand_dims(batch_x, 0), tf.expand_dims(batch_y, 0)
+    return batch_x, batch_y
--- a/tensorflow_io/__init__.py
+++ b/tensorflow_io/__init__.py
@@ -18,3 +18,4 @@ from __future__ import division
 from __future__ import print_function
 
 from tensorflow_io.core.python.ops.io_tensor import IOTensor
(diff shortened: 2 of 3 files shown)


tensorflow_io/core/python/ops/io_sequence.py
```diff
@@ -0,0 +1,72 @@
...
+  ])
+  model.compile(optimizer='adam',
+                loss='sparse_categorical_crossentropy',
+                metrics=['accuracy'])
+
+  model.fit(d_train, epochs=5)
+  ```
+  """
+
+  def __init__(self, batch_size, x, y):
```
terrytangyuan: Add doc string for the constructor? Also could y be optional for unsupervised problems (without the need of labels)?

yongtang:
Thanks @terrytangyuan, I changed the title to discussion as I think there are more underlying mechanisms that may need further input.

Also /cc @BryanCutler  @ivelin @feihugis for anyone that is interested.

When we started IO, our goal was to
1) Bring different format (e.g, video, audio, images), streaming (e.g. Kafka), file systems (e.g, cloud storage by different vendors) into tensorflow.
2) Have an easy way to move data from the above mentioned formats and storages to model training and inference. More specifically, to move data from outside to tf.keras.

The file systems (cloud storages) are straightforward, as they are pure implementation for access raw byte data. The formats and streaming are where the most of the concern arise.

The biggest part of IO is the subclass of tf.data, due to the fact that 1) most of the early tf.contrib components (e.g., Apache Kafka and Apache Ignite) are subclass of dataset before they moved to trio. 2) dataset has the best support from core tensorflow, and has the blessing of  2.0 ready recommendation for tf.keras.

When we work on adding more formats and streaming to tfio, it is clear that there are both pros and cons.

1) dataset is part of TensorFlow graph and by itself it is a data pipeline. Since TensorFlow itself is a transformation component, you really don't need any more components in you system. This greatly simplifies the overall system and reduce any additional infrastructure/devops/sre burden.
2) dataset is a natural fit for inference as dataset itself is an iterable. you can have a streaming inference system with dataset in one-pass. One example could be the combination of Kafka with TensorFlow for inference in a nice and clean way.

There are also limitations. The biggest limitation is that dataset is an iterable so you can not seek to a random position.
1) In many situations user only want to read certain part or portion of the whole file. That is not possible with dataset. (Dataset provides `drop()` and `take()` but it still needs to read from the beginning. (We have IOTensor which more or less address this issue).
2) It is less obvious, but I recently found out that when dataset is fed into tf.keras, at each epoch the dataset will rerun to get the same data. So effectively that means dataset has to be **repeatable**. If you have multiple runs of dataset, it needs to produce exactly the same output.
3) With the above 2), steaming like Kafka has to preset at the beginning an offset. For streaming like prometheus you also have to feed a constant timestamp as otherwise the whole training will not work. This seems to defeat the purpose of iterable as iterable is not the best fit for training.
4) Iterable also means you can not distribute. You can get around with fan-out but iterable itself is still a bottleneck.

We put the emphasis on indexable recently, to highlight the need to read a portion of the data at a random position, without the need to always read from beginning. We also purposely keep the IOTensor as a resource and a collection of functions. So that it is part of the TF graph as well. In this way, when an IOTensor is constructed, it will not read any data. Instead data is only being read when `__getitem__` is invoked (and only related portion of the data are touched).

With issues addressed by indexable, we still need to have a way to feed data to tf.keras. Our attempt was to convert with `to_dataset()`. But that prompt the questions of how exactly tf.keras handled data input.

After this PR is open I checked tf.keras and noticed there are more than one or two ways to feed data. tf.data is clearly one but there is also by `tensor`, by `generator`, by `Sequence`, by `iterator`.

Also, where the data is placed, is quite related. As mentioned before, when training is in place, you need multiple runs (epoch) so you need to read the whole data multiple times. In that situation, *someone* has to hold the whole data in some place (like memory). Reading data from disk multiple times in training seems not so performing. If that is true, maybe we could just pass `tensor` to tf.keras directly?

ivelin:
Yong,

Thank you for the great summary of the current state and challenges with
TFIO data sources. Your conclusions make a lot of sense.

To open a related question - has there been a discussion of formatting and
feeding data for inference differently than data for training?

Some of the areas where I see this making a difference:
- Media Streaming.
  + For live inference, we are optimizing for lowest possible latency and
maximum acceptable packet loss. Repeatability is not as interesting here.
If a chunk of audio or video arrives too late at the inference stage, then
its no use at all. Smart speakers have to be responsive to be useful.
Similarly for AR use cases.
  + Respectively for training, latency of data feed to TF is not the main
issue. Repeatability is, including repeatability of live media streaming
packet latency. However ensuring accurate, repeatable and indexed storage
requires additional processing time.
- Other real time series data - IoT sensor readings (engine status),
financial data live streaming and so on.

Since there is a lot of thought of model optimization (e.g. quantization)
for inference vs training, it might be worth considering optimizing data
sourcing for TF Lite, strictly for inference performance separately from
data sourcing for TF/Keras training purposes.

One random thought is to potentially have a base LiveDataSource type of
framework that can be fed into:
- An inference model directly
- Or a training model via an intermediate PrepareForTraining step, that
adds features needed for training into a TrainingDataSource. TF DataSet
already has a number of operators useful for training data such as
shuffling, randomization, adding noise, etc. Not sure if it can be reused
or extended to handle live streaming data.

Regards,

Ivelin

yongtang:
Thanks @ivelin. Due to the same reason, shuffling and randomization in Dataset are limited and are not *true* shuffling or randomization. Since Dataset is an iterable, it does not have the whole data (whole picture). As a result shuffling and randomization is always limited to a small cacheing chunk or a small batch.That means if your batch size is **1**, then shuffling or randomization for Dataset is no-op.

The IOTensor we had recently could have a true shuffling or randomization. It could be combined with sklearn's preprocessing so this is less of a concern.

For streaming with inference, the PR #473 is always possible to wire up with inference. Let me quickly update the PR to give you a usage example.

tensorflow_io/core/python/ops/io_sequence.py
```diff
@@ -0,0 +1,72 @@
...
+
+class IOSequence(tf.keras.utils.Sequence):
+  """IOSequence
+
+  An `IOSequence` is a subclass of tf.keras.util.Sequence
+  that could be used to pass collections of IOTensor to tf.keras
+  for training purposes.
+
+  Since IOTensor is always repeatable, multiple runs will not change
+  the underlying data, it could avoid the cases where steaming dataset
```
feihugis: nit: `steaming` -> `streaming`

yongtang:
This PR was raised to resolve the repeatable issue for training with Dataset. Initially I thought we may need to use a different mechanism (e.e.g keras.Sequence) to feed to tf.keras for training than Dataset. Now it was found out that as long as the Dataset is **repeatable** training will be OK.

#513 discussed further details. As such I think this PR could be closed as we can stay with Dataset for training, just need special handling.

PR #480, opened 2019-09-15
[Genomics] Convert DNA Sequences to One Hot encodings

suyashkumar:
**Note: This is a work in progress, a better description to come soon, along with some code cleanup, improved eager execution testing, lint, etc**

This change adds functionality to convert string tensor Genome sequences (read in using other `genome` tensorflow-io ops) into one hot encoded representations that are often used downstream to train models. This has been implemented using the `tf.function` decorator, allowing for efficient AutoGraph op optimizations of this logic.


.travis/wheel.test.sh
```diff
@@ -6,7 +6,7 @@ run_test() {
   entry=$1
   CPYTHON_VERSION=$($entry -c 'import sys; print(str(sys.version_info[0])+str(sys.version_info[1]))')
   (cd wheelhouse && $entry -m pip install tensorflow_io-*-cp${CPYTHON_VERSION}-*.whl)
-  $entry -m pip install -q pytest boto3 google-cloud-pubsub==0.39.1 pyarrow==0.14.1 pandas==0.24.2 scikit-learn==0.20.4
+  $entry -m pip install -q pytest boto3 google-cloud-pubsub==0.39.1 pyarrow==0.14.1 pandas==0.24.2 scikit-learn==0.20.4 gast==0.2.2
```
suyashkumar: We need to pin gast to 0.2.2 here and elsewhere because the current version of tf we are using did not pin this dependency appropiately causing problems: https://github.com/tensorflow/tensorflow/issues/32448#issuecomment-530628039 . This should be fixed in the next tensorflow RC! Let me know if there's a better way for us to pin this. Have we thought about spinning up a requirements.txt to hold these dependencies and pins?

suyashkumar: @yongtang let me know if you prefer a better way of pinning this dep! Thanks!

yongtang: @suyashkumar Thanks! I think that should be fine. We can always update later.

The tensorflow-io only depends on tensorflow' python requirement with no extra dependency. If needed we could add the requirement.txt in `tests/` directory.

suyashkumar:
The Mac CI tests seem to be failing for the reason below, but I cannot reproduce locally on my mac (the tests passes correctly).

>  ValueError: Tensor("Const:0", shape=(4,), dtype=int32) must be from the same graph as Tensor("ReadFastq:0", shape=(None,), dtype=string).

I am thinking it might be referring the the ONEHOT constants I defined in `genome_ops.py`, so will look into ensuring those are on the same graph.

tensorflow_io/genome/python/ops/genome_ops.py
```diff
@@ -17,6 +17,69 @@
...
+
+
+
+@tf.function
+def _nucleotide_to_onehot(nucleotide):
+  """Encodes a nucleotide using a one hot encoding."""
+  a_onehot = tf.constant([1, 0, 0, 0])
+  c_onehot = tf.constant([0, 1, 0, 0])
+  g_onehot = tf.constant([0, 0, 1, 0])
+  t_onehot = tf.constant([0, 0, 0, 1])
```
suyashkumar: I moved these tensor constants here (instead of having them as top level module constants), because they apparently must be created on the same graph as the ReadFastq sequences output. By moving these here, this eliminated the following error:
```
ValueError: Tensor("Const:0", shape=(4,), dtype=int32) must be from the same graph as Tensor("ReadFastq:0", shape=(None,), dtype=string).
```
However this is probably slightly less efficient unless tf does some optimizing under the hood for constants like these that might appear multiple times in a graph. If you know of a better way, let me know, otherwise this should be fine for now.

.travis/wheel.test.sh
suyashkumar: Ah, I see! Whenever we update our tensorflow dependency to `v2.0.0-rc1` (released 4 days ago) we should be okay without this, because they've pinned `gast` in that release https://github.com/tensorflow/tensorflow/commit/16cad8c1f6f0180b0de20abf117ed64f42f07fb6 . I imagine we will probably do that soon, and so I agree that this is likely fine for now (though if anyone uses this in the interim, they will have to pin gast or tensorflow on their own I think).

tensorflow_io/genome/python/ops/genome_ops.py
```diff
@@ -17,6 +17,72 @@
...
+  currently mapped to a error one hot encoding [1, 1, 1, 1].
+
+  Args:
+    sequences: A tf.string tensor where each string represents a DNA sequence
+
+  Returns:
+    tf.RaggedTensor: The output sequences with nucleotides one hot encoded.
+  """
+  all_onehot_nucleotides = tf.TensorArray(
+      dtype=tf.int32, size=0, dynamic_size=True)
```
suyashkumar: todo: see if I can statically calculate this size in a way that tf.function is okay with

suyashkumar:
@yongtang the Travis build is passing 🎉. See if you can give this a look whenever you have a chance, and perhaps you can force the kokoro again too? Thanks! :)

yongtang:
Thanks @suyashkumar! The PR looks good to me, and all tests pass now 👍 🎉

PR #490, opened 2019-09-18
More changes for BigQuery connector

vlasenkoalexey:
- Fixed typo in the Dockerfile
- Added NULL fields support to BigQuery connector
- Returning dataset in a form of Dictionary from BigQuery connector

@sshrdp FYI

@googlebot I fixed it. - not sure why last commit wasn't recognized.

### DIFF
--- a/tensorflow_io/bigquery/kernels/bigquery_lib.h
+++ b/tensorflow_io/bigquery/kernels/bigquery_lib.h
@@ -184,6 +184,9 @@ class BigQueryReaderDatasetIterator : public DatasetIterator<Dataset> {
         case avro::AVRO_ENUM:
           dtype = DT_STRING;
           break;
+        case avro::AVRO_NULL:
+          dtype = output_types[i];
+          break;
         default:
           return errors::InvalidArgument("unsupported data type: ",
                                          field.type());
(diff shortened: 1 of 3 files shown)


vlasenkoalexey:
@googlebot I fixed it.

yongtang:
@vlasenkoalexey I think the last commit was using your local machine's `root` user as the author. If you use `git commit --amend --reset-author` it likely will fix the issue I believe.

yongtang:
@vlasenkoalexey Looks like the CLA still has some issues, I think the `second-to-last` commit ("more style tweaks") may need to be updated.

I think if you squash the last two commit then the second-to-last commit of "more style tweaks" will collapse with  the last commit and the author issue will be fixed.

Alternatively I think the following may also work:
```sh
$ git rebase -i HEAD~2
# change the first line of "pick more style tweaks" to "edit more style tweaks" 
# then update author:
$ git commit --amend --reset-author
$ git rebase --continue
```

vlasenkoalexey:
@yongtang thanks for the pointer. Indeed looks like I committed one of the changes from Docker container and just reverting it didn't help.

PR #491, opened 2019-09-19
Wrap the IOTensor function as a callable class

yongtang:
This PR wrap the IOTensor function as a callable class, so that it is easy to manage. In the past, when IOTensor function is called, it expect several non-regular args such as component, column names, label field etc. This makes it hard to manage. 

This PR wrap the IOTensor function as a callable class and any non-input are part of the class member. At the end IOTensor expect a function of :
```
function(resource, start, stop) # or __call__(self, resource, start, stop).
```

In this way, all functions could be passed around with the same type of processing.

This PR also removed `step` arg as we could not support it anyway. 


yongtang:
This is a refactor without any API change or new addition, so will just merge shortly.

PR #494, opened 2019-09-20
Add ffmpeg substream support (partial)

yongtang:
This PR is part of the effort to support ffmpeg substream, as specified in #453.

This PR is the first step to obtain the shape and dtype of the audio/video substream in ffmpeg.


yongtang:
This PR is still partial though I think it is fine to merge it. Ffmpeg consists of video/audio/subtitle and is too complex to fit into one PR. I will create separate follow up PRs to get them in shape.

PR #497, opened 2019-09-23
Fix 1.15.0 compatibility issue

yongtang:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -83,7 +83,7 @@ jobs:
     osx_image: xcode9.3
     script:
     - bash -x -e .travis/python.release.sh --nightly ${TRAVIS_BUILD_NUMBER}
-    - echo bash -x -e .travis/wheel.test.sh
+    - bash -x -e .travis/wheel.test.sh
     after_success: bash -x -e .travis/after-success.sh
 
   - stage: build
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -83,7 +83,7 @@ jobs:
     osx_image: xcode9.3
     script:
     - bash -x -e .travis/python.release.sh --nightly ${TRAVIS_BUILD_NUMBER}
-    - echo bash -x -e .travis/wheel.test.sh
+    - bash -x -e .travis/wheel.test.sh
```
terrytangyuan: I thought you just changed this in a previous PR. Why is this reverted?

yongtang: @terrytangyuan This is a way to get the Travis CI cache in place. Our macOS build, without Travis CI cache, will run more than 50 minutes.

With Travis CI cache, the execution time is about 20minutes. 

However, Travis CI cache will not be properly populated if the execution is terminated when 50min limit is reached. That means Travis CI cache will never be able to build up properly.

So in the previous PR I comment out the test which reduce the execution time to about 40min. Then Travis CI cache is build up properly, and if we revert and enable the test again, the cache is in place (see result where all tests finish in around 20min).

This is a not a good way, though it is what we have at the moment.

Ideally, if we are able to setup a Bazel cache server then cache could be build adaptively I think. We don't have a remote Bazel cache server so we are limited.

PR #503, opened 2019-09-25
Performance tweak for BigQuery connector

vlasenkoalexey:
- Moving expensive GenericDatum initialization out of ReadRow to be called once per dataset initialization
- Removing implicit conversions to String


vlasenkoalexey:
@sshrdp FYI

vlasenkoalexey:
Sure, first I tried using tensorflow_io/avro when building reader, but it didn't work for me as is.
But eventually it would be nice to have one AVRO parser that works for all.
And besides current implementations only support flat rows (no nested fields).

Issue #507, opened 2019-09-26
Add Avro Deserializer for Kafka in tensorflow-io


yongtang:
This is an issue to track the progress for adding Avro deserializer for Kafka in tensorflow-io.

Initially I though Avro deserializer could be applied to Kafka directly though the raw bytes I received does not exactly fit a generic datum. There might be some additional framing in between. So decided to setup a Kafka Schema Registry to check for any raw bytes to resolve this issue.

Here is the list of items:
- [x] Change test setup in our Travis CI to use confluent kafka + Schema Registry (PR #508)
- [ ] Add test schema to Schema Registry (PR #512)
- [ ] Add Avro deserializer for kafka message parsing.  (PR #512)

/cc @kaiwaehner

kaiwaehner:
Thanks for investing time into this topic, @yongtang .

Please note that there are **two options for Avro and Kafka:**

1) Most people use **KafkaAvroSerializer** (see https://docs.confluent.io/current/schema-registry/serializer-formatter.html), which is implement by Confluent and open source on Github. It uses a magic byte in each message as Schema ID. Therefore, this KafkaAvroSerializer is only working if you also use Confluent Schema Registry (https://github.com/confluentinc/schema-registry). 

2) You could also **write your own Avro Serializer** (or use another open source implementation). This does not use the magic byte and does not know Schema Registry. 

Most projects I have seen so far use 1) because the Schema Registry adds a lot of value to most projects (and it is free and available on Github). Thus, I think it is the right way to go forward here, too.
I just wanted to explain both options and the differences because your statement "Initially I though Avro deserializer could be applied to Kafka directly" is actually correct - but you can either do 1) or 2), not a mix of both...

kaiwaehner:
Another comment regarding your task "_Change test setup in our Travis CI to use confluence kafka + Schema Registry_"

First of all, it is "Confluent", not "Confluence" (which is the wiki tool). No worries, many people use the wrong name sometimes :-)

More important, there is no "Confluent Kafka". This is another thing many people say wrongly. 
You can use the Confluent Platform which includes Apache Kafka (and additional optional components, e.g. Schema Registry). 

This is an important difference to clarify that Confluent also uses exactly the same Apache Kafka as you download from the Apache website.

yongtang:
Thanks for the pointers @kaiwaehner! I am terribly sorry to mistype the `confluent` :flushed: ( I did type the name right in PR #507, though. :) 

[...]

PR #508, opened 2019-09-26
Update Kafka to use Confluent version 5.3.1, and start Schema Registry

yongtang:
This is part of the effort in #507


yongtang:
Linux works though macOS still need some fix in Travis CI.

PR #516, opened 2019-09-30
Separation of IODataset and IOStreamDataset

### LINKED ISSUE #513: Separation of bounded (and repeatable) and unbounded (or not repeatable) Dataset
For tensorflow-io we always start out IO related implementation with Dataset. The reason is to tie to TF 2.0 overall and tf.keras. In the past we only have limited input sources covered and everything was fine. However, as different input sources are exposed some issues arise.

While Dataset has been well supported in tf.keras, the mechanism of inference and training are different and that caused confusions and issues when a standard Dataset was passed to tf.keras and some functionalities stop working. To summarize:
1) Dataset is an iterable and could be unbounded (no EOF) and non-repeatable (re-run does not have to get the same result).
2) Inference could take any Dataset as input
3) Training could take a Dataset that is iterable. However, training also requires that
    a) The Dataset has to be bounded (with an EOF)
    b) The Dataset has to be repeatable (re-run will get the same result), as the Dataset node will be recreated at each epoch.

The issue was observed when Kafka dataset was passed and starting offset was not set. In that case re-run of Dataset causes data inconsistency across different epoch runs. However, Kafka itself is repeatable as long as the offset has been provided (as Kafka has an excellent offset management). In fact Kafka is even indexable and you can build even SQL on top of it. 

One counter example of Dataset is to read data from standard input (or network packet). There is no way you can repeat (unless caching the data on disk). In that case reading standard input can only be used for inference.

There were lots of confusion about Dataset and users encounter many issues as far as I know.

Since tensorflow-io intends to cover many input sources, I think it is best for us to split out the concept of repeatable and bounded Dataset, from the non-repeatable or non-bounded Dataset.

I am thinking (in combination to top level API exposure) about the following approach:
1) Expose an tfio.IODataset which is reserved for repeatable and bounded Dataset, and let user know IODataset is safe to be used with training.
2) Expose an tfio.StreamIODataset which is used for Dataset that is not safe to pass to training. Examples include standard input, grpc stream, from video caputuring and audio recording devices, etc. We also should let user know StreamIODataset can only be used for inference or iterate with `for d in dataset`.

In addition, I think we could follow the similar API of factory and have:
```
tfio.IODataset.from_parquet
tfio.IODataset.from_kafka
tfio.IODataset.from_ffmpeg
```

Note: as was discussed, Kafka could be either training safe (when offset is provided), or inference only (in a continuous fashion). From that standpoint we should have
```
tfio.IODataset.from_kafka

tfio.StreamIODataset.from_kafka 
```
to cover different scenarios.

yongtang:
As was discussed in #513, this PR separate IODataset and IOStreamDataset, so that user is aware of the implications for different types of Dataset: repeatable Dataset (IODataset) is suitable for both training and inference, while non-repeatable/stream(IOStreamDataset) is only suitable for inference.

The API has also be consolidated so that we expose top level factory method, e.g.,
```
IODataset.from_ffmpeg
IOStreamDataset.from_kafka
```

This PR fixes #513


yongtang:
Added additional update so that now Avro/Parquet/JSON could specify a list of columns selected for dataset. This brings the new kernel in parity with the old handcrafted C++ one.

Issue #431, opened 2019-08-20
tf-nightly failure cases


yongtang:
Since there are several recent changes in tf.data, the following has been disabled temporarily:

- Ignite tests has been temporarily disabled (9c8dce4)
- Arrow tests has been temporarily disabled (b704dee)
- Arrow has to comment out the optimization_only usage (4e6718c)

We may want to fix those before 0.8.0 release is out.

BryanCutler:
I'm looking into the Arrow test failures now

BryanCutler:
Apologies for the delay, @yongtang . I believe I have fixed the arrow test failures now.

BryanCutler:
Is anyone looking at the Ignite test failures? I can give it a shot if no one else is..

yongtang:
@BryanCutler If you can take a look at ignite test failure that would be great. I think we are very close to release the new version soon.

yongtang:
With PR #519 and #526 merged, this issue could be closed.

PR #519, opened 2019-09-30
Enable and Fix Arrow tests for TF 1.15.0

BryanCutler:
This corrects API issues with TF 1.15.0 and fixes a bug in reading ArrowStreamDataset messages causing test failures. Arrow tests are now all enabled.

### DIFF
--- a/tensorflow_io/arrow/python/ops/arrow_dataset_ops.py
+++ b/tensorflow_io/arrow/python/ops/arrow_dataset_ops.py
@@ -27,8 +27,7 @@ import tempfile
 
 import tensorflow as tf
 from tensorflow import dtypes
-from tensorflow.compat.v2 import data
-from tensorflow.python.data.ops.dataset_ops import flat_structure
+from tensorflow.python.data.ops import dataset_ops
 from tensorflow.python.data.util import structure as structure_lib
 from tensorflow_io.core.python.ops import core_ops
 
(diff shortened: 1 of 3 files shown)


BryanCutler:
Fixes Arrow test failures from #431 
cc @yongtang @terrytangyuan

tensorflow_io/arrow/python/ops/arrow_dataset_ops.py
```diff
@@ -88,7 +87,7 @@ def arrow_schema_to_tensor_types(schema):
   return tensor_types, tensor_shapes
 
 
-class ArrowBaseDataset(data.Dataset):
+class ArrowBaseDataset(dataset_ops.DatasetV2):
```
BryanCutler: Is this ok to use as a base for 1.15.0?

yongtang: That should be OK I think. As long as dataset could be passed to tf.keras (with 1.15 and 2.0) we should be fine. Both DatasetV1 and DatasetV2 works with tf.keras. 👍

yongtang:
Actually it might be easier to PR against master, then cherry-pick or rebase in R0.8 instead. As the Google CLA will post some issues in R0.8 branch. Let me fix that.

yongtang:
Cherry-picked to master and fixed the CLA issue.

BryanCutler:
Thanks @yongtang !

PR #526, opened 2019-10-01
Fix IgniteDataset test failure

### LINKED ISSUE #431: tf-nightly failure cases

BryanCutler:
Fix Ignite test failure and enable tests again. The failure was due to using the IgniteDataset as the return value for the `Dataset.interleave` function and seems that TensorFlow is unable to build the `TypeSpec` for graph mode. The workaround is to define the IgniteDataset outside the lambda function for `interleave`.

Fixes #431

### DIFF
--- a/tests/test_ignite.py
+++ b/tests/test_ignite.py
@@ -24,9 +24,6 @@ import pytest
 import tensorflow as tf
 tf.compat.v1.disable_eager_execution()
 
-pytest.skip(
-    "ignite test is disabled temporarily", allow_module_level=True)
-
 from tensorflow import dtypes            # pylint: disable=wrong-import-position
 from tensorflow import errors            # pylint: disable=wrong-import-position
 from tensorflow import test              # pylint: disable=wrong-import-position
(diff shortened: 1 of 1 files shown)


tests/test_ignite.py
```diff
@@ -307,12 +304,20 @@ def test_ignite_dataset_with_plain_client_with_interleave(self):
...
     """
     self._clear_env()
+
+    igds_local = ignite_io.IgniteDataset(
+        cache_name="SQL_PUBLIC_TEST_CACHE",
+        schema_host="localhost",
+        host='localhost',
+        port=10800)
+
+    # TODO: this is a workaround due to failure to build a TypeSpec for
```
BryanCutler: @yongtang I think this should have technically worked before, but the failure seems due to being in non-eager mode. After the 0.8.0 release, we will probably want to change these over to eager mode, so this is just a temporary fix. WDYT?

PR #538, opened 2019-10-13
Expose __version__ in tensorflow_io

### LINKED ISSUE #482: Add __version__ to the package
While trying to build docs in #481, noticed that our packages misses a `__version__` symbol. We should be able to find a way to embed the `__version__` symbol into the package in setup.py I think.

yongtang:
This PR exposes __version__ in tensorflow_io, as was discussed
in #482. The single source of truth is located in
tensorflow_io/core/python/ops/io_info.py

so that any change only happens in one place.

A test case has been added as well to guarantee the correctness of the __version__
string for package release.

This PR fixes #482.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -180,12 +180,24 @@ setup(
 )
 """
 
-package = 'tensorflow>=2.0.0,<2.1.0'
-version = '0.9.0'
-project = 'tensorflow-io'
+# read package and version from:
+# tensorflow_io/core/python/ops/io_info.py
+with open("tensorflow_io/core/python/ops/io_info.py") as f:
+  entries = [e.strip() for e in f.readlines() if not e.startswith("#")]
+  assert sum(e.startswith("package = ") for e in entries) == 1
+  assert sum(e.startswith("version = ") for e in entries) == 1
+  package = list([
+      e[10:] for e in entries if e.startswith("package = ")])[0].strip("'")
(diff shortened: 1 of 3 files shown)


setup.py
```diff
@@ -180,13 +180,25 @@ def get_outputs(self):
 )
 """
 
-package = 'tensorflow>=2.0.0,<2.1.0'
-version = '0.9.0'
-project = 'tensorflow-io'
+# read package and version from:
+# tensorflow_io/core/python/ops/io_info.py
```
terrytangyuan: Why are we moving this to that file?

yongtang: @terrytangyuan We need `__version__` string in two places:
1) It needs to be in a python file shipped in wheel package so that in runtime it is exposed. 
2) It also need to be in setup.py so that it is inserted in wheel's egg.

However, setup.py will not be shipped by wheel package directly so we could not put the string in setup.py (we also don't want to place two places for the purpose of single source of truth).

So we move the string to this `tensorflow_io/core/python/ops/io_info.py` file:
1) In runtime, tensorflow_io/__init__.py will import the string `__version__`.
2) setup.py also needs the string when package the wheel file. In this case, we just read the file directly to grab the string and pass to setup.py.

terrytangyuan: Got it

PR #543, opened 2019-10-13
Implicitly import azure/oss file systems

yongtang:
This PR made several changes so that azure/oss file system could be implicitly registered as long as tensorflow_io is imported:
```
import tensorflow_io as tfio
```

This should help consolidate the API exposure.

### DIFF
--- a/tensorflow_io/__init__.py
+++ b/tensorflow_io/__init__.py
@@ -17,6 +17,7 @@ from __future__ import absolute_import
 from __future__ import division
 from __future__ import print_function
 
+# tensorflow_io.core.python.ops is implicitly imported (along with file system)
 from tensorflow_io.core.python.ops.io_info import version as __version__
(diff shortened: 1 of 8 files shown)


yongtang:
@terrytangyuan I don't have access to css file system. If possible, can you help do a test locally if you have access?

With this PR, as long as `import tensorflow_io` is in place all file system will be available. 

So the following will work I believe (also see https://github.com/tensorflow/io/blob/0ce5d9de61ff9345d46620bc414f7ed1324e500c/tensorflow_io/oss/README.md) :
```
import tensorflow as tf
import tensorflow_io as tfio

tf.io.gfile.mkdir('oss://${bucket}\x01id=${access_id}\x02key=${access_key}\x02host=${host}/test_dir')
```

That should avoid unnecessary expose of API/modules.

PR #545, opened 2019-10-13
Move all libraries under libtensorflow_io.so

yongtang:
This PR is the effort to move all libraries under  libtensorflow_io.so,
so that there is only one copy of grpc (from google cloud etc).

The reason is that when multiple copies exists with different .so,
issues arise.


yongtang:
/cc @vlasenkoalexey I moved some of the grpc related libraries into one `.so` (`libtensorflow_io.so`), namely bigquery, gcs_config, grpc. I think this could help avoid multiple copies of grpc (in different .so) issue.

I also tried with moving `bigtable` and `cloud` into this one `.so` as well, though it looks like there are some duplicated class/functions in between. I think they could be moved at a later time.

vlasenkoalexey:
Looks good

PR #547, opened 2019-10-14
[WIP] Expose tfio.experimental.image.decode_tiff

yongtang:
This is part of the effort to rethink about namespace management
in tensorflow-io, as was discussed in 537 and RFC tensorflow/community/135

This is the first op introduced, we could move more to tfio.experimental.

There are several reasons for having a tfio.experimental namespace:
1) Some API are not stable enough or we are not sure what is the best way to expose it.
2) Some API has been implemented but not generic enough, e.g, we process image as uint8 though we  know 16 bit or higher format is possible.

This PR is to start tfio.experimental with decode_tiff for sveral reasons:
1) TIFF could have multiple frames with different shapes, that adds complexity
2) TIFF could have higher resolution than uint8, currently it is not handled yet (See 534).

As such decode_tiff could be a candidate of tfio.experimental. In the future once
implementation is complete and solde we could promote and remove `experimental`.

### DIFF
--- a/tensorflow_io/core/python/ops/experimental/__init__.py
+++ b/tensorflow_io/core/python/ops/experimental/__init__.py
@@ -1,4 +1,4 @@
-# Copyright 2018 The TensorFlow Authors. All Rights Reserved.
+# Copyright 2019 The TensorFlow Authors. All Rights Reserved.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


yongtang:
We are going to release 0.8 and 0.9 really soon, let me put this PR on hold for now. We can continue the discussion after 0.8 and 0.9 release.

yongtang:
This PR is superseded by PR #605.

PR #540, opened 2019-10-13
PR #510: Add prefix for all implemented ops in .cc modules

ricky3dec:
1. Add prefix to all the ops under tensorflow_io package.
2. Update copyright year.
3. Correct some spaces.

PR #549, opened 2019-10-14
Follow up PR #540

yongtang:
This is a follow up to PR #540  /cc @ricky3dec 


ricky3dec:
@yongtang thanks for working on this fix, looks like I should have re-checked changes after the merge from master. I shall take care of it next time.

That brings to me another point, could you please share the build and test steps reference, so that I can do some sanity checks on my end?
Also, I think we should document these naming conventions in development guide, so that anyone contributing can follow them. I know it's not fool proof, but still some guidance would be there. Just a suggestion!

Issue #459, opened 2019-08-27
tensorflow-io 0.8.0 release


BryanCutler:
This is to start the planning for the release of 0.8.0 for tensorflow-io.

Here is the list of items for 0.8.0:

- [x] Fix test errors in #431 
- [x] Update the version in setup.py
- [x] Update README.md to add additional supported ops since 0.7.0
- [x] Update RELEASE.md to capture 0.8.0 changes and version compatibility.
- [x] Push *.whl files to PyPI.org, including macOS package.
- [x] Release R package to CRAN (optionally).

/cc @yongtang @terrytangyuan @dmitrievanthony

yongtang:
Created a branch R0.8, and PR #470 as a working-in-progress. The tensorflow==1.15.0 was supposed to be rebased last month though it looks like there is some delay. So we may have to wait.

In the meantime, we are switching master and nightly build to use tensorflow==2.0.0rc0, but keep R0.8 branch to use tf-nightly (latest version was from 08/21).

yongtang:
The new PR is in #472

BryanCutler:
I think we need to fix the test errors in https://github.com/tensorflow/io/issues/431 before release also

yongtang:
As https://github.com/tensorflow/tensorflow/pull/33263 has been merged, I expect that TensorFlow 1.15.0 to be released in a couple of days (Monday or Tuesday).

We are good with PR #472 and is pretty ready for our v0.8.0 release.

yongtang:
All tests pass now. The candidate wheel files for 0.8.0 release are located in:

https://www.dropbox.com/home/Apps/tensorflow-io-nightly/release/1990/wheelhouse

Will do some testing, if everything is fine, will label v0.8.0 release then push to PyPI.org (likely this afternoon).

yongtang:
Tested on Ubuntu 16.04 and 18.04, all works fine (python 2.7, 3.5, 3.6, 3.7). Below are the related wheel files:
```
5a4a70ad39e3147b20ac40b5c22dca9064f350e70774bd93e4a343e1b65617e4  tensorflow_io-0.8.0-cp27-cp27m-macosx_10_13_x86_64.whl
45ba91020b248181181ffa7d218e91791baef35e92936a78a4c6f4e5da5684e5  tensorflow_io-0.8.0-cp27-cp27mu-manylinux2010_x86_64.whl
[...]

PR #555, opened 2019-10-17
Update for 0.9.0 release

yongtang:


yongtang:
/cc @terrytangyuan @BryanCutler @vlasenkoalexey, with 0.8.0 released in #459, now 0.9.0 should be ready for review.

I think we could release 0.9.0 tomorrow (so that to create one-day difference for two versions), if everything goes well.

PR #561, opened 2019-10-19
HTTP file system support

### LINKED ISSUE #137: http/https filesystem support
We already have support for filesystems with `igfs://`, `gs://`, `s3://` (either through tensorflow-io or through tensorflow repo). One protocol that is missing is the `https://` or `http://`. There are lots of data are referring an image (or video/audio) through the URL so it really would be great if we could provide support for `https://` or `http://`.

While HTTP protocol supports `Accept-Ranges` header, not all web servers provides this feature. So this has to be taken into consideration. Also, there is no need to re-download the file again if already save to local disk. It would make sense to have a local cache similar to pip or bazel.

The implementation likely will be in stages. We could have an early version first then expand features as needed.

yongtang:
This PR is an initial effort to have HTTP file system, as was mentioned in #137.

This PR added two functions, Read and GetFileSize, which should be enough
for most RandomAccess file related  operations.

NOTE:
1) GetFileSize replies on "Content-Length" header. If this header is not available, then file system fails.
2) Read relies on "Range" header. If this header is not supported, then file system fails.

We could build up additional functionalities and fine tuning later.

This PR fixes #137.

### DIFF
--- a/tests/test_http/LICENSE-2.0.txt
+++ b/tests/test_http/LICENSE-2.0.txt
@@ -1,0 +1,202 @@
+
+                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
+
+   TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
+
+   1. Definitions.
+
+      "License" shall mean the terms and conditions for use, reproduction,
+      and distribution as defined by Sections 1 through 9 of this document.
+
+      "Licensor" shall mean the copyright owner or entity authorized by
+      the copyright owner that is granting the License.
+
+      "Legal Entity" shall mean the union of the acting entity and all
+      other entities that control, are controlled by, or are under common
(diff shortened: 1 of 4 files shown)


tests/test_http/LICENSE-2.0.txt
```diff
@@ -0,0 +1,202 @@
+
+                                 Apache License
```
terrytangyuan: Is this file accidentally included or intentional?

yongtang: Oh, this file is a `local test file`. I use `http://www.apache.org/licenses/LICENSE-2.0.txt` for testing http file system. So I saved a local copy of `LICENSE-2.0.txt` to compare remote and local version of:

`tf.read_file("http://www.apache.org/licenses/LICENSE-2.0.txt")`

`tf.io.read_file("test_http/LICENSE-2.0.txt")`

PR #565, opened 2019-10-19
Merge pcap_ops.so into libtensorflow_io.so

yongtang:
to reduce the number of .so (and avoid potential duplicate copies of lib in multiple .so).
also expose tfio.IODataset.from_pcap().

### DIFF
--- a/tests/test_pcap_eager.py
+++ b/tests/test_pcap_eager.py
@@ -39,7 +39,10 @@ def test_pcap_input():
   for v in dataset:
     (packet_timestamp, packet_data) = v
     if packets_total == 0:
-      assert np.isclose(packet_timestamp.numpy()[0], 1084443427.311224) # we know this is the correct value in the test pcap file
+      assert np.isclose(
+          packet_timestamp.numpy()[0],
+          1084443427.311224,
+          rtol=1e-15) # we know this is the correct value in the test pcap file
       assert len(packet_data.numpy()[0]) == 62 # we know this is the correct packet data buffer length in the test pcap file
     packets_total += 1
   assert packets_total == 43 # we know this is the correct number of packets in the test pcap file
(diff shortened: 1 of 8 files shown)


tests/test_pcap_eager.py
```diff
@@ -34,8 +31,19 @@ def test_pcap_input():
...
-  url_filenames = [file_url]
-  dataset = pcap_io.PcapDataset(url_filenames, batch=1)
+
+  dataset = tfio.IODataset.from_pcap(file_url, capacity=5).batch(1)
+
+  packets_total = 0
+  for v in dataset:
+    (packet_timestamp, packet_data) = v
+    if packets_total == 0:
+      assert packet_timestamp.numpy()[0] == 1084443427.311224 # we know this is the correct value in the test pcap file
```
BryanCutler: better to use `numpy.is_close`?

yongtang: @BryanCutler updated. Thanks.

tests/test_pcap_eager.py
```diff
@@ -34,8 +32,19 @@ def test_pcap_input():
...
-  url_filenames = [file_url]
-  dataset = pcap_io.PcapDataset(url_filenames, batch=1)
+
+  dataset = tfio.IODataset.from_pcap(file_url, capacity=5).batch(1)
+
+  packets_total = 0
+  for v in dataset:
+    (packet_timestamp, packet_data) = v
+    if packets_total == 0:
+      assert np.isclose(packet_timestamp.numpy()[0], 1084443427.311224) # we know this is the correct value in the test pcap file
```
BryanCutler: since it's a big number you might want to add the `rtol` param to something like `rtol=1e-15`

yongtang: @BryanCutler Updated. Thanks.

PR #567, opened 2019-10-20
Add HDR image format support

yongtang:
This PR add HDR image format (RGBE or Radiance HDR) support.
Part of the effort is to implement a `decode_hdr` that extract
HDR format into float rgb channels.


yongtang:
This PR has been opened for some time, and is a relatively less impactful one as it is only a basic ops. Intend to merge if no objections.

PR #571, opened 2019-10-21
fixing typo in .travis/bazel.build.sh

vlasenkoalexey:


yongtang:
I will create a follow up PR to get the rest of the fix in place.

Issue #489, opened 2019-09-18
Missing imported modules


yongtang:
See #487 for more details. At the moment some of the modules in tensorflow-io are missing when importing:

- bigquery
```
from tensorflow_io.bigquery.python.ops.bigquery_api import BigQueryClient
ModuleNotFoundError: No module named 'tensorflow_io.bigquery.python.ops.bigquery_api'
```
- cloud
```
  File "/usr/local/google/home/markdaoust/venv3/lib/python3.6/site-packages/tensorflow_io/cloud/__init__.py", line 22, in <module>
    from tensorflow_io.cloud.python.ops.bigquery_reader_ops import *
  File "/usr/local/google/home/markdaoust/venv3/lib/python3.6/site-packages/tensorflow_io/cloud/python/ops/bigquery_reader_ops.py", line 25, in <module>
    _bigquery_reader_so = _load_library("_bigquery_reader_ops.so")
  File "/usr/local/google/home/markdaoust/venv3/lib/python3.6/site-packages/tensorflow_io/__init__.py", line 59, in _load_library
    "{}, from paths: {}\ncaused by: {}".format(filename, filenames, errs))
NotImplementedError: unable to open file: _bigquery_reader_ops.so, from paths: ['/usr/local/google/home/markdaoust/venv3/lib/python3.6/site-packages/tensorflow_io/cloud/python/ops/_bigquery_reader_ops.so']
caused by: ['/usr/local/google/home/markdaoust/venv3/lib/python3.6/site-packages/tensorflow_io/cloud/python/ops/_bigquery_reader_ops.so: undefined symbol: _ZTVN10tensorflow10ReaderBaseE']
```
- dicom
```
  File "/usr/local/google/home/markdaoust/venv3/lib/python3.6/site-packages/tensorflow_io/dicom/__init__.py", line 26, in <module>
    from tensorflow_io.dicom.python.ops.dicom_ops import decode_dicom_data
  File "/usr/local/google/home/markdaoust/venv3/lib/python3.6/site-packages/tensorflow_io/dicom/python/ops/dicom_ops.py", line 25, in <module>
    decode_dicom_data = dicom_ops.decode_dicom_data
AttributeError: module '0bb84602ad9137f5e6d7401094b24438' has no attribute 'decode_dicom_data'
```

We will need to fix those bugs. Thanks @MarkDaoust for providing the information!

Kat-ML:
Have you already fixed those bugs?

yongtang:
[...]

PR #587, opened 2019-10-24
Add `linkstatic = True` to grpc to remove the unneed .so build

yongtang:
grpc has been built into libtensorflow_io.so and the libendpoint_cc_grpc.so
is not needed for wheel to run. This also causes the issue in #489.

This PR adds `linkstatic = True,` to skip the building of libendpoint_cc_grpc.so

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -245,6 +245,8 @@ if datapath is not None:
       for filename in [
           f for f in filenames if fnmatch.fnmatch(
               f, "*.so") or fnmatch.fnmatch(f, "*.py")]:
+        if filename.endswith("_proto.so"):
+          continue
         src = os.path.join(rootname, filename)
         dst = os.path.join(
             rootpath,
--- a/setup.py
+++ b/setup.py
@@ -245,6 +245,18 @@ if datapath is not None:
       for filename in [
           f for f in filenames if fnmatch.fnmatch(
               f, "*.so") or fnmatch.fnmatch(f, "*.py")]:
+        # NOTE:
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -245,6 +245,8 @@ def get_outputs(self):
       for filename in [
           f for f in filenames if fnmatch.fnmatch(
               f, "*.so") or fnmatch.fnmatch(f, "*.py")]:
+        if filename.endswith("_proto.so"):
+          continue
```
BryanCutler: Could you note how this is related to the grpc package? Also, is it possible to match a pattern with "grpc" in the filename, just in case some other package has a dependency on a library also ending in "proto"?

yongtang:
@BryanCutler The PR has been updated with comment added. Below is the comment part:
```python
        # cc_grpc_library will generate a lib<name>_cc_grpc.so
        # proto_library will generate a lib<name>_proto.so
        # both .so files are not needed in final wheel.
        # The cc_grpc_library only need to pass `linkstatic = True`
        # to the underlying native.cc_library. It is not exposed
        # but we applied a patch (see third_party/grpc.patch) so
        # cc_grpc_library is covered and lib<name>_cc_grpc.so will
        # not be generated.
        # proto_library is a native library in bazel which we could
        # not patch easily.
        # For that reason we skip lib<name>_proto.so here:
        if filename.endswith("_proto.so"):
          continue
```

Please take a look and let me know if there are any questions.

setup.py
BryanCutler: I see, that explains it very well, thanks!

PR #601, opened 2019-11-01
Use balanced sharding strategy in bigquery read session.

sshrdp:
We use sharding to provide a way of shuffling the read order. With liquid sharding, a single stream can read all data which wouldn't be shuffled.


sshrdp:
+vlasenkoalexey FYI

vlasenkoalexey:
Looks good

PR #602, opened 2019-11-02
^^

CerberusDevOps:


yongtang:
@CerberusDevOps Azure Pipeline is a nice CI. However, it requires write permission and will directly write to the repo. To conform to tensorflow org's policy we could not accept related PRs.

I will close this PR but thanks for your contribution!

PR #604, opened 2019-11-03
Add decode_jpeg_exif to support extract orientation information

yongtang:
This PR tries to address the issue raised in  tensorflow/tensorflow#33932
where decode_jpeg does not taking into account the Exif information.

This PR adds decode_jpeg_exif which returns orientation of the image
if possible.

Further enhancements for the following could be added if there is enough
interests:
1) Expands to adds more fields of exif information
2) Add another kernel (or a python function reusing existing C++ kernel)
   to automatically adjust EXIF orientation.


tensorflow_io/core/python/experimental/image_ops.py
```diff
@@ -40,6 +40,21 @@ def draw_bounding_boxes(images, boxes, texts=None, colors=None, name=None):
   return core_ops.io_draw_bounding_boxes_v3(
       images, boxes, colors, texts, name=name)
 
+def decode_jpeg_exif(contents, name=None):
+  """
+  Decode Exif information from an JPEG image.
+
+  TODO: Add additional fields than just orientation.
```
terrytangyuan: "than just" -> "besides"

yongtang: @terrytangyuan PR updated. Thanks.

PR #608, opened 2019-11-04
Remap mnist to tfio.IODataset.from_mnist for API consistency

yongtang:
Remap mnist to tfio.IODataset.from_mnist for API consistency

Also update README.md, to emphasis the ease of usage in MNIST example with HTTP file system.


yongtang:
By the way, since we already support `http` file system now, it is possible to show the convenience of tensorflow-io with the following sample code. Note there is no need to copy files to local disk anymore, you can just use `http` directly.


```python
import tensorflow as tf
import tensorflow_io as tfio

# Read MNIST into Dataset
d_train = tfio.IODataset.from_mnist(
    'http://yann.lecun.com/exdb/mnist/train-images-idx3-ubyte.gz",
    'http://yann.lecun.com/exdb/mnist/train-labels-idx1-ubyte.gz').batch(1)

# By default image data is uint8 so conver to float32.
d_train = d_train.map(lambda x, y: (tf.image.convert_image_dtype(x, tf.float32), y))

model = tf.keras.models.Sequential([
  tf.keras.layers.Flatten(input_shape=(28, 28)),
  tf.keras.layers.Dense(512, activation=tf.nn.relu),
  tf.keras.layers.Dropout(0.2),
  tf.keras.layers.Dense(10, activation=tf.nn.softmax)
])
model.compile(optimizer='adam',
              loss='sparse_categorical_crossentropy',
              metrics=['accuracy'])

model.fit(d_train, epochs=5, steps_per_epoch=10000)
```

Issue #611, opened 2019-11-06
KafkaDataset enhancement


yongtang:
At the moment we are actively working on having the KafkaDataset combining with schema registry, pipeline, message write back, etc. to reflect production usage. We noticed several areas that could be improved:

- [x] In order to support schema registry, we use `decode_avro` messages. However, this is not really efficient because `decode_avro` takes the `schema` as input and parse the schema repeatedly. Instead, we should create a `DT_RESOURCE` handle to persists the blob of preprocessed schema. In this way, we could skip the schema parsing in `decode_avro`.
   [PR #628]

- [x] In order to support the maximum range of users, we use the most generic CPU instructions when compiling C++ code. This ensures user will not encounter any invalid CPU instructions if their machine is old. However, this is really bad in performance as the binary we shipped does not use modern instructions such as AVX/etc which has been around for a long time. For example, SSE4.2 was released in around 2008 (Nehalem).
      I believe we really should enable at least instructions that have been around for some time. This will help the performance greatly.
      [PR #613]

- [ ] While doing training, at the end of the epoch run the following always shows up:
      ```
      2019-11-06 00:31:19.292962: W \
          tensorflow/core/common_runtime/base_collective_executor.cc:216] \
          BaseCollectiveExecutor::StartAbort Out of range: End of sequence \
          [[{{node IteratorGetNext}}]]
      ```
      The issue is likely caused by tf.data pipeline (things like EOF is not returned as expected, or vice versa).  I don't think the above is critical, as I saw that happens in other places.  Though this is not pretty and for cosmetic reasons we should fix the issue. If needed we may need to created PR in core tensorflow repo.

- [x] We could enhance the logging of KafkaDataset to make is more visible when issues happens. [PR #614]

/cc @kaiwaehner @sbaier1

/cc @terrytangyuan @BryanCutler , and @zhjunqin in case you are interested in this area.

yongtang:
CPU instruction compiler flag is covered in PR #613.

yongtang:
Additional lodging is added in #614.

yongtang:
PR #613 and PR #614 have been merged.

yongtang:
PR #628 has been opened to compile avro schema once, and reuse multiple times (avoid recompile for each message).

yongtang:
All items have been resolved. The `BaseCollectiveExecutor::StartAbort Out of range: End of sequence` is from tensorflow which I believe have been fixed in https://github.com/tensorflow/tensorflow/issues/35314

PR #613, opened 2019-11-06
Enable SSE4.2 and AVX (not AVX2) for tensorflow-io

### LINKED ISSUE #612: Enable CPU instructions when building the binary.
This was initially raised in #611, but I think it deserves wide discussion by its own:

In order to support the maximum range of users, we use the most generic CPU instructions when compiling C++ code. This ensures user will not encounter any invalid CPU instructions if their machine is old.

However, this is really bad in performance as the binary we shipped does not use modern instructions such as AVX/SSE4.2 etc which has been around for a long time (SSE4.2 was released in 2008 if I remember correctly).

I believe we really could enable at least instructions that have been around for some time. This will help the performance greatly.

/cc @vlasenkoalexey, I remember you mentioned at one point that there were some performance issues with avro in bigquery. Maybe we could enabling additional CPU instructions to give it a try?

yongtang:
This PR is part of the effort for #612 and #611. Basically we enables SSE4.2 and AVX support by default on Linux.

If machines used does not have the SSE4.2 and AVX support then it will thrown out an error on Linux.

User could still build their own binary with any instruction set.

Note SSE4.2 and AVX are enabled by tensorflow's core library already, so we are basically inline with tensorflow.

Note also that AVX2 are not enabled yet, also inline with tensorflow.

This PR fixes #612.

### DIFF
--- a/.kokorun/io_cpu.sh
+++ b/.kokorun/io_cpu.sh
@@ -57,6 +57,7 @@ bash -x -e tests/test_azure/start_azure.sh
 bash -x -e tests/test_dicom/dicom_samples.sh download
 bash -x -e tests/test_dicom/dicom_samples.sh extract
 
+export BAZEL_CPU_OPTIMIZATION=yes
 bash -x -e .travis/python.release.sh
 
 if [[ $PYTHON_VERSION == "2" ]]; then
--- a/.travis.yml
+++ b/.travis.yml
@@ -66,6 +66,7 @@ jobs:
   - stage: build
     name: "Nightly Release Build on Linux"
     script:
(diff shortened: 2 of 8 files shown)


yongtang:
@BryanCutler Yes by default `BAZEL_CPU_OPTIMIZATION` is not enabled locally when user build with `.travis/python.release.sh`. On Travis CI (and the binary releases) SSE4.2 and AVX are enabled (AVX2 is not enabled).

yongtang:
The failure is only related to the Travis CI cache. PR #576 is also caused by Travis CI cache not populated (thus exceeded the execution timeout) I will just merge in combination with both PRs, and fix the Travis CI cache in subsequent commits.

PR #615, opened 2019-11-07
Add graph mode support for audio IOTensor and IODataset

### LINKED ISSUE #581: tf.data doesn't work with AudioIOTensor
AFAIK `tf.data` always applies `@tf.function` to everything for performance reasons (e.g. distributed compute). This means that `tfio.IOTensor.from_audio` cannot be used in a `tf.data` pipeline as the constructor calls `.numpy()`. Am I misunderstanding how `tfio.IOTensor` is intended to be used or is this a bug?

# MCVE
```py
import tensorflow_io as tfio


def load(path):
    wav = tfio.IOTensor.from_audio(path)
    offset = int(1.0 * wav.rate)
    duration = int(3.0 * wav.rate)
    return wav[start:start + offset].to_tensor()


tf.data.Dataset.from_tensor_slices(["my_audio.wav"]).map(load)
```

```
AttributeError: in converted code:

    <ipython-input-87-7e40b48025cf>:5 load  *
        wav = tfio.IOTensor.from_audio(path)
    /usr/local/lib/python3.6/dist-packages/tensorflow_io/core/python/ops/io_tensor.py:243 from_audio  *
        return audio_io_tensor_ops.AudioIOTensor(filename, internal=True) # pylint: disable=protected-access
    /usr/local/lib/python3.6/dist-packages/tensorflow_io/core/python/ops/audio_io_tensor_ops.py:68 __init__
        shape = tf.TensorShape(shape.numpy())

AttributeError: 'Tensor' object has no attribute 'numpy'
```

yongtang:
This PR tries to address the issue in #581 where IOTensor
from audio (WAV) file does not support graph mode, thus
could not be wired up inside tf.data.Dataset.map() function.

There were several reasons:
1) Initially tf.data was designed as a pipeline in graph mode
   with efficiency as the goal.
2) It was not possible to random access with tf.data as
   tf.data always iterate from beginning to end.
3) It was not possible to get meta data as graph mode essentially
   is a one-shot operation and all of the information about shape
   and dtype has to be defined exactly before hand.

Some of the goals of tensorflow/io were to address 2) and 3).
While addressing 3), IOTensor utilize eager mode to obtain the metadata
in one run, then run the graph.

This reduces user knobs and there is no need to specify Tensor
shape and dtype for any operations.

However, since it utilize eager mode, it is not possible to use IOTensor
inside a tf.data.Dataset's `map` function, because `map` will wrap the
function (or lambda) as a subgraph with autograph.

This PR tries to address the issue by avoid eager mode in IOTensor.

However, by nature of graph, at the minimium the dtype of the Tensor
has to be specified for the output node. There is no real way to
get around (other than infered or casted).

The solution in this PR, is to introduce the graph mode and requiring
user to provide the dtype. With dtype in place, this PR also did some
additional plumbing so that AudioGraphIOTensor and AudioGraphIODataset
could be used in graph mode, thus they could be used in combination
with tf.data.Dataset's map function (and tf.data pipeline).

The loss is the need to provide dtype before hand. But user still have
the option of not providing dtype and use AudioIOTensor and AudioIODataset
in eager mode.

Below is the usage of IOTensor with audio in graph mode (in combination with tf.data.Dataset):
```python
import tensorflow as tf
import tensorflow_io as tfio

filename_dataset = tf.data.Dataset.from_tensor_slices(
    [audio_path, audio_path])
position_dataset = tf.data.Dataset.from_tensor_slices(
    [tf.constant(1000, tf.int64), tf.constant(2000, tf.int64)])

dataset = tf.data.Dataset.zip((filename_dataset, position_dataset))

# Note: @tf.function is actually not needed, as tf.data.Dataset
# will automatically wrap the `func` into a graph anyway.
# The following is purely for explanation purposes.
# Return: audio chunk from position:position+100, and the rate.
@tf.function
def func(filename, position):
  audio = tfio.IOTensor.graph(tf.int16).from_audio(filename)
  return audio[position:position+100], audio.rate

dataset = dataset.map(func)

for (data, rate) in dataset:
  print(data, rate)

```

Below is the usage of IODataset with audio in graph mode (in combination with tf.data.Dataset), essentially a `dataset inside an outer scope dataset`:
```python
import tensorflow as tf
import tensorflow_io as tfio

filename_dataset = tf.data.Dataset.from_tensor_slices(
    [audio_path, audio_path])
position_dataset = tf.data.Dataset.from_tensor_slices(
    [tf.constant(1000, tf.int64), tf.constant(2000, tf.int64)])

dataset = tf.data.Dataset.zip((filename_dataset, position_dataset))

# Note: @tf.function is actually not needed, as tf.data.Dataset
# will automatically wrap the `func` into a graph anyway.
# The following is purely for explanation purposes.
# Return: an embedded dataset (in an outer dataset) for position:position+100
@tf.function
def func(filename, position):
  audio_dataset = tfio.IODataset.graph(tf.int16).from_audio(filename)
  return audio_dataset.skip(position).take(100)

dataset = dataset.map(func)

# Notice audio_dataset in dataset:
for audio_dataset in dataset:
  for value in audio_dataset:
    print(value)
```

This PR fixes #581.

### DIFF
--- a/tensorflow_io/core/kernels/audio_kernels.cc
+++ b/tensorflow_io/core/kernels/audio_kernels.cc
@@ -67,18 +67,15 @@ Status ValidateWAVHeader(struct WAVHeader *header) {
   return Status::OK();
 }
 
-class WAVReadable : public IOReadableInterface {
- public:
-  WAVReadable(Env* env)
-  : env_(env) {}
 
-  ~WAVReadable() {}
-  Status Init(const std::vector<string>& input, const std::vector<string>& metadata, const void* memory_data, const int64 memory_size) override {
-    if (input.size() > 1) {
-      return errors::InvalidArgument("more than 1 filename is not supported");
-    }
-    const string& filename = input[0];
-    file_.reset(new SizedRandomAccessFile(env_, filename, memory_data, memory_size));
+class WAVReadableResource : public ResourceBase {
+ public:
+  WAVReadableResource(Env* env) : env_(env) {}
+  ~WAVReadableResource() {}
+
+  Status Init(const string& input) {
+    const string& filename = input;
+    file_.reset(new SizedRandomAccessFile(env_, filename, nullptr, 0));
     TF_RETURN_IF_ERROR(file_->GetFileSize(&file_size_));
 
     StringPiece result;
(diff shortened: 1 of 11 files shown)


faroit:
@yongtang thats pretty cool. Does this only work with wav files or also with the ffmpeg ops?

yongtang:
@faroit This PR only addresses the WAV issue. The FFmpeg needs additional plumbing. Also FFmpeg does not exactly search based on offset, rather it only offer search based on timestamp. I will see if I can convert FFmpeg to be able to run inside graph mode as well in the next several days.

PR #620, opened 2019-11-09
[Genomics] Add op to calculate Phred Quality Scores

suyashkumar:
This change adds an initial op to convert the [phred quality score](https://en.wikipedia.org/wiki/Phred_quality_score) in a FASTQ file to a probability that a base was called incorrectly.

### DIFF
--- a/tests/test_genome.py
+++ b/tests/test_genome.py
@@ -111,6 +111,7 @@ def test_genome_phred_sequences_to_probability():
     example_quality = tf.constant(example_quality_list)
     converted_phred = genome_io.phred_sequences_to_probability(example_quality)
     out = sess.run(converted_phred)
+
   # Compare flat values
   assert np.allclose(out.flat_values.flatten(), expected_probabilities)
   # Ensure nested array lengths are correct
--- a/tests/test_genome.py
+++ b/tests/test_genome.py
@@ -111,7 +111,7 @@ def test_genome_phred_sequences_to_probability():
 
   with tf.compat.v1.Session() as sess:
     example_quality = tf.constant(example_quality_list)
-    converted_phred = genome_io.phred_sequences_to_probability(example_quality)
+    converted_phred = tfio.genome.phred_sequences_to_probability(example_quality)
     out = sess.run(converted_phred)
 
   # Compare flat values
@@ -121,8 +121,8 @@ def test_genome_phred_sequences_to_probability():
 
 def test_genome_phred_sequences_to_probability_with_other_genome_ops():
   with tf.compat.v1.Session() as sess:
-    raw_data = genome_io.read_fastq(filename=fastq_path)
-    data = genome_io.phred_sequences_to_probability(
+    raw_data = tfio.genome.read_fastq(filename=fastq_path)
+    data = tfio.genome.phred_sequences_to_probability(
       phred_qualities=raw_data.raw_quality)
     out = sess.run(data)
 
--- a/tests/test_genome.py
+++ b/tests/test_genome.py
@@ -102,5 +102,27 @@ def test_genome_sequences_to_onehot():
 
   assert np.all(out.to_list() == expected)
 
+
+def test_genome_phred_sequences_to_probability():
+  example_quality_list = [b'BB<', b'ABFF']
+  expected_probabilities = [0.0005011872854083776, 0.0005011872854083776, 0.0019952619913965464, 0.0006309572490863502, 0.0005011872854083776, 0.00019952621369156986, 0.00019952621369156986]
+
+  with tf.compat.v1.Session() as sess:
+    example_quality = tf.constant(example_quality_list)
+    converted_phred = genome_io.phred_sequences_to_probability(example_quality)
+    out = sess.run(converted_phred)
+  # Compare flat values
+  assert np.allclose(out.flat_values.flatten(), expected_probabilities)
+  # Ensure nested array lengths are correct
+  assert np.all([len(a) == len(b) for a,b in zip(out.to_list(), example_quality_list)])
+
+def test_genome_phred_sequences_to_probability_with_other_genome_ops():
+  with tf.compat.v1.Session() as sess:
+    raw_data = genome_io.read_fastq(filename=fastq_path)
+    data = genome_io.phred_sequences_to_probability(
+      phred_qualities=raw_data.raw_quality)
+    out = sess.run(data)
+
+
 if __name__ == "__main__":
   test.main()
--- a/tests/test_genome.py
+++ b/tests/test_genome.py
@@ -104,27 +104,32 @@ def test_genome_sequences_to_onehot():
 
 
 def test_genome_phred_sequences_to_probability():
+  """Test conversion of phred qualities to probabilities"""
   example_quality_list = [b'BB<', b'ABFF']
   expected_probabilities = [0.0005011872854083776, 0.0005011872854083776,
-      0.0019952619913965464, 0.0006309572490863502, 0.0005011872854083776,
-      0.00019952621369156986, 0.00019952621369156986]
+                            0.0019952619913965464, 0.0006309572490863502,
+                            0.0005011872854083776, 0.00019952621369156986,
+                            0.00019952621369156986]
 
   with tf.compat.v1.Session() as sess:
     example_quality = tf.constant(example_quality_list)
-    converted_phred = tfio.genome.phred_sequences_to_probability(example_quality)
+    converted_phred = tfio.genome.phred_sequences_to_probability(
+        example_quality)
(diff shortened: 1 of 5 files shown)


yongtang:
Thanks @suyashkumar! We are in the process of rework on API exposure. In the past, to get the API the following is needed:
```
import tensorflow_io.genome as genome_io
```

Now we updated the API so that it is possible to do the following:
```
import tensorflow_io as tfio
# use directly
tfio.genome....
```

To add new function in genome, you can add the code:

1) Add the definition of `phred_sequences_to_probability` in tensorflow_io/core/python/ops/genome_ops.py

2) C++ kernel ops is located in:
   tensorflow_io/core/kernels/genome_fastq_kernels.cc
   tensorflow_io/core/ops/genome_ops.cc

3) Add the following line to tensorflow_io/core/python/api/v0/genome.py:
   ```
   from tensorflow_io.core.python.ops.genome_ops import phred_sequences_to_probability # pylint: disable=unused-import
   ```
Hopefully the above is not too much of a burden.

Once those are updated, new API are exposed in two endpoints:
1) tfio.genome.phred_sequences_to_probability
2) tfio.v0.genome.phred_sequences_to_probability

Note the above two endpoints are the same except we have both `tfio` and `tfio.v0`. The idea is to expose `v0` to maintain API compatibility. When we release 1.0 we will expose `tfio.v1`.

yongtang:
@suyashkumar Also, I am not very familiar with FastQ and training/inference. Though looking at some of the tests you write, I think it might be possible to wire up with `tf.data` + `tf.keras`, so that   models could be written in `tf.keras` easily for training and inference purposes.

Given `tf.keras` is the recommended high level API for 2.0, and `tf.data` is the recommended API for pipeline to keras, I think this could really help a lot of the users to adopt TF 2.0.

If you are interested maybe you could write an article or blog, I will see if I can coordinate to help promote or publish the blog/article to tf's web site +1.

yongtang:
@suyashkumar It is a little complicated, as genome uses RaggedTensor and there are some discussions about RaggedTensor with tf.keras  (tensorflow/tensorflow#27170).

For the moment, I think you can just add `phred_sequences_to_probability` in this PR.

We may need additional follow up PRs to wire up genome with tf.keras and tf.data. I will do some more investigation in this area.

suyashkumar:
@yongtang sounds great, I think the new API (with versioning) makes a lot of sense. I'll rebase on your changes and update this PR to that end!

I agree we should try to conform to tf.data and keras api as much as possible. It seems like [RaggedTensors are supported in tf.data](https://github.com/tensorflow/tensorflow/commit/5fe90dc9b4a608bbf2e8a63970a83d87be2648ec) at least. I would be happy to explore this in a future PR. Basically we'd want an op to output a `tf.data.Dataset`representing the information in the fastq file (or set of fastq files)? Would we want to build the Dataset in the C++ op or in python (probably in python)?

To give some quick background, FASTQ files can contain multiple genome sequences of different length (thus the use of RaggedTenosrs). For each letter (nucleotide) in the Genome sequence, there is also an associated quality score that indicates the probability that the nucleotide at that position was read incorrectly (this comes right off the illumina genome sequencing device). This PR's op converts the quality score in the raw fastq format to a float32 probability. Once we get the ops in and stable, I can also write some through docs!

I'd be happy to write a blog post describing how to build a simple model with the tf io genome API as well, if that makes sense. We can discuss that more after we decide where to go with the API? 

Thanks as always @yongtang!

Will make the requested updates shortly, and some of the tf.data and tf.keras api changes can come in future PRs (and we can discuss more on those)!

tests/test_genome.py
```diff
@@ -102,5 +102,30 @@ def test_genome_sequences_to_onehot():
...
+    example_quality = tf.constant(example_quality_list)
+    converted_phred = tfio.genome.phred_sequences_to_probability(example_quality)
+    out = sess.run(converted_phred)
+
+  # Compare flat values
+  assert np.allclose(out.flat_values.flatten(), expected_probabilities)
+  # Ensure nested array lengths are correct
+  assert np.all([len(a) == len(b) for a,b in zip(out.to_list(), example_quality_list)])
+
+def test_genome_phred_sequences_to_probability_with_other_genome_ops():
```
suyashkumar: this test isn't strictly necessary, more of a sanity example test plugging in this op with the base `read_fastq` in the same graph. Can rm.

yongtang:
Thanks @suyashkumar. I think python implementation would make more sense, as it is really easier to maintain in python than in C++. (We could always write a fused C++ version later, if the performance ever becomes an issue.)

I give a quick look at `tf.data` and `tfio.genome`. `tf.data` translate map function into graph before it runs (no matter if `tf.data` runs in graph or eager mode). However, since `tfio.genome` is already written in graph mode, `tfio.genome` more or less already supports `tf.data`.

There is only one change that needs:
```diff
diff --git a/tensorflow_io/core/python/ops/genome_ops.py b/tensorflow_io/core/python/ops/genome_ops.py
index bdfe13e..cbfe07a 100644
--- a/tensorflow_io/core/python/ops/genome_ops.py
+++ b/tensorflow_io/core/python/ops/genome_ops.py
@@ -84,5 +84,5 @@ def sequences_to_onehot(sequences):
         sequence_splits.size(), global_nucleotide_idx)
   return tf.RaggedTensor.from_row_splits(
       values=all_onehot_nucleotides.stack(),
-      row_splits=sequence_splits.stack()
+      row_splits=tf.cast(sequence_splits.stack(), tf.int64)
   )
```

If you apply the above diff, then the following code will work (in eager mode with tf.data).

```
# Note: works in eager mode.
import tensorflow as tf
import tensorflow as tfio

fastq_path = 'test.fastq'

dataset = tf.data.Dataset.from_tensor_slices([fastq_path])
dataset = dataset.map(tfio.genome.read_fastq)
dataset = dataset.map(lambda d: tfio.genome.sequences_to_onehot(sequences=d.sequences))
for d in dataset:
    print("Value: ", d)
```

Would you like to apply the above diff to this PR?

After this PR is merged, I think the next step is to see how far away to pass `tf.data` + `tfio.genome` into `tf.keras` directly. 

Could start with converting RaggedTensor to dense Tensor first just to build a working keras model. Then trying to remove the conversion (of RaggedTensor to dense Tensor).

If there are any issue in RaggedTensor + tf.keras arise, I can help trying to fix it in tensorflow, or finding someone for help.

suyashkumar:
@yongtang sounds good, thanks for the info! Applied the suggested diff. 

Also, it appears that the kokoro builds are failing due to #621. Since I work at Google, I identified the internal CL that caused the breakage in `common.sh` and can send a patch (will need to wait for someone internally to sign off before it can get submitted though).

yongtang:
> Also, it appears that the kokoro builds are failing due to #621. Since I work at Google, I identified the internal CL that caused the breakage in common.sh and can send a patch (will need to wait for someone internally to sign off before it can get submitted though).

Awesome! thanks for the help 👍

suyashkumar:
@yongtang when you get a chance, can you try to rerun the kokoro builds again? Thanks!

yongtang:
@suyashkumar All tests pass now. Thanks for the Kokoro help!

suyashkumar:
@yongtang applied the diff you mentioned and added some eager mode tests. Is there anything else you think should go in this PR?

In future PRs I was thinking I could do the following:
1. Add a `read_fastq_dataset(filenames, convert_quality=false, convert_to_onehot=false)` which calls the underlying read_fastq op for all the files in `filenames` but returns it as a `tf.data.Dataset`. And it can optionally process the fastq data by converting the quality scores to probabilities and converting the nucleotides to onehot encoding if the caller wants.

2. Write markdown documentation for using the various ops

3. Explore if any further changes will be necessary to pass that dataset of ragged tensor into the tf.keras model api. 

If this PR looks good, we can go ahead and merge! Thanks for your comments!

yongtang:
@suyashkumar The PR looks good! 👍  Will merge after Kokoro complete.

> Add a read_fastq_dataset(filenames, convert_quality=false, convert_to_onehot=false) which calls the underlying read_fastq op for all the files in filenames but returns it as a tf.data.Dataset.

I think  a place to expose API might be `tfio.IODataset.from_fastq`. We use `tfio.IODataset` as an entry point for exposing Dataset related APIs.

> Write markdown documentation for using the various ops
> Explore if any further changes will be necessary to pass that dataset of ragged tensor into the tf.keras model api.

That would be fantastic! 👍

suyashkumar:
@yongtang sounds good! I had to add another commit to fix some lint stuff (I need to setup vim to take tfio's pylint config) so we will have to rerun kokoro, but should be good to go after that!

`tfio.IODataset.from_fastq` sounds good, I will look into that next. 

Thanks!

PR #634, opened 2019-11-12
Add tf.keras native io processing

### LINKED ISSUE #633: Provide Keras native I/O though IO-Layer for Kafka
This comes from one discussion about Kafka + TensorFlow IO for predict.

> How can we turn the script into something that continuously streams/predicts from/to kafka?

During `model.predict` we used to rely on KerasCallback to capture the output of the data in a continuous fashion. There are several issues:
1) KerasCallback is in eager mode so it means switching back and forth in graph/eager mode.
2) It is not easy to combine the input data with output data when we need to do some further processing (e.g., calculate the statistics while in inference.).

Thinking again, I think in the past we over-complicates the situation. Keras essentially is a TF graph and every layer is a collection of subgraph. Keras's logic at the input data processing is getting completed because it need to support generator, tensor,  tf.data, numpy arrays, etc. (so many ways to pass the data).

However, in its core Keras is still a TF graph. So for any IO operation all we need is to just add a node in the TF graph, or, similar to tf.keras.layers, add a collection of nodes in the TF graph.

As such I think it is possible to construct an IO operation as a layer of the tf.keras.

Below is a rough diagram:

Kafka+Keras

yongtang:
This PR adds IOLayer as an attempt to have a tf.keras-native io processing.

The idea about IOLayer is that IOLayer is a special layer that
pass through input/output of the keras model, and send a copy (`tee`) to
a storage such as text file or kafka streaming.

As such the following is possible:
```
  model = tf.keras.Sequential([
      tf.keras.layers.Flatten(input_shape=(28, 28)),
      tf.keras.layers.Dense(128, activation=tf.nn.relu),
      tf.keras.layers.Dense(10, activation=tf.nn.softmax)
  ])

  model.compile(optimizer='adam',
                loss='sparse_categorical_crossentropy',
                metrics=['accuracy'])

  model.fit(train_images, train_labels, epochs=5)

  model.summary()

  # For predict, let's create an additional layer to tee to kafka

  io_model = tf.keras.models.Model(
      inputs=model.input,
      outputs=tfio.IOLayer.kafka("topic")(model.layers[-1].output))

  predictions = io_model.predict(test_images)

  io_model.layers[-1]
```

You could even pass IOLayer to the middle of the keras model for training:
```
  f, filename = tempfile.mkstemp()
  os.close(f)

  model = tf.keras.Sequential([
      tf.keras.layers.Flatten(input_shape=(28, 28)),
      tfio.IOLayer.text(filename),
      tf.keras.layers.Dense(128, activation=tf.nn.relu),
      tf.keras.layers.Dense(10, activation=tf.nn.softmax)
  ])

  model.compile(optimizer='adam',
                loss='sparse_categorical_crossentropy',
                metrics=['accuracy'])

  model.fit(train_images, train_labels, epochs=5)

  model.summary()`
```

Though the above may not be that useful in most of the situations as
it will be substentially slower during traing to also write data back to storage.

There are several advantages of using layer for IO:
1) It guarantees that everything runs in graph mode. In the past we tried to use keras'
   callback function to achieve the IO but that runs in eager mode in callback
2) It provides a simple API that ties to users that are already familiar with keras.

This PR fixes #633 


yongtang:
@terrytangyuan @BryanCutler Thanks for the review. The Keras is based on TF primitive ops so for Kafka Layer there is no Keras specific C++ code. In fact the same kind of primitive ops could be used to write a dataset into kafka in a similar way as well, something like:
```
# the same primitive ops in Kafka for Keras could be reused here

dataset = tf.data.Dataset.from....

def write_dataset_to_kafka(dataset):
   resource = io_kafka_init(topic)
   dataset.map(lamba e: io_kafka_write(resource, e)
   io_kafka_sync(resource)
```

Will raise another PR for discussion about dataset write.

Also, the Kafka Layer is written into Kafka in an asynchronous way.

This is mainly for performance reasons: it will be very costly for each message to be confirmed by Kafka server. So this PR by default send the message to kafka and does not expect a confirm. 

There is really no need to explicitly calling `sync()` since **eventually** all messages will be delivered. However, for testing purposes, want to make sure all messages have been confirmed by Kafka server before the test assert and program exit.

I think we could add an additional configurations option to have something like auto_sync=True, so that messages will be sync automatically in background without explicit calling `sync()`

terrytangyuan:
@yongtang Sounds good. That makes sense to me.

PR #644, opened 2019-11-17
Add GitHub workflow

yongtang:


yongtang:
GitHub Workflow could have been a nice way to get around the 50min limit on Travis CI. Though it looks like since a couple of days ago GitHub changed the way it works and it is not working (for many open source orgs like Ansible, etc).

Will close the PR for now, though I do want to revisit the workflow option again when everything settles down.

Issue #647, opened 2019-11-17
improve `from_audio` loading speed


faroit:
I am currently [evaluating](https://github.com/faroit/python_audio_loading_benchmark/pull/8/files#diff-76e989ed1c4ff81f51d6a966464d71af) the performance on tfio for fast wav loading compared to other libs. 

I am using the approach as proposed in #615 by @yongtang. Compared to python function loaders such as scipy.io or soundfile, the native tensorflow ops are significantly slower (even though they do not need to convert from/to numpy first). 

![benchmark_tf](https://user-images.githubusercontent.com/72940/69014652-c7c6fe00-098c-11ea-95a0-8db114cde37f.png)

It would be great if the performance of the tfio audio ops could further be improved

yongtang:
Thanks @faroit. I am currently running the test myself. Will be back soon.

One thing I want to note, is that in PR #613 that was merged a weeks ago, we enabled AVX and SSE4.2, which might have a positive impact on performance.

I also notice in your branch, the `tensorflow-io==0.10.0` was specified in `requirement.txt`. Since  tensorflow-io==0.10.0 has not been released, Not sure which version you are using.

I will just use the latest tensorflow-io-nightly to take a look.

yongtang:
Thanks @faroit for the work. I tried to run the branch `tf2_librosa7`, but seems to get a different result.

I did make several changes in the code, as `ar_mad` does not work inside the container. Also I only enabled `wav`.

Here are the places I made the change:

In benchmark_tf.py, comment out ar_mad, also use `loaders.load_tfio_fromaudio` if lib is `tfio_fromaudio`:
```diff
diff --git a/benchmark_tf.py b/benchmark_tf.py
index 6fd907a..e8d3bfa 100644
--- a/benchmark_tf.py
+++ b/benchmark_tf.py
@@ -50,7 +50,7 @@ if __name__ == "__main__":
     libs = [
         'ar_gstreamer',
         'ar_ffmpeg',
-        'ar_mad',
+        #'ar_mad',
         'aubio',
         'pydub',
         'soundfile',
@@ -72,11 +72,16 @@ if __name__ == "__main__":
                 )
 
                 dataset = tf.data.Dataset.from_tensor_slices(audio_files)
-                if lib in ["tf_decode_wav", "tfio_fromaudio"]:
+                if lib in ["tf_decode_wav"]:
                     dataset = dataset.map(
                         lambda x: loaders.load_tf_decode_wav(x, args.ext),
[...]

PR #649, opened 2019-11-18
Enable SSE 4.2 and AVX on macOS as well, and further optimization

yongtang:
While looking into issue #647 I noticed several areas that could be further optimized.
1) SSE 4.2 and AVX are not enabled on macOS
2) By default bazel seems to use fastbuild mode which equal to -O0.
   Need to further confirm as bazel's behavior is not exactly clear. 
3) The -O2 may also apply to linux as well.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -75,6 +75,7 @@ jobs:
     os: osx
     osx_image: xcode10
     script:
+    - export BAZEL_CPU_OPTIMIZATION=yes
     - bash -x -e .travis/python.release.sh --nightly ${TRAVIS_BUILD_NUMBER}
     - bash -x -e .travis/wheel.test.sh
     after_success: bash -x -e .travis/after-success.sh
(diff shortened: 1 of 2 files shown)


yongtang:
Continue to face issues on Travis CI execution limit. Otherwise the PR passed all tests on Kokoro CI.

Apparently the `-O3` flag makes the build time super long.

Given tensorflow 2.1 rc0 likely will be released within a week. My plan is to only enable SSE4.2 and AVX in this PR. and cut a release soon.

Had a chat in Contributor Summit with several guys, we might be able to get some help in either having a bazel remote cache, or have another CI than Travis CI. Though that may take some time before it is ready.

We will see what we could do to resolve the Travis CI issue later (or maybe we could think about GitHub workflow which gives us 6 hours for each execution).

yongtang:
All tests pass except the Travis CI cache is causing one test to timeout. This PR is really low risk as SSE4.2 and AVX has been enabled on Linux anyway.

I am planning to merge first, then have one or two follow up commits to get Travis CI cache addressed.

PR #655, opened 2019-11-19
updated for tf2.0

omartin2010:
This didn't work for me with 1.14 and with 2.0. Otherwise it with this class.


omartin2010:
@googlebot I signed it!

PR #660, opened 2019-12-04
Rename tfio.IOLayer to tfio.experimental.IOLayer

yongtang:
As IOLayer was only recetnly introduced, it makes sense
to first place it under `tfio.experimental`.


yongtang:
Once this PR is in I think we are good with 0.10.0 release.

PR #664, opened 2019-12-05
Add `alwayslink = 1` for Bazel rules to support bazel 1.0+

yongtang:
For some reason, Bazel 1.0+ changed the way cc_library is generated,
as a result, with bazel 1.0+, building tensorflow-io fails.

This PR adds `alwayslink = 1` so that code base of tensorflow-io
still works with bazel 1.0+.

Note on Travis CI we still use 0.29.0 for now (as TF 2.1 release is comming).

We will upgrade Travis CI's Bazel version later.

### DIFF
--- a/tensorflow_io/arrow/BUILD
+++ b/tensorflow_io/arrow/BUILD
@@ -25,4 +25,5 @@ cc_library(
         "//tensorflow_io/core:dataset_ops",
         "@arrow",
     ],
+    alwayslink = 1,
 )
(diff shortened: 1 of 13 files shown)


yongtang:
This is a pure mechanical change and all tests pass, so I will just merge. The current configuration in Travis CI and Kokoro CI are not impacted at all (we still use 0.29.0 in Travis CI and Kokoro CI).

It allows user with Bazel 1.0+ to build tensorflow-io locally (not possible before).

PR #662, opened 2019-12-04
Add Ogg support for AudioIOTensor

yongtang:
This PR adds Ogg support for `IOTensor.from_audio`. No API updates as it automatically reroute to Ogg if file type support.

This PR fixes #650.

Signed-off-by: Yong Tang <yong.tang.github@outlook.com>

Issue #651, opened 2019-11-18
Standardize tests for dataset


yongtang:
We are adding more and more formats to dataset, though the tests are pretty much ad hoc at the moment. Since tf.data's pipeline is pretty standard, we could standardize the tests for almost all Dataset in tensorflow-io by parametrization and fixture. We also need benchmark (with pytest-benchmark?).

/cc @BryanCutler as quite a few columnar dataset are from Arrow, I think it might be easier to standardize tests for those.

The tests for audio and streaming dataset could also be standardized as well.

/cc @terrytangyuan

BryanCutler:
Sounds like a really good idea to me!

yongtang:
Our tests for the majority of the dataset has been standardized in test_io_dataset_eager.py, with different scenarios testing increased functionality. Will continue to enhance the tests, but will close this issue as tests are there.

PR #665, opened 2019-12-06
Add initial benchmark tests, and add initial standarized IODataset tests

yongtang:
This PR is an effort to:
1) Add benchmark tests
2) Add standarized IODataset tests
so that all IODataset (especially graph mode GraphIODataset)
behaves the same across the board.

At the moment only from_audio has been moved to this standarized
tests with the help of fixtures and parametrized tests in pytest.

This PR also found out that from_lmdb can not repeat the iteration,
in other words:
```
dataset = tfio.IODataset.from_lmdb(...)
entries = [e for e in dataset]
entries = [e for e in dataset] <= failure!
```

This PR has not fixed the lmdb yet, will have a follow up PR
to fix lmdb.


yongtang:
@terrytangyuan @BryanCutler This PR is the one we discussed before in #651. I am thinking using this PR as a starting point to standardize `IODataset` so that all dataset carries the same feature (if applicable).

@faroit I added a benchmark tests in this PR. Will use this as a start point to optimize the Ogg performance (as was mentioned in https://github.com/tensorflow/io/pull/662#issuecomment-562730933)

PR #666, opened 2019-12-07
Add FLAC audio support

yongtang:
This PR adds FLAC audio support so that from_audio could
process FLAC file (in addition to WAV and OGG)

This PR is part of the effort in #647.


yongtang:
This PR is ready for review. It add basic FLAC audio support.

/cc @faroit as it might be relevant.

yongtang:
@faroit  The latest nightly build is 2280:
https://pypi.org/project/tensorflow-io-nightly/0.10.0.dev2280/

which should have the FLAC audio support. Please take a look if interested, and let me know if you encounter any issues.

faroit:
benchmark_tf

works fine. Still a bit slower than sndfile/librosa (also uses sndfile for flac). Maybe adding a sndfile interface would be a good idea since this covers ogg and flac?

But then, I am really more curious into how much better an ffmpeg interface performs compared to stdio parsing (such as done in `ar_ffmpeg`)

yongtang:
Thanks @faroit. Both FLAC and OGG are based on Xiph's reference implementation library (unlike  WAV which is our native implementation). I will take a look at Xiph's code base to see if there is any room for speed up, or if alternative library could be used for FLAC and OGG.

yongtang:
@faroit With maximum compiler flags:
```
-march=native -O3 -funroll-loops -ffast-math -finline-functions
```

I am able to get very good numbers. So I think it is merely a matter of fine tuning compiler flags.

Notice in the above `-march=native` has been passed. However, we could not use `-mach=native` directly as otherwise it may not run on certain machines. For compatibility reasons we may only enable certain flags.

Let me spend more time and get back to you and see what we could do to improve the speed.

yongtang:
@faroit With latest nightly 0.11.0.dev2320:
https://pypi.org/project/tensorflow-io-nightly/0.11.0.dev2320/

The `-O2` has been enabled. I could see pretty good performance gain on my local machine. Please take a look if interested.

faroit:
@yongtang can confirm: nice speedup. That is probably sufficient given that FLAC is not that widespread. 

These are the results using `num_parallel_calls=1`

benchmark_tf

Interestingly, using more cores (`num_parallel_calls=4`) does not significantly improve the decoding performance for tf-io but seems to scale well for `soundfile`. Is this because of the compiler flags you've set?

benchmark_tf

yongtang:
@faroit The reason num_parallel_calls does not make a difference is that the FLAC dataset was lock protected. It should not be very difficult to parallelize it though. It could even be done at the python level (not underlying C++ level). let me take a look and see how to speed up with num_parallel_calls>1.

PR #679, opened 2019-12-13
Change Peek-then-Read to Read in one op (with lock)

yongtang:
In the past we use Peek-then-Read which is really two ops
and that may potenitally make the op non-atomic. This PR
changes to one step Read (lock protected).


yongtang:
Small enhancement with all tests pass. Will just merge.

PR #685, opened 2019-12-15
Move Kinesis to tfio.experimental

yongtang:
And move tests to tests/test_io_dataset_eager.py


yongtang:
It is not as impactful as other wider used ops such as Kafka and BigQuery. Also many user are in transition from 1.x to 2.0 so change should be fine for this one.

We do need to take a look and see how to make adjustment for Kafka and BigQuery, but I think we can defer the discussion after 0.11.0 (TF 2.1).

PR #695, opened 2019-12-17
Add decode_json support

### LINKED ISSUE #692: Support decode_json to decode json into Tensor
This is a feature request (from gitter channel) to add `decode_json` so that it is possible to convert json strings such as below:
```
{
    "Foo": 208.82240295410156,
    "Bar": "93",
    "Background": "~/0109.jpg",
    "Focal Length": 36.9439697265625,
    "Location": [ 7.8685874938964844, -4.7373886108398438, -0.038147926330566406 ],
    "Rotation": [ -4.592592716217041, -4.4698805809020996, -6.9197754859924316 ]
}
```

into Tensors.

The method `decode_json` could be efficiently combined with `tf.data` pipeline so that the following is possible:
```
dataset = tf.data.Dataset.from_tensor_slices(["file_0.json", "file_1.json", "file_2.json"])
dataset = dataset.map(tf.io.read_file)
dataset = dataset.map(lambda e: decode_json(e))
...
# feed dataset to tf.keras directly:
model.fit(dataset)
```

/cc @noahtren

yongtang:
This PR adds decode_json support so that it is possible to use
```
values = tfio.experimental.serialization.decode_json(json, specs)
```
to decode a JSON string into a structured Tensors.

The specs field is the structured tf.TensorSpec that is necessary
to reconstruct the JSON values. The reason being JSON could be intepreted
differently in lists:
```
{
"v": [1, 2, 3]
}
```
could be considered as one Tensor of shape `(3)`, or a 3 Tensor structure within dict "v".

Note structured tf.TensorSpecs need to change list `[]` into tuple `()`
as list is not supported in structures in TensorFlow.

This PR fixes #692.

### DIFF
--- a/tensorflow_io/core/BUILD
+++ b/tensorflow_io/core/BUILD
@@ -369,6 +369,21 @@ cc_library(
 )
 
 cc_library(
+    name = "serialization_ops",
+    srcs = [
+        "kernels/serialization_kernels.cc",
+        "ops/serialization_ops.cc",
+    ],
+    copts = tf_io_copts(),
+    linkstatic = True,
+    deps = [
+        "//tensorflow_io/core:dataset_ops",
+        "@rapidjson",
+    ],
+    alwayslink = 1,
+)
+
+cc_library(
     name = "grpc_ops",
     srcs = [
         "kernels/grpc_kernels.cc",
@@ -424,6 +439,7 @@ cc_binary(
         "//tensorflow_io/core:parquet_ops",
         "//tensorflow_io/core:pcap_ops",
         "//tensorflow_io/core:pubsub_ops",
+        "//tensorflow_io/core:serialization_ops",
         "//tensorflow_io/core:text_ops",
         "//tensorflow_io/core/azure:azfs_ops",
         "//tensorflow_io/core/oss:oss_ops",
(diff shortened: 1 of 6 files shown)


noahtren:
Thanks so much! This is a huge help to my pipeline which includes JSON alongside a lot of other data types. So helpful to have a function to turn these into tensors.

djl11:
The following example does not parse the json correctly:

`
core_ops.io_decode_json(tf.constant('{"x":[[1.0]]}'), ['/x'], [tf.float32])
`

Gives:

`
[<tf.Tensor: shape=(1,), dtype=float32, numpy=array([4.2949673e+09], dtype=float32)>]
`

However, 1-D works, for example:

`
core_ops.io_decode_json(tf.constant('{"x":[1.0, 2.0]}'), ['/x'], [tf.float32])
`

Gives:

`
[<tf.Tensor: shape=(2,), dtype=float32, numpy=array([1., 2.], dtype=float32)>]
`


Is this desired behaviour?
I am not sure how to load a multi-dimensional tensor stored in a json file as a nested list.

yongtang:
@djl11 A PR #1101 has been opened to support the 2-D tensor in your example.

djl11:
I think with recursion the multi-dimensional case shouldn't be too hard, unless I'm missing something. If I find the time I'll try and implement this myself. Thanks for putting in the new PR @yongtang!

PR #715, opened 2020-01-04
remove parallel_interleave from bigquery_api and reroute to `tf.data.Dataset.interleave`

pshiko:
Happy New Year to all the maintainers! And always thank you for the development! 🎉 🎉 

---

`parallel_interlave` is deprecated, so we should use `tf.data.Dataset.interleave`!

The change is very easy, but,  there are some problems.
Currently, we can pass `sloppy` flag to parallel_read_rows.
This is used to specify wether the order of the elements is deterministic or not-deterministic.
However, `tf.data.Options.experimental_deterministic` also has same role, and is global!

I think we should use `tf.data.Options.experimental_deterministic` as default, and when if we want to specify deterministic or not-deterministic, then we should change global configurations of `tf.data.Options.experimental_deterministic` or use `tf.data.Dataset.with_options`.

For the above reasons, it is recommended to deprecated `floppy` or set the default value to None which means using `tf.data.Options.experimental_deterministic`.

In this PR, I focused on maintaining backward compatibility and only added a warning that deprecated argument floppy.

### DIFF
--- a/tensorflow_io/bigquery/python/ops/bigquery_api.py
+++ b/tensorflow_io/bigquery/python/ops/bigquery_api.py
@@ -33,8 +33,8 @@ from operator import itemgetter
 from tensorflow.python.data.ops import dataset_ops
 from tensorflow.python.framework import dtypes
 from tensorflow.python.framework import tensor_spec
-from tensorflow_io.core.python.ops import core_ops
 from tensorflow.python.util import deprecation
+from tensorflow_io.core.python.ops import core_ops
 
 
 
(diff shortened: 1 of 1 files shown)


tensorflow_io/bigquery/python/ops/bigquery_api.py
```diff
@@ -179,7 +179,13 @@ def read_rows(self, stream):
     return _BigQueryDataset(self._client_resource, self._selected_fields,
                             self._output_types, self._avro_schema, stream)
 
-  def parallel_read_rows(self, cycle_length=None, sloppy=False, block_length=1):
+  @deprecation.deprecated_args(
```
vlasenkoalexey: why do you need this?
As far as I understand you added support for sloppy using tf.data.Options.experimental_deterministic below

vlasenkoalexey: Just saw a comment on PR, I think it makes sense

PR #723, opened 2020-01-10
Update Release note for v0.11.0

yongtang:
Update Release note for v0.11.0

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,6 +89,7 @@ version of TensorFlow I/O according to the table below:
 
 | TensorFlow I/O Version | TensorFlow Compatibility | Release Date |
 | --- | --- | --- |
+| 0.11.0 | 2.1.x | Jan 10, 2019 |
 | 0.10.0 | 2.0.x | Dec 5, 2019 |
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -89,6 +89,7 @@ version of TensorFlow I/O according to the table below:
 
 | TensorFlow I/O Version | TensorFlow Compatibility | Release Date |
 | --- | --- | --- |
+| 0.11.0 | 2.1.x | Jan 10, 2019 |
```
terrytangyuan: Is 2.0.x still supported though? Backwards compatibility for major TF versions should be ensured IMO.

yongtang: @terrytangyuan The 2.0.x support may or may not work, due to the nature that we have to compile against C++ API in 2.1.x TensorFlow Core.

It might work (if there is no major C++ header file change) but that was never tested.

In the long run, once core tensorflow releases C API (ongoing) then we could guarantee backward compatibility.

terrytangyuan: Yea it would be better to guarantee that in the future as it might not be as easy to upgrade TensorFlow for certain companies and then they won't be able to use these new features.

PR #728, opened 2020-01-10
Update Kokoro to only use Python 3

yongtang:


yongtang:
This PR is ready now.

Issue #501, opened 2019-09-24
Windows support for tensorflow-io


hagianga21:
Hi, I got this error:
ERROR: Could not find a version that satisfies the requirement tensorflow-io (from versions: none)
ERROR: No matching distribution found for tensorflow-io

How could I fix it?

yongtang:
@hagianga21 Can you provide more details about your tensorflow version, python pip version, and python version + platform (e.g. Ubuntu 16.04)?

hagianga21:
Hi, I use Window, tensorflow gpu 2.0 (anaconda env), python 3.6

yongtang:
@hagianga21 We don't have windows build yet. It might be possible to enable Windows build as long as we could have a reliable CI to generate the binary. 

I have changed the title to "windows support for tensorflow-io". We might be able to have the Windows build soon.

yongtang:
With GitHub Actions supported (See #645), Windows support will be easier as GitHub actions give Windows through Azure pipelines (my understanding). We could start looking into adding windows support really soon.

yongtang:
PR #741 will enable nightly build for Windows, it only includes lmdb for now though, as porting everything to Windows will take some time.

yongtang:
The nightly build of tensorflow-io for Windows platform is now available in:
https://pypi.org/project/tensorflow-io-nightly/0.11.0.dev20200117174350/#files

At the moment only LMDB is supported, though we plan on porting most of the ops into Windows gradually over time. Hopefully before 0.12.0 we should have most of the ops available in Windows.

yongtang:
With quite a few recent PRs, the Windows support has been enabled in more than half of the all available ops. 

There are some remaining ops that has not been supported yet:
https://github.com/tensorflow/io/blob/c31de505c9ff67a8d5646e8bd6dbef4be23ee232/tensorflow_io/core/BUILD#L514-L551

Some file systems ops can only be enabled after TF 2.2 due to some issues in upstream TensorFlow. (See https://github.com/tensorflow/tensorflow/pull/35947).

yongtang:
With recent changes, most of the ops for Windows build are enabled, and nightly is releasing Windows at the same time as Linux and macOS.

What are not ported to Windows yet are below, mostly due to a header file issue in TensorFlow. We will have to wait for TF 2.2:
- azfs (requires at least TF 2.2 to get a file system header file fix)
- ignite (same as azfs)
- oss (same as azfs)
[...]

PR #734, opened 2020-01-13
[Working-in-Progress] Add Windows support for tensorflow-io

yongtang:
This is an initial effort to add Windows support for tensorflow-io (#501).

Any help is welcomed.


yongtang:
Made some progress but it is still far away from getting complete. At the moment Windows build times out in Bazel analysis. Will need to look further.

PR #735, opened 2020-01-14
Small fixes

yongtang:
This PR adds a couple of small fixes. One is that in lint,
`find` command has an extra `-f` (`find <path>` instead of `find -f <path>`).
Another fix is that grpc_python_dependencies is not actually invoked,
also removed unneeded llvm_register_toolchains invoke.


yongtang:
This is a minor change to address a few obvious errors. Will just merge.

PR #736, opened 2020-01-14
[Working-in-Progress] Add Windows support for tensorflow-io

yongtang:
This is an initial effort to add Windows support for tensorflow-io (#501).

Any help is welcomed.


yongtang:
Get a stub Bazel working. Looks promising.

yongtang:
Looks like we get a basic windows wheel working, the wheel could be installed and `tests/test_text_eager.py::test_re2_extract` passed. Most of the modules are not build yet but it is matter of time now.

One glitch is that I have to delete tensorflow_io directory as otherwise the imported module will always check for the source directory (not the wheel install). (maybe some cache issue?)

Overall, think we could release windows wheel pretty soon.

yongtang:
Close this for now. Additional support for Windows will be gradually pulled in.

PR #739, opened 2020-01-15
Enable Windows Build for Azure File System

yongtang:
This is the first step to have Windows release. As there are lots of packages in tensorflow-io, the plan is to enable AZFS first (which fits the scenario as well in Windows users).

Once AZFS is enabled, we could start releasing windows nightly and gradually add more components into the build (after fixing all issues).


yongtang:
TensorFlow's FileSystem may have any issue and file_system.h may have to be patched (TranslateName => TranslateNameA has to be undefined).

Looks like Azure File system is not a good candidate for first windows port. Will update after upstream patched).

PR #741, opened 2020-01-16
Enable LMDB build for Windows

yongtang:
Enable LMDB build for Windows,

This is part of the effort for #501


yongtang:
All issue seems to have been fixed and we should be ready for Windows release. Though it looks like GitHub changed the base image of windows-latest, and that is causing CI failures? Will take a look.

yongtang:
This PR is ready now. With this PR, we get python wheels on windows for python 3.5, 3.6, 3.7.

Due to the complexity of the windows build, this PR only makes LMDB works. However, this could be a first step to get tensorflow-io working in windows.

Once this PR is merged, nightly for windows will shown up in PyPI.org.

/cc @terrytangyuan @BryanCutler @vlasenkoalexey

Issue #680, opened 2019-12-13
decode video causes memory leak in ram.


hollowgalaxy:
Code to replicate the issue.

```
import tensorflow as tf
from tensorflow_io.core.python.experimental.ffmpeg_ops import decode_video
def decode_video_profiling(files=[]):
    assert len(files) > 0, 'need files to profile'
    @tf.function
    def file_to_video(video_path):
        content = tf.io.read_file(video_path)
        frames = decode_video(content, 0)
        frames = tf.image.resize(frames, [224, 224])
        return frames[:20]
    video_paths = tf.data.Dataset.from_tensor_slices(files).repeat()
    videos = video_paths.map(file_to_video).batch(16, drop_remainder=True)
    memory_profiling(videos)

def memory_profiling(ds, steps=600, stop_and_repeat=False):
    from tqdm import tqdm
    import psutil
    pbar = tqdm(range(steps))
    percents = [psutil.virtual_memory().percent]
    pbar.set_description_str(f'{percents[-1]}')
    pbar.update()
    while pbar.n < pbar.total:
        for i, dps in enumerate(ds):
            if isinstance(dps, tuple):
                print(f'{i}:', ','.join([dp.shape for dp in dps]))
            else:
                print(f'{i}: {dps.shape}')
            percents += [psutil.virtual_memory().percent]
            pbar.set_description_str(f'Memory usage: {percents[-1]}%')
            pbar.update()
            if stop_and_repeat or pbar.n >= pbar.total:
                break

files = [...] # a bunch of mp4 valid mp4 files which are about 24 fps and 10 seconds long
decode_video_profiling(files)
```
The code above will print the metrics overtime *Note: This is CPU RAM* . However I put into a chart,
![Figure_2_all](https://user-images.githubusercontent.com/52365584/70837124-5fb2de80-1db7-11ea-8427-33245cb5de53.png)

hollowgalaxy:
with the code it never reaches the full 600 steps. Throws an OOM error.

hollowgalaxy:
I tried using the suggestion from https://github.com/tensorflow/io/pull/622#issue-339178915
however there is also a memory leak there as well. Code used:
```
import tensorflow_io as tfio
@tf.function
def func(filename):
  return tfio.IODataset.graph(tf.uint8).from_ffmpeg(filename, "v:0").batch(20).take(1)

def decode_video_profiling(files=[]):
    assert len(files) > 0, 'need files to profile'
    @tf.function
[...]

PR #743, opened 2020-01-17
fix memory leak in decode_video

kyamagu:
Current `decode_video` op leaks memory due to not calling `av_packet_unref()` after every `av_read_frame()`. This PR fixes this issue.

#680


yongtang:
Thanks!

PR #744, opened 2020-01-17
Fix Windows Nightly build where build number env was not populated

yongtang:
The build number was not populated so it did not reach to PyPI.org,
This PR fixes it.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -456,7 +456,7 @@ jobs:
           @echo on
           python --version
           python -m pip install -U wheel setuptools
-          python setup.py --data bazel-bin -q bdist_wheel --nightly $BUILD_NUMBER
+          python setup.py --data bazel-bin -q bdist_wheel --nightly %BUILD_NUMBER%
           ls -la dist
       - uses: actions/upload-artifact@v1
         with:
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yml
```diff
@@ -456,7 +456,7 @@ jobs:
           @echo on
           python --version
           python -m pip install -U wheel setuptools
-          python setup.py --data bazel-bin -q bdist_wheel --nightly $BUILD_NUMBER
+          python setup.py --data bazel-bin -q bdist_wheel --nightly %BUILD_NUMBER%
```
terrytangyuan: I am not familiar with this. What kind of syntax is this?

yongtang: @terrytangyuan `%ENV%` is the Windows CMD way of specifying environmental variable (since we are building on windows platform).

terrytangyuan: Got it. Thanks!

yongtang: @terrytangyuan It is quite old age... here is the documentation:
https://support.microsoft.com/en-us/help/121170/how-to-access-environment-variables-in-an-ms-dos-batch-file

PR #751, opened 2020-01-21
Revert "Update librdkafka to 1.3.0 (#750)"

yongtang:
Looks like upgrading librdkafka to 1.3.0 is causing test failures. This PR reverts back.

This reverts commit a641ca46c9afc81292b5df2f5e3a6e2f5dd5f720.


terrytangyuan:
Why aren't we catching this in CI?

yongtang:
@terrytangyuan At the time Kokoro tests passed (https://source.cloud.google.com/results/invocations/daa2aec0-6595-4c97-9509-78b1d8f73150/targets)

So the GitHub Actions failure was (incorrectly) considered as flaky. Later it was found that Kokoro only covers Ubuntu 18.04 + Python 3.7, while GitHub Actions' failure was on Ubuntu 16.04 + Python 3.5. 

I will update the Kokoro to cover Ubuntu 16.04 as well.

(Also from the above I tends to think the issue is related to Ubuntu 16.04 only.)

yongtang:
All tests pass now 👍 after the revert.

Issue #745, opened 2020-01-18
H5::DataSetIException when using HDF5 IOTensor in graph mode


zaccharieramzi:
After #710 was fixed by #732 , I tried it on a single file (which I sent to @yongtang ), and it worked perfectly.

However, when trying it on the whole dataset, I got the following error (I don't know which file is causing this and I don't know how I could know except iterating through all of them):

```
HDF5-DIAG: Error detected in HDF5 (1.10.5) thread 0:
  #000: external/hdf5/src/H5Dio.c line 185 in H5Dread(): could not get a validated dataspace from file_space_id
    major: Invalid arguments to routine
    minor: Bad value
  #001: external/hdf5/src/H5S.c line 254 in H5S_get_validated_dataspace(): selection + offset not within extent
    major: Dataspace
    minor: Out of range
terminate called after throwing an instance of 'H5::DataSetIException'
```

Apparently this error appeared [here](https://github.com/steven-varga/h5cpp/issues/4) in `h5cpp` (which is what is used in tf-io from what I understand) but was fixed. There is not much info though so I don't really know what happened, maybe this is a new case, in which case I probably need to write the issue over there.

If you have no idea what's happening, I can try and identify the one file causing this issue.

yongtang:
@zaccharieramzi I don't have a good idea about the issue, as the related issue in `h5cpp` is not very clear. Though I think I can add some enhancement to the error message of the HDF5 Dataset (PR #748), so that it will shown up the file that triggers the error. 

Once PR #748 is merged, maybe you could give it a try with the nightly, and see which file is causing the issue?

yongtang:
The latest nightly should have the error message enhancement now.

zaccharieramzi:
Hi @yongtang !

Thanks for this update, it's very useful. I actually have a lot of failing files, you can find one here: https://we.tl/t-qJL5uVdB7o.

The failing code is the following (on CPU):
```python
def image_and_kspace_from_h5(fpath):
    spec = {
        '/kspace': tf.TensorSpec(shape=[None, None, None], dtype=tf.complex64),
        '/reconstruction_esc': tf.TensorSpec(shape=[None, None, None], dtype=tf.float32),
    }
    h5_tensors = tfio.IOTensor.from_hdf5(fpath, spec=spec)
    image = h5_tensors('/reconstruction_esc').to_tensor()
    kspace = h5_tensors('/kspace').to_tensor()
    return image, kspace

[...]

PR #765, opened 2020-01-31
Add ArrowIOTensor with input from a pyarrow.Table

BryanCutler:
This adds an `ArrowIOTensor` based on a `_TableIOTensor` with multiple columns. An `ArrowIOTensor` is create with the factory function `tfio.IOTensor.from_arrow()` with a `pyarrow.Table` as input.

### DIFF
--- a/tests/test_io_tensor_eager.py
+++ b/tests/test_io_tensor_eager.py
@@ -278,6 +278,25 @@ def fixture_hdf5_scalar(request):
 
   return args, func, expected
 
+
+@pytest.fixture(name="arrow", scope="module")
+def fixture_arrow():
+  """fixture_arrow"""
+  import pyarrow as pa # pylint: disable=import-outside-toplevel
+
+  arr1 = pa.array(list(range(100)), pa.int32())
+  arr2 = pa.array(list(range(100)), pa.int64())
+  arr3 = pa.array([x * 1.1 for x in range(100)], pa.float32())
+  table = pa.Table.from_arrays([arr1, arr2, arr3], ['a', 'b', 'c'])
+
+  args = table
+  column = 'b'
+  func = lambda t: tfio.IOTensor.from_arrow(t)(column)
+  expected = table[column].to_pylist()
+
+  return args, func, expected
+
+
 # scalar is a special IOTensor that is alias to Tensor
 @pytest.mark.parametrize(
     ("io_tensor_fixture"),
@@ -309,6 +328,7 @@ def test_io_tensor_scalar(fixture_lookup, io_tensor_fixture):
         pytest.param("audio_flac"),
         pytest.param("hdf5"),
         pytest.param("kafka"),
+        pytest.param("arrow"),
     ],
     ids=[
         "audio[wav]",
@@ -317,6 +337,7 @@ def test_io_tensor_scalar(fixture_lookup, io_tensor_fixture):
         "audio[flac]",
         "hdf5",
         "kafka",
+        "arrow",
     ],
 )
 def test_io_tensor_slice(fixture_lookup, io_tensor_fixture):
@@ -351,6 +372,18 @@ def test_io_tensor_slice(fixture_lookup, io_tensor_fixture):
         pytest.param("hdf5_graph", 2),
         pytest.param("kafka", None),
         pytest.param("kafka", 2),
+        pytest.param(
+            "arrow", None,
+            marks=[
+                pytest.mark.skip(reason="TODO"),
+            ],
+        ),
+        pytest.param(
+            "arrow", 2,
+            marks=[
+                pytest.mark.skip(reason="TODO"),
+            ],
+        ),
     ],
     ids=[
         "audio[wav]",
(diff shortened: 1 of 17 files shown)


BryanCutler:
@yongtang this is mostly working, but still a few issues to iron out and I need to make a pass to fix code style. I labelled it a WIP for now, and I should be able to finish soon.

BryanCutler:
@yongtang I think this is ready for review. There are still a few issues, but they could be added in followups. Thanks!

tools/dev/Dockerfile
```diff
@@ -34,7 +34,7 @@ ARG CONDA_ADD_PACKAGES=""
 
 RUN conda create -y -q -n tfio-dev python=3.6 ${CONDA_ADD_PACKAGES}
 
-ARG ARROW_VERSION=0.14.1
+ARG ARROW_VERSION=0.15.1
```
BryanCutler: TODO change to 0.16.0 once available

yongtang:
@BryanCutler Thanks for the PR! Overall it looks good.

I would suggest to add a several simple tests in `test_io_tensor_eager.py` as well (in addition to the already added tests in `test_arrow_eager.py`. `test_arrow_eager.py` is a much more comprehensive tests suites, though `test_arrow_eager.py` is more like a `feature comparision` for each different op types.

For example, in the past hdf5 was not possible in graph mode so we skipped the test with:
```
        pytest.param(
             "hdf5", None,
             marks=[		
                 pytest.mark.skip(reason="TODO"),		
             ],		
         ),
```

until PR #730 


(Note there are still some issues as reported by #745, though at least we have some coverage and knows which "feature" is available).

For Arrow, I think we could add a simple test in `test_io_tensor_eager.py` with one or two column only, just to show if some features are supported or not. If not supported we could just use `pytest.marks` to skip and add TODO.

The added tests in `test_arrow_eager.py` are used to serve a more through coverage.

yongtang:
@BryanCutler Also recently I saw quite some user request to ask for support in graph mode (for both dataset and iotensor). The need for graph mode support is mostly to combine with tf.data itself to form a full pipeline. In other words, user would prefer to use:
```
dataset = tf.data.Dataset.from_tensor_slices([filename1, filename2, filename3])
dataset = dataset.map(lambda f: tfio.IOTensor.from_...())
```

This will constitutes a complete data pipeline that could be deployed in production in a infinitely continuous way (after converting from_tensor_slices to a steam source of filename entries).

I think the above is a useful scenario that will help in production usage.

While IOTensor starts with eager mode in focus, it is not very far away to support graph mode. The only requirement for graph mode, is the `dtype` which has to be known at the graph construction time (`REGISTER_OP`).

We could ask user to provide the dtype (or tf.TensorSpec) information as an additional arg when graph mode is used. HDF5IOTensor is an example where dtype/spec is provided in graph mode.

Also, inside `REGISTER_OP`, it will help shape inference if the shape information is provided explicitly. However, shape is not absolutely needed. In most cases, a rank would be suffice in graph construction time. The shape information could be passed as a tensor in graph execution time later on.

Based on the above, it is possible to only ask user for `dtype` in graph mode. And convert `shape` in `REGISTER_OP("IO>ArrowReadableRead")` from attribute to tensor input. In this way, shape is not needed by user input. The `SetShapeFn([](shape_inference::InferenceContext* c)` could use Unknown shape.

yongtang:
@BryanCutler The graph mode support could be added later in follow up PRs I think. Though it would be good to add a TODO, or add a stub test in io_tensor_eager_test.py, and mark the test as TODO, e.g.,
```
        pytest.param(
             "arrow", None,
             marks=[		
                 pytest.mark.skip(reason="TODO"),		
             ],		
         ),
```

yongtang:
@BryanCutler Let me know if there are any questions or I can be of any help.

tests/test_io_tensor_eager.py
```diff
@@ -278,6 +278,24 @@ def fin():
...
+def fixture_arrow(request):
+  import pyarrow as pa # pylint: disable=import-outside-toplevel
+
+  arr1 = pa.array(list(range(100)), pa.int32())
+  arr2 = pa.array(list(range(100)), pa.int64())
+  arr3 = pa.array([x * 1.1 for x in range(100)], pa.float32())
+  table = pa.Table.from_arrays([arr1, arr2, arr3], ['a', 'b', 'c'])
+
+  args = table
+  column = 'b'
```
BryanCutler: I couldn't figure out how to parametrize the fixture, so this is only testing 1 column. That's probably good enough for now though

BryanCutler:
@yongtang I added tests in `test_io_tensor_eager` and plan to add graph support as a followup. Please take another look when you can, thanks!

tests/test_io_tensor_eager.py
yongtang: @BryanCutler There are some ways to parameterize the fixture:
https://stackoverflow.com/questions/42228895/how-to-parametrize-a-pytest-fixture/56871701#56871701

though they are quite complex. Probably not worth the effort for now as the current is good enough.

BryanCutler:
Thanks @yongtang !  I'll be working on followups and a tutorial too.

PR #767, opened 2020-02-03
Update Arrow library to enable all compression support, and reuse exisitng libs

yongtang:
This PR updates arrow library to enable all compression support, and reuse exisitng libs (to avoid multiple copies of the same code like lz4/xxhash/etc).

In the past xxhash was compiled in multiple places including Kafka and Arrow.

Note the latest arrow is 0.16.0 while we are still using 0.14.1. We will hold on the update of 0.16.0 as PR #765 is in progress.

We will try to update to 0.16.0 once PR #765 is in.


BryanCutler:
> We will try to update to 0.16.0 once PR #765 is in.

Sounds good, I was thinking the same thing :smile:

PR #768, opened 2020-02-05
Update Arrow to 0.16.0, and enables Windows Build for Arrow

yongtang:
This PR updates Arrow to 0.16.0, and enables Windows Build for Arrow Dataset. The test is not enabled yet, but the build on Windows has been successful.


BryanCutler:
Hey @yongtang , I forgot that 0.16.0 hasn't officially been released yet and the RC candidate is still being voted on. I'm kind of surprised the tarball is available and I kind of wonder if it's the latest RC# or what? I guess it's not a big deal since it will be released very soon..

yongtang:
@BryanCutler I didn't find rc label, it was downloaded from GitHub:

https://github.com/apache/arrow/releases/tag/apache-arrow-0.16.0

I downloaded the tag.gz file (source).

BryanCutler:
Yeah, I'm saying I don't think it should have been tagged yet since it's not officially released :smile: . But it's ok I gues, just FYI in case the checksum changes..

yongtang:
@BryanCutler Ah got it. We can always update once the final 0.16.0 release is out. I would assume the change will be fairly minor.

PR #775, opened 2020-02-09
Add tfio.genome tutorial

suyashkumar:
This adds an initial tutorial of the `tfio.genome` functionality. [Here's a link to the colab](https://colab.research.google.com/drive/1YQRJ1_v1lxJT7IVvwnRWJUoHAHgwjJ_N). 

It's pretty barebones right now, as there's a fair bit more that needs to be added to the genome package (including a `tf.data` API). Let me know if you think we should proceed as this is for now, or if we would want to wait to build out some of the genome API farther first--either way seems reasonable to me!

### DIFF
--- a/docs/tutorials/genome.ipynb
+++ b/docs/tutorials/genome.ipynb
@@ -1,0 +1,238 @@
+{
+  "nbformat": 4,
+  "nbformat_minor": 0,
+  "metadata": {
+    "colab": {
+      "name": "Tensorflow IO Genomics Tutorial",
+      "provenance": [],
+      "private_outputs": true,
+      "collapsed_sections": [
+        "Tce3stUlHN0L"
+      ],
+      "toc_visible": true,
+      "include_colab_link": true
+    },
+    "kernelspec": {
+      "display_name": "Python 3",
+      "language": "python",
+      "name": "python3"
+    },
+    "language_info": {
+      "codemirror_mode": {
+        "name": "ipython",
+        "version": 3
+      },
+      "file_extension": ".py",
+      "mimetype": "text/x-python",
+      "name": "python",
+      "nbconvert_exporter": "python",
+      "pygments_lexer": "ipython3",
+      "version": "3.6.3"
+    }
+  },
+  "cells": [
+    {
+      "cell_type": "markdown",
+      "metadata": {
+        "id": "view-in-github",
+        "colab_type": "text"
+      },
+      "source": [
+        "<a href=\"https://colab.research.google.com/github/suyashkumar/io/blob/s%2Fadd-genome-tutorial/docs/tutorials/genome.ipynb\" target=\"_parent\"><img src=\"https://colab.research.google.com/assets/colab-badge.svg\" alt=\"Open In Colab\"/></a>"
+      ]
+    },
+    {
+      "cell_type": "markdown",
+      "metadata": {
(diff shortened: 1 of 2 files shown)


docs/tutorials/genome.ipynb
```diff
@@ -0,0 +1,267 @@
...
+    {
+      "cell_type": "markdown",
+      "metadata": {
+        "colab_type": "text",
+        "id": "xHxb-dlhMIzW"
+      },
+      "source": [
+        "## Overview\n",
+        "\n",
+        "This tutorial demonstrates the `tfio.genome` package functionality (which is currently __a work in progress__--contributions are welcome). The goal of this package is to provide commonly used genomics IO functionality--namely reading several genomics file formats and also providing some common operations for preparing the data (for example--one hot encoding or parsing Phred quality into probabilities). \n",
```
lamberta:
```suggestion
        "This tutorial demonstrates the `tfio.genome` package that provides commonly used genomics IO functionality--namely reading several genomics file formats and also providing some common operations for preparing the data (for example--one hot encoding or parsing Phred quality into probabilities). \n",
```

I don't think it's necessary to mention it's a work in progress since what isn't ;)
But if you want to mention it's experimental, you can do that with a call out box:

Note: The `tfio.genome` module is experimental.

docs/tutorials/genome.ipynb
```diff
@@ -0,0 +1,267 @@
...
+        "\n",
+        "We use the [Google Nucleus](https://github.com/google/nucleus) library to provide some of the core functionality. "
```
lamberta: Avoid "we" since it's not clear who you are speaking for. Just say what uses it.

docs/tutorials/genome.ipynb
```diff
@@ -0,0 +1,267 @@
...
+      "cell_type": "markdown",
+      "metadata": {
+        "id": "qxHjVKXzdx5W",
+        "colab_type": "text"
+      },
+      "source": [
+        "As you see, the returned `fastq_data` has `fastq_data.sequences` which is a string tensor of all sequences in the fastq file (which can each be a different size) along with `fastq_data.raw_quality` which includes Phred encoded quality information about the quality of each base read in the sequence.\n",
+        "\n",
+        "### Quality\n",
+        "We can use a helper op to convert this quality information into probabilities if we are interested."
```
lamberta: Google dev docs style prefers second person "you" instead of "we": https://developers.google.com/style/person

docs/tutorials/genome.ipynb
suyashkumar: Makes sense, thank you!

docs/tutorials/genome.ipynb
suyashkumar: Makes sense--done!

docs/tutorials/genome.ipynb
suyashkumar: Makes sense--done!

suyashkumar:
@lamberta Thanks for the comments! I went ahead and addressed them--let me know what you think!

suyashkumar:
Also thanks @MarkDaoust for the review and update!

MarkDaoust:
published to tensorflow.org: https://www.tensorflow.org/io/tutorials/genome

Issue #783, opened 2020-02-12
Publish tfio 0.12 to tensorflow.org


MarkDaoust:
Hi, I'm trying to execute the azure notebook for tensorflow.org

I get this error:

```
InternalError                             Traceback (most recent call last)
<ipython-input-5-0fe6a29d3aa3> in <module>
      1 pathname = 'az://{}/aztest'.format(account_name)
----> 2 tf.io.gfile.mkdir(pathname)
      3 
      4 filename = pathname + '/hello.txt'
      5 with tf.io.gfile.GFile(filename, mode='w') as w:

/tmpfs/src/tf_docs_env/lib/python3.5/site-packages/tensorflow_core/python/lib/io/file_io.py in create_dir_v2(path)
    423     errors.OpError: If the operation fails.
    424   """
--> 425   pywrap_tensorflow.CreateDir(compat.as_bytes(path))
    426 
    427 

InternalError: Failed to create directory az://devstoreaccount1/aztest (unknown_error - -598609368)
```
Any ideas?

yongtang:
Thanks @MarkDaoust. I am running the azure notebook from colab.research.google.com:

https://colab.research.google.com/github/tensorflow/io/blob/master/docs/tutorials/azure.ipynb

and it seems to be fine. Are you running the notebook locally?

Wondering if there is any difference between the environment you run and the colab.research.google.com?

yongtang:
One possible scenario is that I am pushing the button one by one on Google Colab. However, the line:
```
tf.io.gfile.mkdir(pathname)
```

will need the previous action `get_ipython().system_raw('azurite-blob -s &')` fully up. 

So if the notebook runs locally, the `azurite-blob -s &` may not be fully up yet.

Wondering if it makes a difference if you add a sleep (e..g, 5 sec) before `tf.io.gfile.mkdir(pathname)`?

MarkDaoust:
Thanks for the tips. It looks like the publisher got silently tripped up on the `npm install` step.

yongtang:
@MarkDaoust With tensorflow-io 0.12.0 released (see #813), I have updated docs in several places (PR #772 #819 #820), mostly to replace tensorflow-io-nightly with tensorflow-io. Can you take a look and see if those changes make sense. Maybe we could just push to tensorflow.org/io once the related PRs ( #772 #819 #820) are merged?

MarkDaoust:
Sure. 

I've got the api-docs for 0.12 done. 

[...]

PR #785, opened 2020-02-13
Maybe fix npm install.

MarkDaoust:
For #783

The install was failing silently.

### DIFF
--- a/docs/tutorials/azure.ipynb
+++ b/docs/tutorials/azure.ipynb
@@ -98,6 +98,16 @@
     {
       "cell_type": "markdown",
       "metadata": {
+        "id": "MsyhTesh4LSy",
+        "colab_type": "text"
+      },
+      "source": [
+        "Caution: In addition to python packages this notebook uses `npm install --user` to install packages. Be careful when running locally. \n"
+      ]
+    },
+    {
+      "cell_type": "markdown",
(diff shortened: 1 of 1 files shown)


MarkDaoust:
Wait, let me clear the output before merging.

MarkDaoust:
Done.

yongtang:
@MarkDaoust When I try the notebook on Google Colab, it give the following error:
```
!npm --user install -g azurite@2.7.0

Error: setuid user id does not exist
    at /tools/node/lib/node_modules/npm/node_modules/uid-number/uid-number.js:49:16
    at ChildProcess.exithandler (child_process.js:267:7)
    at emitTwo (events.js:126:13)
```

On the other hand, if I try with the following (remove `-g` from the original, also removes `--user`):
```
!npm install azurite@2.7.0
```
Then the install is successful. 

Don't know about your local setup, though I guess the error you encountered locally before (with `!npm install -g azurite@2.7.0`) might be due to the usage of `-g`?

Would you mind give `!npm install azurite@2.7.0` a try locally? That might clear both the local execution error and Colab error?

MarkDaoust:
> !npm install azurite@2.7.0

Works locally, and in colab. Worth a shot. let's merge and see if that fixes the publisher.

PR #796, opened 2020-02-19
Enable pcap build on Windows, and remove unneeded headers

yongtang:
Enable pcap build on Windows, and remove unneeded headers


yongtang:
@terrytangyuan Yes with some recent changes this file is not needed anymore.

PR #801, opened 2020-02-21
Add MP3 support (with minimp3) for Audio Dataset and IOTensor

yongtang:
This PR adds MP3 support (with minimp3) for Audio Dataset and IOTensor.
The difference between this PR and FFmpeg is that FFmpeg is only supported
on specific Ubuntu 16.04 and 18.04, though external .so.
This PR builds minimp3 as part of the program and support Windows/Linux/macOS.

/cc @faroit in case you might be interested.


lieff:
The error of skipping the not decodable first frames in callback mode was fixed here https://github.com/lieff/minimp3/commit/2b9a0237547ca5f6f98e28a850237cc68f560f7a . It's rare case, but may be worth updating.

yongtang:
Thanks @lieff for the update! I have created a PR in #807 to bring minimp3 to 2b9a023.

By the way, thanks for your great work for minimp3  👍 !

lieff:
Thanks. Glad to help :)

faroit:
@yongtang sorry, didn't had much time recently. I update my benchmarks to include 0.12.0 and tested the minimp3 performance. 

Its almost factor 10 faster than ffmpeg! 🐎  

benchmark_tf

yongtang:
@faroit Nice 👍 !

PR #805, opened 2020-02-24
Add MP4A support for macOS with minimp4 and AVFoundation/Swift

yongtang:
This PR add initial mp4a support for macOS with minimp4 for container parsing and macOS's native AVFoundation in swift for decoding.

The swift code for calling AVFoundation is exported as __cdecl so that it could be used by C++/C.

This PR is only a "starter" that setup a basic pipeline with only mp4 LC tested. It also only works on macOS. The plan is to add Windows native API, and add FFMpeg into Linux in the future.

**Update**

Summary of the PR:
1. On macOS, mp4a is supported through minimp4 + Swift/AVFoundation.
2. On Linux, mp4a is supported through minimp4 + FFmpeg/libavcodec
3. Unlike old ffmpeg ops, this PR does not use libavformat (which is quite hard to integrate)
4. Only FFmpeg on Ubuntu 18.04 is supported. It is hard to write one code base that suite Ubuntu 16.04's FFmpeg. We could add Ubuntu 20.02 support (higher version) once it is released.
5. Performance is not exactly great, as we are doing packet by packet processing. Additional work could be done to batch-process a big number of packets in follow up PRs.
6. This PR also requires Xcode install on macOS. This is an additional requirement, however, it is not exactly too big of a burden as normally macOS developer always install Xcode.

Advantages:
1. We use system native APIs on macOS. Avoids FFmpeg as it is really not a solution on macOS (too hard to guarantee API compatibility).
2. We setup the Swift on macOS in place, could create interesting applications in the future.


yongtang:
This PR is now ready.

PR #816, opened 2020-02-28
Add graph mode support for ArrowIOTensor

BryanCutler:
This enables graph mode support for `ArrowIOTensor`, e.g. columns from a `pyarrow.Table` can be converted to Tensors inside of a `tf.data.Dataset.map` function.

### DIFF
--- a/tensorflow_io/arrow/kernels/arrow_kernels.cc
+++ b/tensorflow_io/arrow/kernels/arrow_kernels.cc
@@ -241,13 +241,7 @@ class ArrowReadableFromMemoryInitOp : public ResourceOpKernel<ArrowReadableResou
 class ArrowReadableSpecOp : public OpKernel {
  public:
   explicit ArrowReadableSpecOp(OpKernelConstruction* context)
-      : OpKernel(context), column_index_(-1) {
-    int32 column_index;
-    Status status = context->GetAttr("column_index", &column_index);
-    if (status.ok()) {
-      column_index_ = column_index;
-    }
-  }
+      : OpKernel(context) {}
 
   void Compute(OpKernelContext* context) override {
     ArrowReadableResource* resource;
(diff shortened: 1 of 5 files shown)


BryanCutler:
@yongtang please take a look, thanks!

yongtang:
@BryanCutler There is also a floating point exception on Linux. Can you take a look?

Note when we build the wheel on GitHub, we pass the bazel optimization flag `--copt=-msse4.2 --copt=-mavx --compilation_mode=opt` to maximize the performance. This might be different compared to default flag where there is no optimization. Can this be the difference?

yongtang:
Since we plan on releasing 0.13 as soon as TF 2.2 is released, and TF 2.2. release is imminent (in several weeks). We could release 0.12 first, and this PR could be picked up on 0.13.

BryanCutler:
Thanks @terrytangyuan and @yongtang , its totally fine to get this in the next release. There are still a few issues to work out too.

BryanCutler:
Strange, I couldn't reproduce any floating point errors, even with the additional build opts. Let's rerun the tests.

yongtang:
Seems to be fine? Let me give it another try. If it is still ok then maybe it is not truly reproducible.

yongtang:
Actually looks like I don't have the access to retry the GitHub actions (might be only tied to the author). @BryanCutler would you mind to invoke the run one more time? Just want to make sure the previous crash was not related.

BryanCutler:
>@BryanCutler would you mind to invoke the run one more time? Just want to make sure the previous crash was not related.

Sure, let's run it a couple more times. I don't see the button to restart either, so let me just rebase.

BryanCutler:
Last tests passed also, let's try again

BryanCutler:
@yongtang this passed the last 2 tests, so perhaps that error  was a fluke. Does it look ok to merge and then I can investigate further if the error happens again?

yongtang:
@BryanCutler Looks good to me 👍 . We can come back if the issue resurface.

BryanCutler:
Thanks @yongtang and @terrytangyuan !

PR #827, opened 2020-03-03
Update Azure doc tutorial to prefix npm bin path

yongtang:
This PR tries to address the issue raised that Azure notebook
tutorial does not work on Colab any more. It looks like the
bin path of the npm packages is not exposed in PATH environmental variable
any more.

This PR add the following to get the npm_bin_path, and prefix
the azurite-blob package to have the full path available:
```
npm_bin_path = get_ipython().getoutput('npm bin')[0]
```

Note `npm bin` will output the package bin path, see
https://docs.npmjs.com/cli/bin.html

For: #783

### DIFF
--- a/docs/tutorials/azure.ipynb
+++ b/docs/tutorials/azure.ipynb
@@ -3,7 +3,7 @@
   "nbformat_minor": 0,
   "metadata": {
     "colab": {
-      "name": "notebook.ipynb",
+      "name": "azure.ipynb",
       "provenance": [],
       "private_outputs": true,
       "toc_visible": true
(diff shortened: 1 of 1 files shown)


yongtang:
Thanks @MarkDaoust, I upload the notebook to Google Colab, and saved the output. The PR has been updated. 

I do notice that after I run through Colab, the following line has been removed (please see commit)
```
"private_outputs": false,
```

I am not very familiar with Notebook, wondering if this is intended behavior?

MarkDaoust:
> the following line has been removed (please see commit) `"private_outputs": false`

Yes this is a reasonable . False is the default setting, so i guess when Colab re-saved the notebook it removed the field entirely.

yongtang:
@MarkDaoust Thanks for the help 👍 !

PR #837, opened 2020-03-05
Allows check/lint individual files with `bazel run //tools/lint:check -- clang -- file.cc`

yongtang:
This PR update out lint tool so that it is possible to run:
`bazel run //tools/lint:check -- clang -- file.cc`

to check specific file.cc, and run:

`bazel run //tools/lint:lint -- clang -- file.cc`

to reformat specific file.cc.

In addition, this PR also adds support for black so that

`bazel run //tools/lint:lint -- black -- tensorflow_io/python/core/ops/version_ops.py` (for lint)

`bazel run //tools/lint:check -- black -- tensorflow_io/python/core/ops/version_ops.py` (for check)

could be used similiarly.


yongtang:
/cc @terrytangyuan I only used black for one file `tensorflow_io/python/core/ops/version_ops.py` and the CI does not invoke black yet.

Once your PR #812 is complete we can enable black for CI.

Also, you can run 
```
bazel run //tools/lint:check -- black 
```

to check all, and run
```
bazel run //tools/lint:lint -- black 
```

to format all. (That will trigger lots of errors through so we may need to do it in piecemeal.)

BryanCutler:
Great! This will help a lot :)

PR #838, opened 2020-03-05
Update to libkafka v1.4.0-RC2, and patch to fix the segment fault

### LINKED ISSUE #800: Update librdkafka to 1.0.0+
This is an issue to update librdkafka to 1.0.0+. The biggest motivation for update is that our current version of librdkafka does not work correctly with OS X 10.13, due to https://github.com/Blizzard/node-rdkafka/issues/686#issue-499098109

Since we are publishing OS X 10.13 compatible wheels now, it is imperative to update librdkafka so that tensorflow-io on 10.13 works correctly.

Currently, there is another issue with librdkafka that is causing segmentation fault if we update, it is being tracked in https://github.com/edenhill/librdkafka/issues/2716

yongtang:
This PR update to libkafka v1.4.0-RC2, and patch to fix the segment fault.

The patch is pretty ugly, though.
```
sed -i.bak '\|rd_kafka_log(rk,|,/ exceeded);/ s/^/\/\//' src/rdkafka_cgrp.c
```

We will try to remove the patch if the issue could be resolved upstream.

This PR fixes #800 

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -207,6 +207,7 @@ http_archive(
     build_file = "//third_party:kafka.BUILD",
     patch_cmds = [
         "rm -f src/win32_config.h",
+        # TODO: Remove the fowllowing once librdkafka issue is resolved.
         """sed -i.bak '\|rd_kafka_log(rk,|,/ exceeded);/ s/^/\/\//' src/rdkafka_cgrp.c""",
     ],
     sha256 = "2d14551fd87262ec4917db3ae688ca2574d716faecccb1ac5136a478579cee19",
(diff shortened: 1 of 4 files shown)


WORKSPACE
```diff
@@ -207,11 +207,12 @@ http_archive(
     build_file = "//third_party:kafka.BUILD",
     patch_cmds = [
         "rm -f src/win32_config.h",
+        """sed -i.bak '\|rd_kafka_log(rk,|,/ exceeded);/ s/^/\/\//' src/rdkafka_cgrp.c""",
```
terrytangyuan: Can we add a TODO note somewhere?

yongtang: @terrytangyuan added a TODO.

Issue #824, opened 2020-03-03
Remove python2 specific code


yongtang:
As tensorflow-io dropped the support for python 2, all python 2 specific code could be removed. For example, the following code that are common across almost all py files:

```
from __future__ import absolute_import
from __future__ import division
from __future__ import print_function
```

Mark this issue as `contribution welcome` and `good first issue`.

gabrieldemarmiesse:
Hint from TF Addons: We used Pyugrade which was designed just for this :)

yongtang:
Thanks @gabrieldemarmiesse! I added a PR #843 so that it is possible to run pyupgrade through bazel without installation. The command line below will sanitize file.py:
```
bazel run //tools/lint:lint -- pyupgrade -- file.py
```

PR #843, opened 2020-03-06
Add lint option for pyupgrade

yongtang:
Now it is possible to run lint auto format with:
```
bazel run //tools/lint:lint -- pyupgrade -- file.py
```

and run lint check with:
```
bazel run //tools/lint:check -- pyupgrade -- file.py
```

Or full check for every python file
```
bazel run //tools/lint:check -- pyupgrade
```

This PR is part of #824

### DIFF
--- a/README.md
+++ b/README.md
@@ -112,7 +112,7 @@ version of TensorFlow I/O according to the table below:
 
 ### Lint
 
-TensorFlow I/O's code conforms through Pylint, Bazel Buildifier, and Clang Format. The following will check the source code and report any lint issues:
+TensorFlow I/O's code conforms through Pylint, Bazel Buildifier, Clang Format, Black, and Pyupgrade. The following will check the source code and report any lint issues:
 ```sh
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -112,7 +112,7 @@ version of TensorFlow I/O according to the table below:
 
 ### Lint
 
-TensorFlow I/O's code conforms through Pylint, Bazel Buildifier, and Clang Format. The following will check the source code and report any lint issues:
+TensorFlow I/O's code conforms through Pylint, Bazel Buildifier, Clang Format, Black, and Pyupgrade. The following will check the source code and report any lint issues:
```
terrytangyuan: Are these set of linters compatible though?

yongtang: @terrytangyuan I would assume it will be ok. Our goal for pyupgrade is to get the majority of  the code into python3 and remove python2. Once most of the files are cleaned we really don't need 100% for pyupgrade. We could remove the pyupgrade then.

Issue #839, opened 2020-03-05
Audio Processing API and tfio.audio


yongtang:
This is about API design of the audio processing in tfio. Comments and discussions are welcomed.

When we started to get into reading audio for tfio, we start with a feature request from the community of 24bit WAV file   which was not possible in tf core repo. Since then, we also supported `Dataset` types of access that essentially is a sequential read, and random access (AudioIOTensor) which allows python `__getitem__()` and `__len__()` styles access in tensorflow graph.

Later, we added additional audio format support such as `Flac`, `Ogg`, `MP3` and `MP4`, in addition to the already supported `WAV` file. Note `WAV`, `Flac`, `Ogg`, `MP3` does not requires external .so dependency. For `MP4`, we use AVFoundation on macOS, but needs FFmpeg's .so files on Linux. We also added the `resample` as a basic ops (not a python class) recently upon request from community. 

Recently, new requests to add supports for decoding memory input audio into tensors comes in play (See PR #815). So the APIs for audio processing in tfio will expand even further.

Initially we have not truly come up with a good audio api design as those features are added over gradually, span across more than a year. Now as the features are gradually in shape, I think it is point to start lay out tfio.audio to revisit APIs so that:
1. Capture the usage from community for different use cases.
2. Allow expansion in the future (our focus was on decoding but encoding will be added soon).

The audio processing in tf repo, is pretty much limited and is really not a good template for us to follow. I also look into pytorch's audio package as well. While pytorch's APIs are typically very clean, the audio related apis are not as clean as some other parts. pytorch also relies on sox which could be an issue (sometimes could be a benefit depending on the scenario.).

We can summarize the use case that has been covered:
1. AudioIODataset which is a subclass of tf.data.Dataset and could be passed to tf.keras.
    Dataset is a sequential access (not random accessed) that allows iteration from the beginning to the end.
   The shape and dtype are probed automatically in eager mode. dtype has to be provided by user in graph mode.
   It is normally file based (or callback-based) and will not load the whole clip into memory (lazy-loaded).
2. AudioIOTensor which exposes `__getitem__`, `__len__`, `shape`, `dtype`, and `rate` API in python and could be used in graph mode (certain restriction apply).
    IOTensor is a random access that allows reading data at any index (`__getitem__`). This is very useful in situations where user want to access just a small clip of the audio. The `shape` API allows user to get the number of channels and samples. `shape = [samples, channels]`.
[...]

PR #844, opened 2020-03-08
Add tfio.experimental.audio.decode_wav support

yongtang:
This is part of the discussion in #839 by adding
tfio.experimental.audio.decode_wav support. It relies
on existing implementation with additional shape inference.

Test cases for both eager mode and graph mode has been added (
note graph mode is invoked inside a tf.data pipeline).

/cc @jjedele 

### DIFF
--- a/tensorflow_io/core/kernels/audio_flac_kernels.cc
+++ b/tensorflow_io/core/kernels/audio_flac_kernels.cc
@@ -171,10 +171,11 @@ class FlacReadableResource : public AudioReadableResourceBase {
         }) {}
   ~FlacReadableResource() {}
 
-  Status Init(const string& input) override {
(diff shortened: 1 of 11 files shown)


yongtang:
@jjedele If you don't mind I plan to merge this PR soon. This PR does made a small C++ change in that now it is possible to pass optional memory buffer into the AudioResource, but otherwise the rest are pretty much the same.

On the C++ level if you are thinking about changing code it is always possible, just need to make sure to apply the same change to all other audio ops to make them consistent.

jjedele:
@yongtang Sorry for not reacting to this, I'm a bit under water currently. Looks like a good solution. I will try to implement the decode_mp3 op the same way.

PR #847, opened 2020-03-10
Add `tfio.experimental.audio.info(content)` support

yongtang:
This PR adds `tfio.experimental.audio.info(content)` support
which takes an input of audio buffer and outputs
`shape`, `dtype`, `rate`, and `encoding` of the audio.

This is useful in situations where the meta information of the input need to be checked to figure out the next step.

At the moment only wav is supported. However, other formats
will be added once the equivalent decode_format is in place.


terrytangyuan:
@yongtang There are conflicts now

yongtang:
Convert to draft. Need to spend more time to think how to consolidate the presentation discrepancy in between graph vs eager mode.

PR #849, opened 2020-03-10
ArrowIOTensor should hold data references until tensors are evaluated

BryanCutler:
Data buffers in a `pyarrow.Table` can be freed once all references go out of scope. For ArrowIOTensor, the `ArrowBaseIOTensor` computes the component tensor and should hold a reference, also for graph mode the result is not evaluated eagerly, so the resulting `Tensor` needs a reference for when it is evaluated.

### DIFF
--- a/tensorflow_io/core/python/ops/arrow_io_tensor_ops.py
+++ b/tensorflow_io/core/python/ops/arrow_io_tensor_ops.py
@@ -135,11 +135,18 @@ class _ArrowIOTensorComponentFunction():
 
 class ArrowBaseIOTensor(io_tensor_ops.BaseIOTensor):
   """ArrowBaseIOTensor"""
-  def __init__(self, shape, dtype, spec, function, internal=False):
+  def __init__(self,
+               shape,
+               dtype,
+               spec,
+               function,
+               arrow_data_refs,
+               internal=False):
     super().__init__(spec, function, internal=internal)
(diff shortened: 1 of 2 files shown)


BryanCutler:
@yongtang is there still some work to do with the lint script? It says pyupgrade requires python2. Also should I reformat the modified files with black now?

yongtang:
@BryanCutler Can you run the following command
bazel run //tools/lint:lint -- pyupgrade -- tensorflow_io/core/python/ops/arrow_io_tensor_ops.py

This will automatically convert the file to python3 - compatible one.

I tried locally, I think the issue is the following:
```diff
@@ -165,7 +165,7 @@ class ArrowBaseIOTensor(io_tensor_ops.BaseIOTensor):
 
   def __getitem__(self, key):
     """Returns the specified piece of this IOTensor."""
-    result = super(ArrowBaseIOTensor, self).__getitem__(key)
+    result = super().__getitem__(key)
     if not tf.executing_eagerly():
       # Insert refs into tf.Tensor result so data is valid until evaluated
       result._arrow_data_refs = self._arrow_data_refs
```

yongtang:
@BryanCutler For black we can skip for now.

BryanCutler:
Ah, I missed that one. Looks like that was it, thanks @yongtang !

PR #850, opened 2020-03-12
Define ArrowIOResource to initialize Arrow Table in a graph

BryanCutler:
It could be useful to be able to initialize a `pyarrow.Table` inside a `tf.data.Dataset.map` function, e.g. reading a list of files and loading each one at a time into a table.

This creates a `ArrowIOResource` that allows for a function returning a pyarrow.Table to be used to initialize an `ArrowIOTensor` inside a map operation. The function becomes wrapped in a tf.py_function, where it will be eagerly executed to initialize the table that can then be input into the graph.

### DIFF
--- a/tensorflow_io/arrow/kernels/arrow_kernels.cc
+++ b/tensorflow_io/arrow/kernels/arrow_kernels.cc
@@ -33,6 +33,7 @@ namespace {
 class ArrowReadableResourceBase : public ResourceBase {
  public:
   virtual Status Init(const std::shared_ptr<arrow::Table>& table) = 0;
+  virtual int32 GetColumnIndex(const string& column_name) = 0;
   virtual Status Spec(int32 column_index,
                       PartialTensorShape* shape,
                       DataType* dtype) = 0;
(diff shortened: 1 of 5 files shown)


BryanCutler:
cc @yongtang and @feihugis since we discussed this

BryanCutler:
Test failure is a little strange, I'll have to look into it a bit more..

yongtang:
@BryanCutler I think this test is flaky. I noticed in one test it failed (https://github.com/tensorflow/io/runs/506664447?check_suite_focus=true) but another is successful (https://github.com/tensorflow/io/runs/506705373?check_suite_focus=true)

It always happens on python 3.5 though.

yongtang:
@BryanCutler I added a PR #852 to temporarily disable the failing test.

BryanCutler:
Thanks @yongtang , I can reproduce this locally but it is flaky and only seems to be with Python 3.5. I opened #858 to investigate further.

BryanCutler:
The flaky tests for python 3.5 were because I was assuming a dict was in order, but forgot that for 3.5 the order isn't preserved, duh!

BryanCutler:
Looks like tests finally passed, going to merge

PR #856, opened 2020-03-15
Add tfio.experimental.audio.[decode|encode]_flac support

yongtang:
This PR adds Add tfio.experimental.audio.[decode|encode]_flac support, as part of the effort to revisit audio APIs.

This PR is part of #839

### DIFF
--- a/tensorflow_io/core/kernels/audio_flac_kernels.cc
+++ b/tensorflow_io/core/kernels/audio_flac_kernels.cc
@@ -16,6 +16,7 @@ limitations under the License.
 #include "tensorflow_io/core/kernels/audio_kernels.h"
 
 #include "FLAC/stream_decoder.h"
+#include "FLAC/stream_encoder.h"
 
 namespace tensorflow {
 namespace data {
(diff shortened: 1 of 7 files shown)


tensorflow_io/core/kernels/audio_flac_kernels.cc
```diff
@@ -262,6 +310,180 @@ class FlacReadableResource : public AudioReadableResourceBase {
...
+    encoder.reset(FLAC__stream_encoder_new());
+
+    FLAC__bool ok;
+
+    ok = FLAC__stream_encoder_set_verify(encoder.get(), true);
+    OP_REQUIRES(context, ok, errors::InvalidArgument("unable to set verify"));
+
+    // ok = FLAC__stream_encoder_set_compression_level(encoder.get(), 5);
+    // OP_REQUIRES(context, ok, errors::InvalidArgument("unable to set
+    // compression level"));
```
terrytangyuan: Are these commented code needed?

yongtang: @terrytangyuan I am thinking the compression level could be an input value passed from user. Not sure if it makes sense yet, so comment out.

I have added a TODO comment in the code.

Issue #861, opened 2020-03-17
Adjust docstring of python code to align top and bottom


yongtang:
After we use` black` to format python code, it looks like the doctsring of python is not aligned as `black` indent the top one but does not indent the whole string. As an example, below is the docstring after `black`:
```
def resample(
    input, rate_in, rate_out, quality, name=None
):  # pylint: disable=redefined-builtin
    """Resample audio.
  Args:
    input: A 2-D `Tensor` of type `int16` or `float`. Audio input.
    rate_in: The rate of the audio input.
    rate_out: The rate of the audio output.
    quality: The quality of the resample, 1-10.
    name: A name for the operation (optional).
  Returns:
    output: Resampled audio.
  """
```

But instead we should have the following:
```
def resample(
    input, rate_in, rate_out, quality, name=None
):  # pylint: disable=redefined-builtin
    """Resample audio.
    Args:
      input: A 2-D `Tensor` of type `int16` or `float`. Audio input.
      rate_in: The rate of the audio input.
      rate_out: The rate of the audio output.
      quality: The quality of the resample, 1-10.
      name: A name for the operation (optional).
    Returns:
      output: Resampled audio.
    """
```

Otherwise docs team may have trouble generating the website. (and the non-aligned one does not look good aesthetically).

This is an issue we need fix before v0.13.0.

kvignesh1420:
@yongtang I am currently adjusting the docstrings for the modules in `tensorflow_io/core/python/ops`. Will raise a PR for the same.

PR #864, opened 2020-03-19
Add test for Arrow datasets with variable length arrays and no batching

BryanCutler:
Arrow datasets can currently support variable length arrays if not batched. This adds a test to verify correctness.

### DIFF
--- a/tests/test_arrow_eager.py
+++ b/tests/test_arrow_eager.py
@@ -706,12 +706,11 @@ class ArrowDatasetTest(ArrowTestBase):
         self.run_test_case(dataset, truth_data, batch_size=2)
 
     def test_bool_array_type(self):
+        """NOTE: need to test this seperately because to_pandas fails with
+        ArrowNotImplementedError: Not implemented type for list in
+        DataFrameBlock: bool
+        see https://issues.apache.org/jira/browse/ARROW-4370
         """
-    NOTE: need to test this seperately because to_pandas fails with
-    ArrowNotImplementedError:
-      Not implemented type for list in DataFrameBlock: bool
-    see https://issues.apache.org/jira/browse/ARROW-4370
-    """
         truth_data = TruthData(
             [[[False, False], [False, True], [True, False], [True, True]]],
             (dtypes.bool,),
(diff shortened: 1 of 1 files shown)


yongtang:
@BryanCutler I think the windows failure is a CI issue, can you bump the commit to trigger another build and give it a try?

tests/test_arrow_eager.py
```diff
@@ -989,7 +989,7 @@ def test_batch_fixed_lists(self):
 
         self.run_test_case(dataset, truth_data, batch_size=batch_size)
 
-    def test_batch_variable_length_list(self):
+    def test_batch_variable_length_list_batched(self):
         """Test batching with variable length lists raises error
     """
```
yongtang: @BryanCutler Since this file has been touched, would you mind to align the docstring as well? The misalign of docstring is part of the leftovers caused by black, see  issue #861 for more details.

With respect to this file, as you could see, 
```
    def test_batch_variable_length_list(self):
         """Test batching with variable length lists raises error
     """
```

should be (the second `"""` should intent two more spaces.
```
    def test_batch_variable_length_list(self):
         """Test batching with variable length lists raises error
        """
```

tests/test_arrow_eager.py
```diff
@@ -1010,6 +1010,26 @@ def test_batch_variable_length_list(self):
         with self.assertRaisesRegex(errors.OpError, "variable.*unsupported"):
             self.run_test_case(dataset, truth_data, batch_size=batch_size)
 
+    def test_batch_variable_length_list_unbatched(self):
+        """Test unbatched variable length lists
+    """
```
yongtang: Also here.

tests/test_arrow_eager.py
```diff
@@ -1010,6 +1010,26 @@ def test_batch_variable_length_list(self):
...
+            truth_data.output_types,
+            truth_data.output_shapes,
+            batch_size=batch_size,
+        )
+
+        self.run_test_case(dataset, truth_data, batch_size=batch_size)
+
     def test_unsupported_batch_mode(self):
         """Test using an unsupported batch mode
     """
```
yongtang: And here.

tests/test_arrow_eager.py
BryanCutler: Sure, I wasn't sure how the formatting was supposed to be

PR #865, opened 2020-03-20
Add tfio.experimental.audio.decode_mp3 support

yongtang:
This PR 
- adds tfio.experimental.audio.decode_mp3 support,
- Update minimp3 to 55da78c
- Merge audio and video into one library, as minimp4 serves both audio and video
- This PR also adds encode_mp3, through lame on Linux (macOS and Windows are not supported).

This PR closes #815

/cc @jjedele @faroit


jjedele:
@yongtang Looks great, thx for taking over!

jjedele:
I think the only thing left in #815 is this function to ensure a fixed output shape which we talked about (https://github.com/tensorflow/io/pull/815/files#diff-b91bd1e7c18232633b3715fe5e51530dR75-R100).

Not sure if it's helpful/general to add this here of it's simply enough to write on your own when you need it.

yongtang:
@jjedele As we discussed I think the `def _fix_shape` could be a separate API, something like `tfio.experimental.audio.normalize`. Would you like to work on it?

PR #869, opened 2020-03-20
Enable python 3.8 build on windows and linux

yongtang:
tensorflow 2.2.0rc1 adds python3.8 support for Windows and linux. This PR 
- enables python 3.8 build on windows and linux for tensorflow-io
- test for python 3.8 is not possible as tests dependency may not have python3.8 yet
- macOS does not have python3.8 as tensorflow has not released macOS 3.8 yet.

### DIFF
--- a/.github/workflows/build.wheel.sh
+++ b/.github/workflows/build.wheel.sh
@@ -22,6 +22,7 @@ fi
 
 if [[ "$PYTHON_VERSION" == "python3.8" ]]; then
   echo "TODO: Python 3.8 test is not supported yet as dependency might not be available"
+  exit 0
 fi
 
 if [[ $(uname) == "Linux" ]]; then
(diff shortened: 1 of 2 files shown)


.github/workflows/build.wheel.sh
```diff
@@ -19,6 +19,11 @@ if [[ "$#" -gt 0 ]]; then
   PYTHON_VERSION="${1}"
   shift
 fi
+
+if [[ "$PYTHON_VERSION" == "python3.8" ]]; then
+  echo "TODO: Python 3.8 test is not supported yet as dependency might not be available"
```
terrytangyuan: Not available where? GitHub Actions?

yongtang: @terrytangyuan Python 3.8 is available in GitHub Actions, and we could build 3.8 wheel. However, all the testing dependencies, like arrow, google cloud python wheels, may not be available as we pined to a specific version (old version) that does not have 3.8. So the test will fail.

We could gradually upgrade those test dependencies to newer version (if 3.8 is available), then we can re-enable python 3.8 tests.

Issue #171, opened 2019-04-04
Add tfio style test for tensorflow_io/BigTable


henrytansetiawan:
The original contrib.BigTable has its own internal test, but it would be fantastic if we add tensorflow_io style test using BigTable emulator.

yongtang:
@henrytansetiawan 👍 The PubSub emulator is similar, I could help with BigTable emulator if needed.

ricky3dec:
Hello @yongtang, can I please work on adding tests for gcloud bigtable using bigtable emulator? If that is not being taken up already?
Also, I see that pubsub related tests are skipped for now, any particular reason for that?

yongtang:
@ricky3dec That would be great! 👍 Let me know if you run into any issues.

For pubsub, at one point I encountered some issue with credential file on macOS so that was disabled at the moment. We definitely plan to re-enable it very soon. Help would be appreciated as well.

ricky3dec:
Sure @yongtang , please assign the issue to me.

Could you please point me to "original contrib.BigTable has its own internal test"? Would like to know what kind of tests were written earlier. Though I tried a bit to find out, but looks like this structure has changed over last few releases.

Thanks!

yongtang:
Thanks @ricky3dec!

It seems that google Bigtable has a emulator (similar to PubSub):
https://cloud.google.com/bigtable/docs/emulator

I think that might make our life a little easy.

Not sure about the original tensorflow.contrib.bigtable's internal test. Though from my past experience quite a few `tensorflow.contrib.`'s google cloud implementation use a C++ mock.

I am not sure we need to implement C++ mock as it may not exactly match the google cloud behavior, and it is (unnecessarily) harder to implement.

I would suggest to look into Bigtable emulator, and add python test against emulator. This might be a better way to test bigtable.

ricky3dec:
Yes @yongtang , I had gone through the emulator part, as you had originally suggested above.

I just wanted to know, if there are already some baseline tests that minimally we should be have. But considering your comments, would try to write tests that should cover basic usage of the operation that interfaces with "BigTable" alongwith some exception scenarios.

Would be sending out the initial review soon. Thanks again!

yongtang:
Thanks @ricky3dec! We have ongoing discussions about standardize tests in #65 

So far `io/tests/test_io_dataset_eager.py` is the initial efforts which tries to standardize the dataset.

You can also suggest any input to make the tests more consistent.

[...]

PR #879, opened 2020-03-27
Avro Dataset for TensorFlow IO

burgerkingeater:
RFC doc: https://github.com/tensorflow/io/blob/98f52b669496602757ee801afa6b91fab041482b/tensorflow_io/core/kernels/20190307-avro-dataset.md


burgerkingeater:
@fraudies  @ashahab @oliverhu  please sign the CLA if you haven't.
@yongtang some commits were authored under internal employees' username instead of their github account. What should we do with this?

fraudies:
@googlebot I consent

oliverhu:
@googlebot I consent.

ashahab:
@googlebot I consent

yongtang:
@burgerkingeater if you can confirm here I can override the CLA.

yongtang:
@burgerkingeater One thing to take note is that TF 2.1 changed `string` to `tstring` to fit an internal memory layout. So you need to change `tensor->flat<string>` to `tensor->flat<tstring>`.

One thing I noticed that you could use the following to avoid making big changes:
```
const string& value = tensor->scalar<tstring>()()
```

Notice `tstring` to `string` conversion above.

With the above at least much of the code could remains the same.

yongtang:
@burgerkingeater On your local machine, you need to install TF 2.2.0rc1 to see `tstring` in default.

yongtang:
@burgerkingeater See PR #851 for a similar change.

burgerkingeater:
@yongtang  thanks, i can confirm.

burgerkingeater:
@yongtang  thanks for the `tstring` suggestion. Since it's a change from `TF 2.1 `, why do we need to install `tf2.2.0rc1`?

yongtang:
@burgerkingeater ah sorry that was a typo,should be 2.2.

burgerkingeater:
@yongtang  that means we have to rebase on top of tf 2.2 anyway, right?

burgerkingeater:
Also by installing tf 2.2 rc, did you mean pip install tf-nightly ?

yongtang:
@burgerkingeater You could install with `pip install tensorflow==2.2.0rc1` for a specific version.

I don't think you need rebase, as we have minimal dependency on tensorflow itself. You only need to install 2.2.0rc1, rebuild tensorflow-io locally and fix any tstring related errors.

yongtang:
@burgerkingeater With PR #881 merged, TF has been adjusted to 2.2.0rc2 (was 2.2.0rc1). I don't think there is any fundamental differences though you might want to bump your local install to 2.2.0rc2 before build if possible.

burgerkingeater:
@yongtang  thanks, is this change backward compatible? If upgrade to `tstring`, will it work with `tf 2.1`?

yongtang:
@burgerkingeater No due to the fact that tensorflow still using C++ API, most of the tensorflow versions can not be mixed. There are some exceptions, e.g., TF 2.0 and 2.1 could be compatible but this is not the norm.

Note there is an ongoing effort by tensorflow team to convert to C API, which will allows back and forward compatible. But this likely will be some time before it happens.

burgerkingeater:
@yongtang made the `tstring` change, but the lint check is failing. 
```black:  ['/home/runner/work/io/io/bazel-out/host/bin/tools/lint/black_py.runfiles/org_tensorflow_io/tools/lint/black_python.py', '--check', 'tensorflow_io/core/python/api/v0/__init__.py']
Oh no! 💥 💔 💥
1 file would be reformatted.
```
is this the reason for failure?

yongtang:
@burgerkingeater The lint errors are easily fixable as they could be automatically reformatted. So I am not concerned with it.

However, the test `GitHub CI / Wheel 3.7 Linux (pull_request) ` (and 3.5, 3.6) will need to be fixed first. I think your PR deleted previous existing `cpx.json` and `test.bin` and that caused other tests to fail. Can you restore those deleted files?

burgerkingeater:
@yongtang  thanks, updated

yongtang:
@burgerkingeater I think you also need to either rebase or merge the latest master branch, as it seems to be out-of-sync with master.

yongtang:
@burgerkingeater Can you also restore the README.md (don't modify it)? We have tests that will test the content of the README.md. So if you modify the README.md without changing the related tests then it will fail.

burgerkingeater:
@yongtang thanks, yes i did. right now only `GitHub CI / Bazel Windows` is failing, but the failure doesn't seem to come from this PR.

tensorflow_io/core/BUILD
```diff
@@ -552,6 +558,7 @@ cc_binary(
         "//tensorflow_io/core:sql_ops",
         "//tensorflow_io/core:text_ops",
         "//tensorflow_io/ignite:ignite_ops",
+        "@libarchive",
```
yongtang, line 561: @burgerkingeater Can you remove this line? I think this is not needed.

burgerkingeater: thanks, updated

yongtang:
@burgerkingeater We are quite close! One thing I noticed is that the python test files `..._test.py` are not placed under the `tests` directory. This means they are not invoked in GitHub CI.

Can you move all python files of `..._test.py` to tests directory? The GitHub CI will automatically pick those files up and runs.

burgerkingeater:
@yongtang thanks, updated. How do i make sure they are invoked?

yongtang:
@burgerkingeater You can check the `GitHub CI / Wheel 3.6 Linux (pull_request) `, `Details` => `...` => `View raw logs`.

It looks like the disk is full on GitHub though. So some additional work might be needed to get pass it.

burgerkingeater:
@yongtang  thanks, is there anything needed from my side?

yongtang:
@burgerkingeater  Looking at the error further, it might be just a one-time error. Can you touch the last commit (modify the time by `git commit --amend -s`) and push again to trigger another run of the test?

burgerkingeater:
@yongtang still hit the same disk issue.

erwa:
@googlebot I consent.

yongtang:
@burgerkingeater PR #883 has been merged, you can rebase the PR so that the disk issue could be resolved.

burgerkingeater:
@yongtang  thanks, all tests passed, but `Wheel 3.6 Linux` finished in 1 min, and it doesn't seem to invoke those python unit tests.

yongtang:
@burgerkingeater Python 3.8 tests are disabled as the tests requires some other dependencies that may not be available for python 3.8 yet.

yongtang:
@burgerkingeater We are trying to clean up the top level directories as the code structure is used for both python and C++.

For C++ part, can you:
1. Create a directory `tensorflow_io/core/avro`, and place all C++ files in this PR under this directory (including util)?
2. Leave `tensorflow_io/core/ops/avro_ops.cc` alone (as it is under `tensorflow_io/core/ops`).

After that, can you lint the code with clang format, e.g.:
```
bazel run //tools/lint:lint -- clang -- tensorflow_io/core/kernels/avro_dataset_kernels.cc
```

Please only lint the files you touch, and leave unrelated files alone. This will clean up the C++ part.

We will take care of the python lint part after that.

burgerkingeater:
@yongtang  thanks, running this command `bazel run //tools/lint:lint  -- clang  -- tensorflow_io/core/avro/avro_dataset_kernels.cc` gave me below output
```
[chren@chren-ld2 tensorflow-io]$ bazel run //tools/lint:lint  -- clang  -- tensorflow_io/core/avro/avro_dataset_kernels.cc
DEBUG: Rule 'com_grail_bazel_toolchain' indicated that a canonical reproducible form can be obtained by modifying arguments sha256 = "d312c8e3a19ff843fce3065bb9ff40964401e8525674c842a5724b939cb6e1ac"
ERROR: /home/chren/work/gitli/tensorflow-io-mp_trunk_backup/tensorflow-io/tools/lint/BUILD:1:1: First argument of 'load' must be a label and start with either '//', ':', or '@'
ERROR: /home/chren/work/gitli/tensorflow-io-mp_trunk_backup/tensorflow-io/tools/lint/BUILD:1:1: file 'defs.bzl' was not correctly loaded. Make sure the 'load' statement appears in the global scope in your file
ERROR: /home/chren/work/gitli/tensorflow-io-mp_trunk_backup/tensorflow-io/tools/lint/BUILD:6:1: name 'lint' is not defined (did you mean 'int'?)
ERROR: /home/chren/work/gitli/tensorflow-io-mp_trunk_backup/tensorflow-io/tools/lint/BUILD:11:1: name 'lint' is not defined (did you mean 'int'?)
WARNING: /home/chren/work/gitli/tensorflow-io-mp_trunk_backup/tensorflow-io/tools/lint/BUILD:28:1: target 'pyupgrade' is both a rule and a file; please choose another name for the rule
WARNING: /home/chren/work/gitli/tensorflow-io-mp_trunk_backup/tensorflow-io/tools/lint/BUILD:54:1: target 'black' is both a rule and a file; please choose another name for the rule
WARNING: /home/chren/work/gitli/tensorflow-io-mp_trunk_backup/tensorflow-io/tools/lint/BUILD:85:1: target 'pylint' is both a rule and a file; please choose another name for the rule
ERROR: Skipping '//tools/lint:lint': no such target '//tools/lint:lint': target 'lint' not declared in package 'tools/lint' defined by /home/chren/work/gitli/tensorflow-io-mp_trunk_backup/tensorflow-io/tools/lint/BUILD
WARNING: Target pattern parsing failed.
ERROR: no such target '//tools/lint:lint': target 'lint' not declared in package 'tools/lint' defined by /home/chren/work/gitli/tensorflow-io-mp_trunk_backup/tensorflow-io/tools/lint/BUILD
INFO: Elapsed time: 0.104s
INFO: 0 processes.
FAILED: Build did NOT complete successfully (0 packages loaded)
FAILED: Build did NOT complete successfully (0 packages loaded)
```

yongtang:
@burgerkingeater Are you rebating to the latest master? I think your local branch might be missing several files.

burgerkingeater:
@yongtang yes i rebased with latest master.

yongtang:
@burgerkingeater I downloaded your branch and run:

```
bazel run //tools/lint:lint -- clang -- tensorflow_io/core/kernels/avro_dataset_kernels.cc 
```

Everything works fine. 

I am using Bazel 2.0.0 on macOS Catalina. Which platform and bazel version are you using?

burgerkingeater:
@yongtang i'm using bazel 0.25.2 and redhat 7

burgerkingeater:
@yongtang thanks for the help! i did lint on all c++ files and all check passed. Is there anything remaining?

yongtang:
@burgerkingeater The remaining ones are python namespace cleanup. As you might notice we are trying to limit the number of top level namespaces. See related discussion in https://github.com/tensorflow/io/issues/171#issuecomment-605667767

For avro related ops in this PR, I think the following probably makes sense:
1. Move tensorflow_io/avro directory to tensorflow_io/core/python/experimental/avro
2. Expose tfio.experimental.columnar.[AvroRecordDataset|parse_avro|make_avro_record_dataset|make_avro_dataset] by:
   a. Create a python file `tensorflow_io/core/python/api/columnar.py` similar to `tensorflow_io/core/python/api/audio.py`.
   b. Import symbols in 1. to  `tensorflow_io/core/python/api/columnar.py` 
   c. Add `tensorflow_io/core/python/api/columnar.py` to `tensorflow_io/core/python/api/experimental/__init__.py`.
3. Update tests to use `tfio.experimental.columnar.AvroRecordDataset` directly.

It looks like `AvroRecordDataset` takes the `TFRecordDataset` approach. To be honest this may not be the best way to handle dataset for Avro, as Avro is splittable/indexable while TFRecord can only be iterated. Following TFRecordDataset's implementation will make it a challenge to distribute the Avro across multiple nodes (either sharing or random access). The TFRecord also construct the subgraph in C++ which makes maintenance harder. (in comparison many the tensorflow-io's ops has been transformed to construct subgraph in python) 

But the above issues could be addressed gradually in follow up PRs later. As a first step let's get the PR merged into the repo for the upcoming TFIO 0.13.0/TF2.2 release (with a room to make further adjustment in the future).

burgerkingeater:
@yongtang thanks. I updated the PR per your suggestion and let me know if there's anything needed.

tensorflow_io/core/python/api/experimental/__init__.py
```diff
@@ -17,9 +17,14 @@
 from tensorflow_io.core.python.experimental.io_dataset_ops import IODataset
 from tensorflow_io.core.python.experimental.io_tensor import IOTensor
 from tensorflow_io.core.python.experimental.io_layer import IOLayer
+#from tensorflow_io.core.python.experimental.columnar import make_avro_dataset
+#from tensorflow_io.core.python.experimental.columnar import AvroRecordDataset
+#from tensorflow_io.core.python.experimental.columnar import parse_avro
+#from tensorflow_io.core.python.experimental.columnar import make_avro_record_dataset
```
yongtang, line 23: Please remove the commented lines.

tensorflow_io/core/python/experimental/avro/python/utils/__init__.py
```diff
@@ -0,0 +1 @@
+# This is a package
```
yongtang, line 1: Let's use an empty file for this. You can delete the `__init__.py`, then use `touch __init__.py` to create an empty file.

tests/__init__.py
```diff
@@ -0,0 +1 @@
+# This is a package.
```
yongtang, line 1: Don't change this file.

tensorflow_io/core/python/experimental/avro/python/README.md
```diff
@@ -0,0 +1,8 @@
+This folder only contains the tests.
```
yongtang, line 1: Is it necessary to have additional python modules `tensorflow_io/core/python/experimental/avro/python` and `tensorflow_io/core/python/experimental/avro/python/util`?

I think you can move `README.md`, and `avro_serialization.py` to `tensorflow_io/core/python/experimental/avro/`,  and delete all directories that does not have a concrete python source file.

yongtang:
@burgerkingeater The tests files under `tests` directory should be named with `test_xxx.py` to be picked up by CI automatically. Think you need to prefix those files you added with `test_`.

tensorflow_io/core/python/experimental/avro/python/README.md
burgerkingeater: updated

tests/__init__.py
burgerkingeater: updated

tensorflow_io/core/python/experimental/avro/python/utils/__init__.py
burgerkingeater: updated

tensorflow_io/core/python/api/experimental/__init__.py
burgerkingeater: updated

burgerkingeater:
@yongtang the latest build failure is not coming from this PR and seems to be related with 2.2 rc3. Do you have any insight?

yongtang:
@burgerkingeater See PR #897 which should fix the 2.2.0rc3 issue.

yongtang:
@burgerkingeater Are the tests for avro running in eager mode or graph mode? I assume it runs in eager mode?

Due to the way TF works, graph mode and eager mode could not run in the same process. 

In our GitHub CI we split the test run in different batches, `test_xxx.py` will run together in graph mode while `test_xxx_eager.py` will run together in eager mode.

So I think all your tests files in `tests` should be named as `test_xxx_eager.py`.

burgerkingeater:
@yongtang  thanks, i'm updating `./.github/workflows/build.yml` to cut windows/linux part for now so that i can get the feedback faster.

yongtang:
@burgerkingeater You could run tests locally first to at least make sure it passes on macOS or Linux:

```
TFIO_DATAPATH=bazel-bin python3 -m pytest -s -v test/test_xxx_eager.py
```

The above will run only specific tests in file `test/test_xxx_eager.py`

burgerkingeater:
thanks, @yongtang are you aware of this Segmentation fault? I saw other PR is also hitting this failure.

yongtang:
@burgerkingeater The crash on the other PR is different. For avro PR I noticed you used a lot of non-public APIs, e.g.,
```
from tensorflow.python.framework import dtypes as tf_types
from tensorflow.python.framework import ops
from tensorflow.python.ops import parsing_ops
```

They may have complications as TF's internal API system has many indirection for the purpose backward-compatible. Even though they may not be directly related to the crash, you may want to avoid using any APIs that are not public. (Public TF API means APIs exposed in tensorflow.org)

burgerkingeater:
@yongtang i have updated the PR with public APIs although there're some internal API usage remaining which don't have public api alternatives.

yongtang:
@burgerkingeater Can you point me to the location of the non-public APIs? I can take a look.

tensorflow_io/core/python/experimental/avro_dataset_ops.py
```diff
@@ -0,0 +1,533 @@
...
+from tensorflow.python.data.ops.dataset_ops import DatasetSource
+from tensorflow.python.data.experimental.ops import readers
+from tensorflow_io.core.python.ops import core_ops
+
+# Note: I've hidden the dataset because it does not apply the mapping for
+# sparse tensors
+# Such mapping is not possible inside the dataset, rather it needs to happen
+# through a map on the output of the dataset which is a map of keys to tensors
+# This can be changed when eager mode is the default and the only mode supported
+class _AvroDataset(DatasetSource):
```
yongtang, line 40: There is nothing special about `DatasetSource`:
https://github.com/tensorflow/tensorflow/blob/9239c61f201af6e8147a8ab7330edd87fea5a6ad/tensorflow/python/data/ops/dataset_ops.py#L2299-L2303

Can you:
1. use `_AvroDataset(tf.data.Dataset)`
2. add the following to the class (in fact it might not even be needed)?
   ```
   def _inputs(self):
       return []
   ```

tensorflow_io/core/python/experimental/avro_record_dataset_ops.py
```diff
@@ -0,0 +1,114 @@
...
+                                                         self._buffer_size,
+                                                         self._reader_schema)
+        super(_AvroRecordDataset, self).__init__(variant_tensor)
+
+    @property
+    def element_spec(self):
+        return tf.TensorSpec([], tf.dtypes.string)
+
+
+class AvroRecordDataset(dataset_ops.DatasetV2):
```
yongtang, line 57: `dataset_ops.DatasetV2` is really `tf.data.Dataset`

burgerkingeater: thanks, updated

tensorflow_io/core/python/experimental/avro_dataset_ops.py
burgerkingeater: updated.

burgerkingeater:
@yongtang thanks, here are the usage of internal APIs
https://github.com/tensorflow/io/pull/879/files#diff-7f6aa9ab6d69fcf5c0eb55337034dc8fR126
https://github.com/tensorflow/io/pull/879/files#diff-7f6aa9ab6d69fcf5c0eb55337034dc8fR590
https://github.com/tensorflow/io/pull/879/files#diff-dc5f32e21084f41cfc5a45c2f7061a95R88
https://github.com/tensorflow/io/pull/879/files#diff-dc5f32e21084f41cfc5a45c2f7061a95R98
https://github.com/tensorflow/io/pull/879/files#diff-dc5f32e21084f41cfc5a45c2f7061a95R41
https://github.com/tensorflow/io/pull/879/files#diff-c8f2bd016cf9bb8913df185a6f056a63R108

tensorflow_io/core/python/experimental/avro_record_dataset_ops.py
```diff
@@ -0,0 +1,121 @@
...
+    def __init__(self, filenames, buffer_size=None, reader_schema=None):
+        """Creates a `AvroRecordDataset`.
+        Args:
+          filenames: A `tf.string` tensor containing one or more filenames.
+          buffer_size: (Optional.) A `tf.int64` scalar representing the number of
+            bytes in the read buffer. 0 means no buffering.
+          reader_schema: (Optional.) A `tf.string` scalar representing the reader schema or None
+        """
+        self._filenames = filenames
+        self._buffer_size = convert.optional_param_to_tensor(
```
yongtang, line 41: The convert.optional_param_to_tensor is pretty simple:

https://github.com/tensorflow/tensorflow/blob/f40a875355557483aeae60ffcf757fc9626c752b/tensorflow/python/data/util/convert.py#L26-L35

Can you just copy the content here? The following should work:
```
def optional_param_to_tensor(argument_name,
                             argument_value,
                             argument_default=0,
                             argument_dtype=dtypes.int64):
  if argument_value is not None:
    return tf.convert_to_tensor(
        argument_value, dtype=argument_dtype, name=argument_name)
  else:
    return tf.constant(
        argument_default, dtype=argument_dtype, name=argument_name)

```

This will avoid future changes in TF that could break tensorflow/io

tensorflow_io/core/python/experimental/avro_dataset_ops.py
```diff
@@ -0,0 +1,627 @@
...
+    if shuffle:
+        dataset = dataset.shuffle(shuffle_buffer_size, shuffle_seed)
+    if num_epochs != 1:
+        dataset = dataset.repeat(num_epochs)
+
+    if any(isinstance(feature, tf.io.SparseFeature) for _, feature in features.items()):
+        # pylint: disable=protected-access
+        # pylint: disable=g-long-lambda
+        dataset = dataset.map(
+            lambda x: parsing_ops._construct_sparse_tensors_for_sparse_features(
```
yongtang, line 590: This `parsing_ops._construct_sparse_tensors_for_sparse_features` need to be replaced by a public API.

If you take a look at the source code in TF, you probably understand why it is dangerous to use it:
https://github.com/tensorflow/tensorflow/blob/5f931045eefcc4d2e8473fcf694a0c52adb7935f/tensorflow/python/ops/parsing_ops.py#L50-L57

There are too many indirections and it is already as a TOTO to switch the location. 

You can copy:
https://github.com/tensorflow/tensorflow/blob/6d0f422525d8c1dd3184d39494abacd32b52a840/tensorflow/python/ops/parsing_config.py#L661

And you can skip `RaggedFeature` part for now.

tensorflow_io/core/python/experimental/avro_record_dataset_ops.py
burgerkingeater: thanks, updated.

tensorflow_io/core/python/experimental/avro_dataset_ops.py
burgerkingeater: thanks, updated.

burgerkingeater:
@yongtang thanks for the help! i updated the PR. What do you recommend for the remaining non-public API usages?

```
        self._element_spec = structure.convert_legacy_structure(
            output_types, output_shapes, output_classes
        )
```
in `avro_dataset_ops.py`

`_create_dataset_reader` and `_create_or_validate_filenames_dataset` in `avro_record_dataset_ops.py`, seems their implementation is not complicated either. would you recommend copying the implementation?

yongtang:
@burgerkingeater `_create_dataset_reader and _create_or_validate_filenames_dataset` could be copied as they are pretty straightforward.

The `convert_legacy_structure` may not truly be necessarily. There was a long history of Dataset start with V1 which requires `dtypes` and `shapes` to be specified in a structured way. Later Dataset has been changed to V2 which uses `element_spec`, thus no need to specify `dtypes` and `shapes` separately. However, due to the volumes of legacy code in TF code base, `convert_legacy_structure` has been used in order to maintain both V1 and V2.

In tensorflow/io, all dataset ops are in V2 so `conversion` is not truly necessarily. In fact for any subclass of `tf.data.Dataset` (V2), the only requirement virtual methods are:
```
    def _inputs(self):
        return []

    @property
    def element_spec(self):
        return ......
```

If you can construct `element_spec` by your self, then you very much don't need `convert_legacy_structure`.

On the other hand, as TF 2.2.0 release is imminent, you can just try copy convert_legacy_structure` into tensorflow/io for now, so that only public APIs are used.

The clean up could be done slightly later, and that could be one of the prerequisite of moving out of `experimental`.

burgerkingeater:
@yongtang sounds good, thanks! I copied implementation for the remaining three, and enforced bazel formatting on all python and build files. Is there anything else remaining?

yongtang:
@burgerkingeater The last item is the:
```
from tensorflow.python.data.kernel_tests import test_base
```

While this class may seems to be large, the content that is relevant to this PR is just the assertEqual (including sparse and structured types). You can copy assertEqual and throw out everything else. Once this is in place we should be ready to go.

burgerkingeater:
@yongtang  thanks!! updated

yongtang:
@burgerkingeater Let's disable macOS test for now, to get CI cleared. Can you use pytest to skip macOS conditionally, e.g.:
```
pytest.mark.skipif(
                    sys.platform == "darwin", reason="macOS fails now"
),
```

Also, I noticed you use `unittest` in several places. Can you change `unittest` to `pytest` for unconditionally skip as well, with things like:
```
pytest.mark.skip(reason="")
```

yongtang:
@burgerkingeater The macOS failure seems to be caused by other test. Can you apply the following patch:
```diff
diff --git a/tests/test_kafka_eager.py b/tests/test_kafka_eager.py
index d954b4e..fd90b57 100644
--- a/tests/test_kafka_eager.py
+++ b/tests/test_kafka_eager.py
@@ -37,7 +37,7 @@ def test_kafka_io_tensor():
     assert len(kafka.to_tensor()) == 10
 
 
-@pytest.mark.xfail(reason="TODO")
+@pytest.mark.skip(reason="TODO")
 def test_kafka_output_sequence():
     """Test case based on fashion mnist tutorial"""
     fashion_mnist = tf.keras.datasets.fashion_mnist
```

Once this is applied, and `unittest` is changed to `pytest`, I think all tests will pass and we will be ready to go.

burgerkingeater:
@yongtang thanks, updated, but I have seen avro-related tests also failed intermittently in mac OS before. So we might still need to skip mac.

yongtang:
@burgerkingeater I think we are almost there. I will take another pass. 

I do want to mention that this will likely be the first step. The implementation will need quite some changes in order to be stabilizing. As an example, as you might already know, the `SaveInternal` is already changed in tf-nightly:

https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/framework/dataset.h

That means once we release v0.13.0, we may be forced to update the code base again immediately just for it to wok for the next release.

That is one of the main reasons that most of the ops implemented in tensorflow/io stays with `resource + index` and python graph, than using `C++ subgraph` (as was the case in this PR).

But those could be gradually improved once this batch of PR is merged.

yongtang:
@burgerkingeater I do see another macOS failure. Think there is definitely something going on with Avro's C++ kernel on macOS. Can you disable all macOS related tests for now?

burgerkingeater:
@yongtang thanks, mac os tests disabled.

yongtang:
@burgerkingeater I think you still need to replace `sparse_ops.sparse_merge(` with `tf.compat.v1.sparse_merge`. There is no v2 equivalent but even v1 public API is fine.

Also, can you merge all avro related files in tests into one file? (including `tests/test_parse_avro_eager.py`, `tests/test_make_avro_record_dataset_eager.py`, `tests/test_avro_record_dataset_eager.py`, `tests/avro_serialization.py`, `tests/avro_dataset_test_base.py`)

The reason is that, locally, developers prefer to use the following to do the testing:
```
TFIO_DATAPATH=bazel-bin python3 -m pytest -s -v tests/test_..._eager.py
```

However, when
```
import avro_serialization
import avro_dataset_test_base
```

are used in those tests, the path will not be visible. This is not an issue for CI as in CI the wheel file is installed.

If you merge those several files into one then it will be a lot easier to do local development.

burgerkingeater:
@yongtang thanks, updated.

burgerkingeater:
@yongtang the latest test case failure is from `test_io_dataset_eager.py::test_io_dataset_basic`, but the rest of the test went through.

yongtang:
@burgerkingeater That is like a one-time failure. Can you bump the PR? I think a small modify like `git commit --amend -s` will trigger the Github Actions to retest.

burgerkingeater:
@yongtang thanks! all green finally.

yongtang:
@burgerkingeater That is great! 👍 🎉 

By default tensorflow/io will squash the commits of the PR to keep the commit history clear. Given the PR consists of contributions from many people, and we want to keep the contributor's name in git history, I will try to import into master manually.

I may squash some commits (by contributor) though, as we still want to at least allows the commits to be clear so that it is possible to maintain in the future wit git history.

yongtang:
The PR has been cherry-picked and squashed into master now:
https://github.com/tensorflow/io/commits/master

I prefixed the squashed commits with `[AVRO]` prefix. This will help checking the history in the future.

The commits are also available in contributors page. Though since some of the commit authors does not have GitHub account or the GitHub account is now linked with the email in history, it may not shown all of them. This can be solved when email is linked to GitHub account. (We want to see all contributors' work properly shown up).

yongtang:
The GitHub CI is running https://github.com/tensorflow/io/actions/runs/86902135

Once tests pass this PR could be closed.

yongtang:
In the future, it would be preferred to see multiple smaller PR submissions. Smaller PRs are easier to be reviewed and easier for resolving any merge/cherry-pick conflict 😃

burgerkingeater:
@yongtang thanks for helping us getting this through! Yeah i have asked them to link email to github account. 

agree smaller PR is definitely preferable. This is the rare case as it’s long accumulated work from our side. In the future we can commit our change on top of TF IO and PR will be much smaller.

burgerkingeater:
@yongtang @annyan09023 updated email, although her github account is shown in the git commit history, but not in the contributor list. does it take sometime to reflect?

yongtang:
@burgerkingeater It may take some time I think. Actually even half an hour ago the number of contributors was still at 49 (even though it should be 53 after the cherry-pick). As long as the git account could be associated with commits eventually it will show up.

yongtang:
Since all tests pass now:

https://github.com/tensorflow/io/actions/runs/86902135

I think this PR could be closed. Thanks all for the great work! 👍 🎉 🎆

PR #881, opened 2020-03-27
Bump TF to 2.2.0rc2

yongtang:


yongtang:
Bumped to 2.2.0rc2, and also need to bump Kafka to 5.4.1 as 5.3.1 does not work with GitHub Actions' macOS anymore (a dns bug).

Issue #834, opened 2020-03-04
[tensorflow_io.oss] Fail to use tf.gfile.Rename to rename oss directory


zouxu09:
Code to reproduce:
```python
import tensorflow as tf
import tensorflow_io.oss

# fill with your configs
endpoint = ''
bucket = ''
access_id = ''
access_key = ''

baseOssPath = "oss://%s\x01id=%s\x02key=%s\x02host=%s/" % (
    bucket, access_id, access_key, endpoint)

tf.gfile.Rename(baseOssPath + "dir/", baseOssPath + "dir_copy/")
``` 

Error message:
```bash
Traceback (most recent call last):
  File "/tmp/t.py", line 30, in <module>
    tf.gfile.Rename(baseOssPath + "dir/", baseOssPath + "dir_copy/")
  File "/home/admin/.local/lib/python3.6/site-packages/tensorflow/python/lib/io/file_io.py", line 445, in rename
    compat.as_bytes(oldname), compat.as_bytes(newname), overwrite, status)
  File "/home/admin/.local/lib/python3.6/site-packages/tensorflow/python/framework/errors_impl.py", line 528, in __exit__
    c_api.TF_GetCode(self.status.status))
tensorflow.python.framework.errors_impl.InternalError: <****> errMsg: object not exists!
```

zouxu09:
@yongtang Could you please assign this issue to me? I'd like to fix this

terrytangyuan:
Could you try assigning yourself? GitHub doesn’t allow us to assign this to you since you are not a member of TensorFlow GitHub org.

yongtang:
I think it is possible to assign as long as you are the issue author.

PR #885, opened 2020-04-05
Fix issue: fail to use tf.gfile.Rename to rename oss directory (#834)

zouxu09:
The problem:
   When there's trailling '/' in directory path, the Rename function incorrectly put two '/'s in the path, which in oss is considered as a differecnt file (thus a 404 NotFound error is reported)

The fix is to correctly handle the url join


yongtang:
@zouxu09 PR #883 has been merged, you can rebase the PR so that the disk issue could be resolved.

yongtang:
All tests pass. Thanks @zouxu09 for the contribution! 👍 🎉

Issue #876, opened 2020-03-25
[bigquery-io] - Support for `offset` in ReadRowsRequest


nmatare:
The v2 BigQuery storage API [apiv1beta2](https://cloud.google.com/bigquery/docs/reference/storage/rpc/google.cloud.bigquery.storage.v1beta2#google.cloud.bigquery.storage.v1beta2.ReadRowsRequest) introduces an `offset` parameter which may be passed to select the starting position of the stream. While I understand the current implementation uses the beta1 API, it would be great to introduce this into the `tfio.bigquery.BigQueryReadSession` so as to start reading the underlying table at a prespecified index.

https://github.com/tensorflow/io/blob/ce58581f71778b3a797586fa1228a399c7b3a167/tensorflow_io/bigquery/kernels/bigquery_kernels.cc#L139

Currently, our workflow looks something like:
```
OFFSET = 1000
session = tf_io_bigquery_client.read_session()

dataset = (tf.data.Dataset
  .from_tensor_slices(session._streams)
  .flat_map(session.read_rows)
  .skip(OFFSET)
)
```
However, this requires us to iterate through the entire stream to skip `OFFSET` number of rows which naturally takes more time and incurres billing costs.

sshrdp:
Do you create a single stream? Because the offset applies to each stream read operation and having a single offset applied to multiple streams doesn't make a lot of sense. 
Is the goal to do sharding for distributed training?

nmatare:
Yup, exactly; we're only creating one stream (requested_streams=1) and then using flat_map to read the data sequentially. 

For each worker, we randomly select a starting position to begin reading the underlying table. We repeat this for each epoch -- randomly selecting the starting the position on each iteration for each worker.

The google.cloud.bigquery.storage docs appear to read that you can use multiple streams while still reading the data sequentially, but I'm unsure of this so we've just stuck to using one stream per each worker.

vlasenkoalexey:
Added code to pass through offset to BQ Storage API, also it does not return any data when offset is not 0. Filed internal bug for BQ team to address that (b/153579992).
It should work, BQ is updated.

Issue #875, opened 2020-03-25
Bigquery NUMERIC column not accessible


ghost:
When trying to get data from BigQuery, whenever a column of type NUMERIC is accessed, the following error is thrown when trying to map the column to **tf.double** data type:
```
>>> bqc = BigQueryClient()
>>> # a is STRING type, b is NUMERIC type, c is INTEGER type in BigQuery
>>> bqread = bqc.read_session("projects/"+PROJECT_ID, PROJECT_ID, TABLE,
...                          DATASET, ['a', 'b', 'c'],
...                          [tf.string, tf.double, tf.int64] )
>>> ds = bqread.parallel_read_rows()

>>> for d in ds.take(1):
...    print(d)
...    break


2020-03-26 01:33:56.053563: W tensorflow/core/framework/op_kernel.cc:1655] OP_REQUIRES failed at iterator_ops.cc:941 : Invalid argument: output type mismatch for column: b expected type: DT_STRING actual type: DT_DOUBLE                  
Traceback (most recent call last):                                                                                                                                                                                                           
  File "/usr/lib/python3.7/site-packages/tensorflow_core/python/eager/context.py", line 1897, in execution_mode                                                                                             
    yield                                                                                                                                                                                                                                    
  File "/usr/lib/python3.7/site-packages/tensorflow_core/python/data/ops/iterator_ops.py", line 659, in _next_internal                                                                                      
    output_shapes=self._flat_output_shapes)                                                                                                                                                                                                  
  File "/usr/lib/python3.7/site-packages/tensorflow_core/python/ops/gen_dataset_ops.py", line 2479, in iterator_get_next_sync                                                                               
    _ops.raise_from_not_ok_status(e, name)                                                                                                                                                                                                   
  File "/usr/lib/python3.7/site-packages/tensorflow_core/python/framework/ops.py", line 6606, in raise_from_not_ok_status                                                                                   
    six.raise_from(core._status_to_exception(e.code, message), None)                                                                                                                                                                         
  File "<string>", line 3, in raise_from                                                                                                                                                                                                     
tensorflow.python.framework.errors_impl.InvalidArgumentError: output type mismatch for column: b expected type: DT_STRING actual type: DT_DOUBLE [Op:IteratorGetNextSync]                                                                    
                                                                                                                                                                                                                                             
During handling of the above exception, another exception occurred:                                                                                                                                                                          
                                                                                                                                                                                                                                             
Traceback (most recent call last):                                                                                                                                                                                                           
[...]

PR #889, opened 2020-04-08
added BigQuery Numeric and offset argument support

vlasenkoalexey:
#876 - passing through offset argument
#875 - returning NUMERIC fields as strings
@sshrdp FYI


yongtang:
Thanks @vlasenkoalexey for the PR!

Can you run:
```
bazel run //tools/lint:lint -- black
```

to sanitize the format and conforming `black`?

You may also considering change the commit history to use your email, as the commits are not associated to your GitHub account it looks like? (the credit will goes to non-associated instead of your GitHub).

Otherwise the PR looks good to me.

PR #896, opened 2020-04-14
Switch to use more light weighted remotes package

terrytangyuan:
`devtools` package is too heavy and depends on many other packages. Here we only actually need `remotes::install_github()`.


yongtang:
The error is caused by out-of-space on GitHub Actions. Created a PR #903 to see if we can free up some spaces.

terrytangyuan:
Should we directly merge this simple doc change then?

PR #899, opened 2020-04-14
Fix hdf5 issue when resource tensor might be reused in graph incorrectly

### LINKED ISSUE #841: from_hdf5 fails with cryptic error messages
in version 0.12.0, installed from pip, I'm trying to use from_hdf function to create datasets, sample code is as follows:
```
import glob

import h5py
import os
import tensorflow_io as tfio
import tensorflow as tf
import numpy as np


def create_datasets(runpath, fileCnt=10):
    os.makedirs(runpath, exist_ok=True)
    for i in range(fileCnt):
        f = h5py.File(f'{runpath}/file_{i}.h5', 'w')
        totalSamples = np.random.randint(50000, 100000)
        f.create_dataset('features', data=np.random.random((totalSamples, 60)))
        f.create_dataset('targets', data=np.random.random((totalSamples, 3)))
        f.close()


def get_paths(runpath):
    return glob.glob(f'{runpath}/*.h5')


if __name__ == '__main__':
    runpath = '/tmp/data'
    if not os.path.exists(runpath):
        create_datasets(runpath)

    cnt = 0
    for p in get_paths(runpath):
        try:
            features = tfio.IODataset.from_hdf5(p, '/features')
            targets = tfio.IODataset.from_hdf5(p, '/targets')
            dataset = tf.data.Dataset.zip((features, targets))

            for t in dataset:
                cnt += t[0].shape[0]
                
        except Exception as e:
            print(f"Failed going through {p}")
            raise e
        print(f"Success going through {p}")


    print(f"Iterated {cnt} items")
```

Although the first file always succeeds, consecutive files (may) fail, where I get an error like:

```
Success going through /tmp/data/file_0.h5
2020-03-06 14:54:30.187273: W tensorflow/core/framework/op_kernel.cc:1655] OP_REQUIRES failed at hdf5_kernels.cc:612 : Invalid argument: dimension [0] out of boundary: start=64512, slice=1024, boundary=65153
2020-03-06 14:54:30.187548: W tensorflow/core/framework/op_kernel.cc:1655] OP_REQUIRES failed at hdf5_kernels.cc:612 : Invalid argument: dimension [0] out of boundary: start=64512, slice=1024, boundary=65153
2020-03-06 14:54:30.187601: W tensorflow/core/framework/op_kernel.cc:1655] OP_REQUIRES failed at iterator_ops.cc:941 : Invalid argument: dimension [0] out of boundary: start=64512, slice=1024, boundary=65153
         [[{{node IO>HDF5ReadableRead}}]]
Failed going through /tmp/data/file_1.h5
  File "/opt/anaconda/lib/python3.7/site-packages/tensorflow_core/python/eager/context.py", line 1897, in execution_mode
    yield
  File "/opt/anaconda/lib/python3.7/site-packages/tensorflow_core/python/data/ops/iterator_ops.py", line 659, in _next_internal
  File "/opt/anaconda/lib/python3.7/site-packages/tensorflow_core/python/ops/gen_dataset_ops.py", line 2479, in iterator_get_next_sync
  File "/opt/anaconda/lib/python3.7/site-packages/tensorflow_core/python/framework/ops.py", line 6606, in raise_from_not_ok_status
tensorflow.python.framework.errors_impl.InvalidArgumentError: dimension [0] out of boundary: start=64512, slice=1024, boundary=65153
         [[{{node IO>HDF5ReadableRead}}]] [Op:IteratorGetNextSync]

During handling of the above exception, another exception occurred:

  File "/tmp/test.py", line 40, in <module>
    raise e
  File "/tmp/test.py", line 35, in <module>
    for t in dataset:
  File "/opt/anaconda/lib/python3.7/site-packages/tensorflow_core/python/data/ops/iterator_ops.py", line 630, in __next__
    return self.next()
  File "/opt/anaconda/lib/python3.7/site-packages/tensorflow_core/python/data/ops/iterator_ops.py", line 674, in next
    return self._next_internal()
  File "/opt/anaconda/lib/python3.7/site-packages/tensorflow_core/python/data/ops/iterator_ops.py", line 665, in _next_internal
  File "/opt/anaconda/lib/python3.7/contextlib.py", line 130, in __exit__
  File "/opt/anaconda/lib/python3.7/site-packages/tensorflow_core/python/eager/context.py", line 1900, in execution_mode
    executor_new.wait()
  File "/opt/anaconda/lib/python3.7/site-packages/tensorflow_core/python/eager/executor.py", line 67, in wait
    pywrap_tensorflow.TFE_ExecutorWaitForAllPendingNodes(self._handle)
tensorflow.python.framework.errors_impl.InvalidArgumentError: dimension [0] out of boundary: start=64512, slice=1024, boundary=65153
         [[{{node IO>HDF5ReadableRead}}]]

```

Is this an expected error? If so, is it possible to make the error reason more apparent?

yongtang:
This PR fixes #841

The reason for the issue was that the resource tensor might be reused in graph incorrectly
in tensorflow code base (likely some change in behavior on how the resource should be intepeareted).

This PR assign a unique container and name to make sure resource created for hdf5 is never re-used.

(The same issue may happen in other ops as well?)

### DIFF
--- a/tensorflow_io/core/python/ops/hdf5_dataset_ops.py
+++ b/tensorflow_io/core/python/ops/hdf5_dataset_ops.py
@@ -25,13 +25,13 @@ class HDF5IODataset(tf.data.Dataset):
 
     def __init__(self, filename, dataset, spec=None, internal=True):
         """HDF5IODataset."""
-        with tf.name_scope("HDF5IODataset") as scope:
+        with tf.name_scope("HDF5IODataset"):
             assert internal
 
             # TODO: unique shared_name might be removed if HDF5 is thead-safe?
             resource, _ = core_ops.io_hdf5_readable_init(
                 filename,
-                container=scope,
+                container="HDF5IODataset",
                 shared_name="{}/{}".format(filename, uuid.uuid4().hex),
             )
             if tf.executing_eagerly():
--- a/tensorflow_io/core/python/ops/hdf5_dataset_ops.py
+++ b/tensorflow_io/core/python/ops/hdf5_dataset_ops.py
@@ -14,6 +14,8 @@
 # ==============================================================================
 """HDF5Dataset"""
 
+import uuid
+
 import tensorflow as tf
 from tensorflow_io.core.python.ops import core_ops
 
@@ -23,11 +25,15 @@ class HDF5IODataset(tf.data.Dataset):
 
     def __init__(self, filename, dataset, spec=None, internal=True):
         """HDF5IODataset."""
-        with tf.name_scope("HDF5IODataset"):
+        with tf.name_scope("HDF5IODataset") as scope:
             assert internal
 
             # TODO: unique shared_name might be removed if HDF5 is thead-safe?
-            resource, _ = core_ops.io_hdf5_readable_init(filename)
+            resource, _ = core_ops.io_hdf5_readable_init(
+                filename,
+                container=scope,
+                shared_name="{}/{}".format(filename, uuid.uuid4().hex),
+            )
             if tf.executing_eagerly():
                 shape, dtype = core_ops.io_hdf5_readable_spec(resource, dataset)
                 dtype = tf.as_dtype(dtype.numpy())
--- a/tests/test_hdf5_eager.py
+++ b/tests/test_hdf5_eager.py
@@ -1,0 +1,65 @@
+# Copyright 2020 The TensorFlow Authors. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may not
+# use this file except in compliance with the License.  You may obtain a copy of
+# the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.  See the
+# License for the specific language governing permissions and limitations under
+# the License.
+# ==============================================================================
+"""Tests for HDF5 file"""
+
+import os
+import glob
+import shutil
+import tempfile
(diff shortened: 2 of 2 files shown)


yongtang:
Still have issues with the underlying hdf5 library.

yongtang:
The issue was a container name restriction. All passes now.

BryanCutler:
@yongtang do you know if this is an issue with all other ops too?

yongtang:
@BryanCutler This more or less impact several things. And because TF 2.0+ enables eager mode by default, combined with autograph, things are little bit complicated.

In fact even the fix in this PR only addressed one scenario. Other scenarios may still fail. I am still trying to get a deeper look at how to avoid the issue altogether. But so far, below is the summary:

- TF's graph has two types of inputs: attributes and inputs. Attributes are compile time attribute, and are part of the graph (will not change). Tensor are running time data that will change each run.
- In TF 2.0, as eager mode is the default, and autograph is enabled by default as well, the data (in numpy, or python type) passed in are converted implicitly to tensor or attributes implicitly. This is the place where it gets dangerous: attributes will not be replaced at each run because it is done in compile time.
- Note in TF 2.0, the tf.data also enabled autograph by default so there are quite a few places that could get unintended side effects.
- Note this PR's fix still passes attributes to create a `Resource` tensor. So it will still fail in some scenarios. I am looking into modify the `Resource` tensor creation kernel to address it.

I think in order to avoid those so many potential side effects, the best way is to build everything in graph mode with autograph disabled. And we test in graph mode as well so that any issue can be discovered in tests.

In TF 1.x, graph mode is not very easy to use. We also really don't want to get back to the old ugly "session.run()" types of code. Really a pain.

In TF 2.0, graph mode is not well documented. However, so far my understanding is that,
```
tf.function(autograph=False)
```
is the most reliable way to run a function in graph mode. If there is ever any side effects, we will quietly find out.

PR #901, opened 2020-04-17
Add more instruction in README for Xcode installation

burgerkingeater:


yongtang:
@burgerkingeater In our test, we actually cut the instruction from each section and test that it successfully runs. See this part for details:

https://github.com/tensorflow/io/blob/master/.github/workflows/build.yml#L35-L46

For that reason, if your instruction is not impactful, you can insert as part of the instruction on macOS:
https://github.com/tensorflow/io#macos

Or you will need to place instructions after the main code block "```" segment and in a separate code block.

burgerkingeater:
@yongtang  thanks, updated.

PR #922, opened 2020-04-28
HDF5 and ResourceOp Fix

yongtang:
This PR tries to address the ResourceOp issue where attriubte is passed
and caused resource fixed at graph compile time.

Related to 898 and several others.

### DIFF
--- a/tensorflow_io/core/kernels/io_kernel.h
+++ b/tensorflow_io/core/kernels/io_kernel.h
@@ -13,8 +13,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 ==============================================================================*/
 
-#include "tensorflow/core/framework/resource_mgr.h"
-#include "tensorflow/core/framework/resource_op_kernel.h"
+#include "tensorflow/core/framework/op_kernel.h"
 #include "tensorflow/core/lib/strings/scanner.h"
 
 namespace tensorflow {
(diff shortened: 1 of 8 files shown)


tensorflow_io/core/kernels/io_kernel.h
```diff
@@ -0,0 +1,120 @@
...
+  // to kernel. Ideally the resource should be deleted when it is no longer held
+  // by anyone, but it would break backward compatibility.
+  ~IOResourceOpKernel() override {
+    if (resource_ != nullptr) {
+      resource_->Unref();
+      if (cinfo_.resource_is_private_to_kernel()) {
+        if (!cinfo_.resource_manager()
+                 ->template Delete<T>(cinfo_.container(), cinfo_.name())
+                 .ok()) {
+          // Do nothing; the resource can have been deleted by session resets.
```
terrytangyuan, line 47: can -> could/would

yongtang: @terrytangyuan Sorry it is too early, It is not a complete fix yet, need several round of patches.

yongtang:
The issue is more or less fixed, though TF's ResourceMgr seems to be not reliable enough to take runtime deletion of the resource entry. ResourceMgr by default will just keep it forever until session reset. Even the destructor of the kernel op will not invoke the deletion.

A commit in this PR override some behavior to delete actively during kernel ops destructor. However, it causes some issues with parallel, likely due to multiple subgraph of in tf.data pipeline. A global lock is used to resolve the issue though this may limit the parallel performance.

A follow up would be to replace ResourceMgr with better customized implementation.

yongtang:
The issue is further complicated as hdf5 itself is not multi-threaded.

yongtang:
This PR is now ready and all tests pass.

Summary:
- ResourceMgr in TF has been replaced by a unodered_map+shared_ptr implementation.
- When a kernel ops passes `container` as attribute and `shared` as tensor, the kernel ops will create a `handle` and place it in unordered_map. (Note ResourceMgr in TF only allows `shared_name` to be attributes as well).
- The unodered_map is shared across all instances of the kernel ops class so the `handle` might be reused across multiple instances of the same class.
- Each instance of the same class can only allocate one `handle` (we could increase the limit later).
- If an instance needs to create another handle when a different tensor is passed to the same instance (note one kernel ops class could have multiple instances), then the old one created by the same instance will be deleted.

Overall, this is not the most elegant solution, but it solves many issues in HDF5 by users.

In the future we should find a better way to deal with this situation. One challenge is that TF's graph optimization is aggressively optimize so we have to make sure any mechanism will not be optimized out the node of the graph. I more or less have a rough thinking about it. That will be handled in follow up PRs in the future.

/cc @terrytangyuan  @BryanCutler

PR #934, opened 2020-05-04
 Skip Avro tests on macOS

yongtang:
It looks like Avro tests is causing some issues on macOS. This PR skips all Avro tests for macOS.


yongtang:
@terrytangyuan in the meantime, it might makes sense to skip avro tests for now, as the avro was only added under the `tfio.experimental.columnar` (#879)

PR #951, opened 2020-05-08
Bump to TF 2.2.0

yongtang:


yongtang:
@terrytangyuan The issue is due to the C++ API instability of TF core. Every TF version has to be build specifically pin to each release. Patch release change (2.2.0 => 2.2.1) **might** be OK but minor release change (2.1 => 2.2) definitely will not work.

In fact, cross minor releases (2.1 => 2.2) will not even going to compile in many cases.

For example, tf-nightly needs quite some code change  (https://github.com/tensorflow/io/commit/ec79e7c18b55537c4d343a61bb8447e7f8f8be3a) before it is able to build. That is why we have to open a nightly branch so that TPU file system could be merged to nightly temporarily.

Different APIs in TF core also have different stabilities. From the observation:
- **Python public** API (stable): This is the most stable API that will never break across versions. 
  This is the reason only public API should be used in our python code (/cc @burgerkingeater)
- **C API** (stable): This is also very stable, however, the only thing that might be used at the moment is the modular file system API that will be released soon.
  Once modular file system API is out we will need to covert all of our file systems to C API to allow cross version stability.
- **C++ kernel ops itself** (stable): This is also stable as TF core has the check to make sure it will not break compatibility. If there are any breaking changes, a new version will be added (e.g., `Stack` + `StackV2`).
- **C++ API to build a kernel ops** (almost stable): This is almost stable, not because of the API guarantee, but because there are thousands of kernel C++ ops in TF core as well. Anyone want to change the API in TF core, will have to go through thousands of kernel C++ ops to make it happen.
  This is also the reason why we want to stay with C++ kernel ops implantation, than a C++ subgraph construction like `tf.data`. We try to use C++ kernel API to build low level ops that could be exposed as `tf.data`.
- **C++ API used for tf.data** (less stable): This is similar to C++ API used for ops. However, this is less guarded as there are fewer ops than kernel ops (thousands). It is easer for code change in TF core.
   We prefer to use kernel ops and emulate the behavior of tf.data for that reason.
- Non-public python APIs and internal C++ code: (no stability).

There are plans to make API stable in TF core and I see quite significant progress in this direction. But this is not the case now. So essentially we do have to pin a specific version.

PR #961, opened 2020-05-12
Added Visual Studio Code tutorial, and updated README

vlasenkoalexey:


yongtang:
@terrytangyuan I think the content is suitable for part of the https://tensorflow.org/io, so the tutorial format might be good for developers to follow.

vlasenkoalexey:
@yongtang is right, I'm following existing tf.io docs structure to make sure that tensorflow docs integration works as expected. Will be happy to move / restructure if necessary.

lamberta:
Seems out of place for tutorials: https://www.tensorflow.org/io/overview ---which I expect to contain example of tasks that use TFIO.

But projects can have install and build guides, for example: https://www.tensorflow.org/federated/install
Quickly looking through and this looks like a build guide, yes? (I.e. not just how to use VS Code)

yongtang:
@lamberta @vlasenkoalexey I think have a page of install like https://www.tensorflow.org/federated/install would make sense. Do you know what will need to be updated or change in the PR?

Should the content be placed in a directory called `docs/install` instead (was in `docs/tutorial`)?

lamberta:
> Should the content be placed in a directory called docs/install instead (was in docs/tutorial)

Yeah, I'd move it out of tutorials to the /docs root. Probably don't need a docs/install directory until there's more content to put there.

A lot of projects have something similiar, here's the TFQ install: https://www.tensorflow.org/quantum/install (and here's how their [_book.yaml](https://github.com/tensorflow/quantum/blob/master/docs/_book.yaml)) is set up)

vlasenkoalexey:
Moved vscode instructions to docs root, @lamberta please take a look

lamberta:
Thanks.

Need to add to the leftnav somewhere. Not sure how this project is set up, I see a [docs/_toc.yaml](https://github.com/tensorflow/io/blob/master/docs/_toc.yaml) file and a [docs/tutorials/_toc.yaml](https://github.com/tensorflow/io/blob/master/docs/tutorials/_toc.yaml) file---I assume this should be added to the former cc: @MarkDaoust

MarkDaoust:
Yes, docs/_toc.yaml makes the first section,  tutorials/_toc.yaml makes the second section in the left-nav here: 

https://www.tensorflow.org/io/overview

vlasenkoalexey:
As far as I understand it should go below Overview section in docs/_toc.yaml similarly to
https://www.tensorflow.org/quantum/install
Please take a look

PR #965, opened 2020-05-14
Fix several issues in avro_dataset_ops.py

burgerkingeater:
1. fix wrong tf APIs
2. rename check_none to check_not_empty
3. implement base class's abstract method in _AvroDataset
4. fix FixedLenFeature shape check

### DIFF
--- a/tensorflow_io/core/python/experimental/avro_dataset_ops.py
+++ b/tensorflow_io/core/python/experimental/avro_dataset_ops.py
@@ -52,15 +52,13 @@ class _AvroDataset(tf.data.Dataset):
         avro_data_buffer_size,
     ):
 
-        self._filenames = tf.ops.convert_to_tensor(
-            filenames, tf.string, name="filenames"
-        )
+        self._filenames = tf.convert_to_tensor(filenames, tf.string, name="filenames")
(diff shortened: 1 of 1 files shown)


yongtang:
@burgerkingeater I think the CLA issue is caused by your first commit where you use the `Cheng Ren <chren@xxx.biz>`.

Can you update your commit to set the author to the email you use for CLA? (the second commit is ok, just the first commit)

You can use the following to update:
```
% git rebase -I HEAD~2

# change the line with the first commit 86a2dfef004f0bf5df163c69d9d28689ddc56e2c from "pick" to "e"

# then use the following to update:
git commit --amend -s --reset-author

git rebase --continue

# finally push with -f
git push -f origin tf-io-oss-master-fix_avro_dataset_ops
```

burgerkingeater:
@yongtang  thanks, updated.

yongtang:
@burgerkingeater Looks like the Linux test is flaky. Can you try touch the commit and push, to trigger GitHub again?

burgerkingeater:
@yongtang updated

yongtang:
@burgerkingeater It looks ok for now. Though It is a surprise some of the incorrect API usage still passes tests. If they are orphaned code they should be removed. Otherwise additional tests could be added to cover.

PR #969, opened 2020-05-19
Adding memcached_file_block_cache to gstpufs

marioecd:


marioecd:
Any idea how to avoid these bazel macos and bazel windows checks?

marioecd:
It's supposed to only include gs_tpu_file_system when it's not darwin. I don't know why macos keeps trying to bazel build gs_tpu_file_system.

yongtang:
@marioecd I will take a look.

PR #970, opened 2020-05-19
Mario1 - scratch PR

marioecd:
A scratch PR I am using to try to debug some darwin MacOs build errors

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -934,3 +934,14 @@ http_archive(
         "https://launchpad.net/libmemcached/1.0/1.0.18/+download/libmemcached-1.0.18.tar.gz",
     ],
 )
+
+http_archive(
+    name = "farmhash_archive",
+    build_file = "//third_party:farmhash.BUILD",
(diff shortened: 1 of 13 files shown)


yongtang:
@marioecd The nightly branch has been rebased. I think you may need to create a new PR again.

You can leave your original nightly branch alone so that I can see if I can fix the macOS build issue.

yongtang:
@marioecd To make sure there is no changing factors, I also pinned nightly build's TF version to `tf-nightly==2.3.0.dev20200519`. 

If your PR requires a different or newer TF-nightly, you can modify it in your PR at this file:

https://github.com/tensorflow/io/blob/nightly/tensorflow_io/core/python/ops/version_ops.py

marioecd:
Sounds good on the version pinning. Thanks.

On the macos build failure, looks like adding some conditional:darwin logic around any target that depends on libmemcached fixes the issue, though it look so ugly since it looks like I need to do it to a few targets.

PR #979, opened 2020-05-21
Adding memcached file system and cache to gstpufs

marioecd:
Active PR to merge the memcached file system and its cache into the gstpufs subdir

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -959,3 +959,14 @@ http_archive(
         "https://launchpad.net/libmemcached/1.0/1.0.18/+download/libmemcached-1.0.18.tar.gz",
     ],
 )
+
+http_archive(
+    name = "farmhash_archive",
+    build_file = "//third_party:farmhash.BUILD",
+    sha256 = "6560547c63e4af82b0f202cb710ceabb3f21347a4b996db565a411da5b17aba0",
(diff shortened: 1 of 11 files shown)


yongtang:
@marioecd  The CI error was caused by two issues, one is the windows python version (due to GitHub CI's upgrade), the other one requires a fix up of libmemecached.BUILD file to include some hashkit dependencies.

I think I can get those issues addressed. Let me update the nightly branch shortly.

yongtang:
@marioecd Actually, I think I can just merge this one and create a follow up commit to address the issue mentioned. This can avoid the rebase on your side again.

yongtang:
@marioecd The PR has been merged in nightly branch now:

https://github.com/tensorflow/io/commits/nightly

I updated the nightly branch on top of that. Now all build should be pass on macOS/linux/windows.

I also fast forward nightly branch so that it matches latest master. If you create new PRs for nightly branch in the future you may need to rebase again.

I understand this is not ideal and sorry for you inconvenience. Though this is the limitation of TF's recent API C++ API change.

We plan to update the master branch to a new version of TF once TF 2.3.0rc0 (release candidate 0) is out. After that I think you can just create PR against master, without the need to rebase against nightly that might be falling behind.

marioecd:
Thanks for the help! I actually think I may not need to add another new PR until TF 2.3 is out.

PR #939, opened 2020-05-05
Arrow format support for BigQuery reader

vlasenkoalexey:
@sshrdp FYI

PR #995, opened 2020-06-12
Support Repeated field in BigQuery api

### LINKED ISSUE #627: BigQueryClient query with arrays
Hey guys. I am attempting to pull a table from bigquery that has nested arrays of values inside of it. I would like to pull it into the dataset then do a flat_map across it to unnest the data.

Currently it seems that the BigQueryClient doesn't actually support arrays which is needed for my work with TensorflowRanking

here is a snippet of code:
```python
from tensorflow_io import bigquery as tfiobq

def build_dataset(*args, **kwargs):
    return (tfiobq.BigQueryClient()
        .read_session(
            "projects/" + billing_project,
            project_id=project_id,
            dataset_id=dataset_id,
            table_id=table_id,
            selected_fields=["feature1", "feature2", "labels"],
            output_types=[tf.string, tf.int64, tf.int64],
        )
        .parallel_read_rows()
    )
```

Here feature2 and labels are both arrays in the table

I received the following exception when I try to get the next value from the dataset:

```
2019-11-12 01:05:55.884373: W tensorflow/core/framework/op_kernel.cc:1622] OP_REQUIRES failed at iterator_ops.cc:929 : Invalid argument: unsupported data type: 10
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python3.6/site-packages/tensorflow_core/python/data/ops/iterator_ops.py", line 622, in __next__
    return self.next()
  File "/usr/local/lib/python3.6/site-packages/tensorflow_core/python/data/ops/iterator_ops.py", line 666, in next
    return self._next_internal()
  File "/usr/local/lib/python3.6/site-packages/tensorflow_core/python/data/ops/iterator_ops.py", line 651, in _next_internal
  File "/usr/local/lib/python3.6/site-packages/tensorflow_core/python/ops/gen_dataset_ops.py", line 2673, in iterator_get_next_sync
    _six.raise_from(_core._status_to_exception(e.code, message), None)
tensorflow.python.framework.errors_impl.InvalidArgumentError: unsupported data type: 10 [Op:IteratorGetNextSync]
```

RuhuaJiang:
Fixes #627 

**Goal**
Support decoding BigQuery repeated field with primitive type as Tensor.

**Non-goal**
Support repeated field with non-primitive type such as Record.

**Impact**
This change fills an important gap in the BigQueryDataset API. The repeated field in BigQuery can be used to represent what equivalent to tf.io.VarLenFeature (normally used to represent Sparse features) in TFExample.  Combining with the existing capability of generating scalar Tensor for nullable/required BigQuery field,  this enables featurewise on-pair with TFExampleDataset (TFRecordDataset+ TFexample parse function https://www.tensorflow.org/tutorials/load_data/tfrecord) . 

See README.md to see how to use this API.  In high level, for a repeated field, it returns a rank-1 variable-length dense tensor  

Let's say,  `_tensor` is the Tensor mentioned above, user can do 
```
l = tf.size(_tensor, tf.int64)
indices = tf.reshape(tf.range(l, dtype=tf.int64), [l, 1])
_SparseTensor= tf.SparseTensor(indices=indices,
values=_tensor,
dense_shape=[l])
```
to get a SparseTensor. Then they can use `dataset.batch` with this SparseTensor without a problem. 

The behaviour of returning this kind of SparseTensor is exactly aligned with how to decode tf.example tf.io.VarLenFeature (which contains a list of values, just like the REPEATED field in BigQuery, returns a SparseTensor).

### DIFF
--- a/tensorflow_io/bigquery/README.md
+++ b/tensorflow_io/bigquery/README.md
@@ -142,7 +142,7 @@ work that Tensor with dataset.batch, then you can use code like
 dataset = read_session.parallel_read_rows()
 def sparse_dataset_map(features, sparse_column_names):
   """
-  Map sparse repeated columns to tf.SparseTensor.
+  Map repeated columns to tf.SparseTensor.
   This matches how VarLenFeature is decoded from tf.Example datasets.
   """
   for col_name in sparse_column_names:
--- a/tensorflow_io/bigquery/README.md
+++ b/tensorflow_io/bigquery/README.md
@@ -89,15 +89,17 @@ It also supports reading repeated mode BigQuery column (each field contains arra
 form like
 
 ```
-        { "field_a_name": {"mode": "repeated", output_type: dtypes.int64},
-          "field_b_name": {"mode": "nullable", output_type: dtypes.string},
+        { "field_a_name": {"mode": BigQueryClient.FieldMode.REPEATED, output_type: dtypes.int64},
+          "field_b_name": {"mode": BigQueryClient.FieldMode.NULLABLE, output_type: dtypes.string},
           ...
-          "field_x_name": {"mode": "repeated", output_type: dtypes.string}
+          "field_x_name": {"mode": BigQueryClient.FieldMode.REQUIRED, output_type: dtypes.string}
         }
 
 ```
-"mode" is BigQuery column attribute concept, it can be 'repeated', 'nullable' or 'required'.The output field order is unrelated to the order of fields in
-selected_fields. If "mode" not specified, defaults to "nullable". If "output_type" not specified, DT_STRING is implied for all Tensors.
+"mode" is BigQuery column attribute concept, it can be 'repeated', 'nullable' or 'required' (enum BigQueryClient.FieldMode.REPEATED, NULLABLE, REQUIRED).The output field order is unrelated to the order of fields in
+selected_fields. If "mode" not specified, defaults to "nullable". If "output_type" not specified, DT_STRING is implied for all Tensors. 
+
+'repeated' is currently only supported when data_format = BigQueryClient.DataFormat.AVRO (which is default).
 
 ```
 from tensorflow.python.framework import ops
@@ -117,9 +119,10 @@ def main():
       "projects/" + GCP_PROJECT_ID,
       DATASET_GCP_PROJECT_ID, TABLE_ID, DATASET_ID,
       selected_fiels={
-          "field_a_name": {"mode": "repeated", output_type: dtypes.int64},
-          "field_b_name": {"mode": "nullable", output_type: dtypes.string},
-          "field_c_name": {"mode": "repeated", output_type: dtypes.string}
+          "field_a_name": {"mode": BigQueryClient.FieldMode.REPEATED, output_type: dtypes.int64},
+          "field_b_name": {"mode": BigQueryClient.FieldMode.NULLABLE, output_type: dtypes.string},
+          "field_c_name": {"mode": BigQueryClient.FieldMode.REQUIRED, output_type: dtypes.string}
+          "field_d_name": {"mode": BigQueryClient.FieldMode.REPEATED, output_type: dtypes.string}
         }
       requested_streams=2,
       row_restriction="num_characters > 1000",
@@ -151,7 +154,7 @@ def sparse_dataset_map(features, sparse_column_names):
     features[col_name] = tf.SparseTensor(indices=indices,
                                          values=features[col_name],
                                          dense_shape=[l])
-dataset_can_be_batched = dataset.map(lambda features: sparse_dataset_map(features, ["field_a_name", "field_b_name", "field_c_name"]))
+dataset_can_be_batched = dataset.map(lambda features: sparse_dataset_map(features, ["field_a_name", "field_d_name"]))
 
 ```
 to map that that as a SparseTensor first, then dataset.batch can work. The behavior of returning this kind of SparseTensor is exactly aligned with how to decode tf.example tf.io.VarLenFeature,
--- a/tensorflow_io/bigquery/README.md
+++ b/tensorflow_io/bigquery/README.md
@@ -85,7 +85,7 @@ if __name__ == '__main__':
 
 ```
 
-It also supports reading repeated mode BigQuery column (each field contains array of values). In this case, selected_fields needs be a dictionary in a
+It also supports reading BigQuery column with repeated mode (each field contains array of values with primitive type: Integer, Float, Boolean, String, but RECORD is not supported). In this case, selected_fields needs be a dictionary in a
 form like
 
 ```
--- a/tensorflow_io/bigquery/README.md
+++ b/tensorflow_io/bigquery/README.md
@@ -85,6 +85,81 @@ if __name__ == '__main__':
 
 ```
 
+It also supports reading repeated mode BigQuery column (each field contains array of values). In this case, selected_fields needs be a dictionary in a
+form like
+
+```
+        { "field_a_name": {"mode": "repeated", output_type: dtypes.int64},
+          "field_b_name": {"mode": "nullable", output_type: dtypes.string},
+          ...
+          "field_x_name": {"mode": "repeated", output_type: dtypes.string}
+        }
+
+```
+"mode" is BigQuery column attribute concept, it can be 'repeated', 'nullable' or 'required'.The output field order is unrelated to the order of fields in
+selected_fields. If "mode" not specified, defaults to "nullable". If "output_type" not specified, DT_STRING is implied for all Tensors.
+
+```
+from tensorflow.python.framework import ops
+from tensorflow.python.framework import dtypes
+from tensorflow_io.bigquery import BigQueryClient
+from tensorflow_io.bigquery import BigQueryReadSession
+
+GCP_PROJECT_ID = '<FILL_ME_IN>'
+DATASET_GCP_PROJECT_ID = "bigquery-public-data"
+DATASET_ID = "certain_dataset"
+TABLE_ID = "certain_table_with_repeated_field"
+
+def main():
+  ops.enable_eager_execution()
+  client = BigQueryClient()
+  read_session = client.read_session(
+      "projects/" + GCP_PROJECT_ID,
+      DATASET_GCP_PROJECT_ID, TABLE_ID, DATASET_ID,
+      selected_fiels={
+          "field_a_name": {"mode": "repeated", output_type: dtypes.int64},
+          "field_b_name": {"mode": "nullable", output_type: dtypes.string},
+          "field_c_name": {"mode": "repeated", output_type: dtypes.string}
+        }
+      requested_streams=2,
+      row_restriction="num_characters > 1000",
+      data_format=BigQueryClient.DataFormat.AVRO)
+  dataset = read_session.parallel_read_rows()
+
+  row_index = 0
+  for row in dataset.prefetch(10):
+    print("row %d: %s" % (row_index, row))
+    row_index += 1
+
+if __name__ == '__main__':
+  app.run(main)
+```
+
+Then each field of a repeated column becomes a rank-1 variable length Tensor. If you want to 
+work that Tensor with dataset.batch, then you can use code like
+
+```
+dataset = read_session.parallel_read_rows()
+def sparse_dataset_map(features, sparse_column_names):
+  """
+  Map sparse repeated columns to tf.SparseTensor.
+  This matches how VarLenFeature is decoded from tf.Example datasets.
+  """
+  for col_name in sparse_column_names:
+    l = tf.size(features[col_name], tf.int64)
+    indices = tf.reshape(tf.range(l, dtype=tf.int64), [l, 1])
+    features[col_name] = tf.SparseTensor(indices=indices,
+                                         values=features[col_name],
+                                         dense_shape=[l])
+dataset_can_be_batched = dataset.map(lambda features: sparse_dataset_map(features, ["field_a_name", "field_b_name", "field_c_name"]))
+
+```
+to map that that as a SparseTensor first, then dataset.batch can work. The behavior of returning this kind of SparseTensor is exactly aligned with how to decode tf.example tf.io.VarLenFeature,
+which essentially just like the repeated column in BigQuery.
+
+
+
+
 Please refer to BigQuery connector Python docstrings and to
 [Enable BigQuery Storage API](https://cloud.google.com/bigquery/docs/reference/storage/rpc/)
 documentation for more details about each parameter.
--- a/tensorflow_io/bigquery/python/ops/bigquery_api.py
+++ b/tensorflow_io/bigquery/python/ops/bigquery_api.py
@@ -49,6 +49,14 @@ class BigQueryClient:
         AVRO = "AVRO"
         ARROW = "ARROW"
 
+    class FieldMode(enum.Enum):
+        """BigQuery column mode.
+        """
+
+        NULLABLE = "NULLABLE"
+        REQUIRED = "REQUIRED"
+        REPEATED = "REPEATED"
+
     def __init__(self):
         """Creates a BigQueryClient to start BigQuery read sessions.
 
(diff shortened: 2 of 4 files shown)


RuhuaJiang:
@googlebot I signed it!

vlasenkoalexey:
@RuhuaJiang thanks for your contribution, could please also add test here: https://github.com/tensorflow/io/blob/master/tests/test_bigquery_eager.py
So it is clear that everything works end-to-end as expected.

RuhuaJiang:
@vlasenkoalexey will do.

vlasenkoalexey:
Sorry for delay, I'm OOO and missed your PR.
Without a sample or test, it is hard to understand how API is going to work.
Is either selected_fields or selected_fields_repeated can be set?
What happens if both of them are set?

yongtang:
@RuhuaJiang You can add a sample test case in tests/test_bigquery_eager.py to cover your changes.

RuhuaJiang:
Thanks @vlasenkoalexey @yongtang, sample unit tests added

tests/test_bigquery_eager.py
```diff
@@ -251,6 +283,45 @@ def tearDownClass(cls):  # pylint: disable=invalid-name
...
+                "string",
+                "boolean",
+                "int",
+                "long",
+                "float",
+                "double",
+                "repeated_string",
+                "repeated_double",
+            ],
+            selected_fields_repeated=[
```
vlasenkoalexey, line 302: I don't think that such API is easy to use. Consider case when you need to select hundreds of columns, and some of them are repeated. Maintaining such code would be a problem.

Could you instead change API so if selected fields value ends with [], field is considered repeated.
So in this case it is going to look like:
            selected_fields=[
                "string",
                "boolean",
                "int",
                "long",
                "float",
                "double",
                "repeated_string[]",
                "repeated_double[]",
            ],

vlasenkoalexey: And could you also add a comment to https://github.com/tensorflow/io/blob/master/tensorflow_io/bigquery/README.md explaining how to select repeated fields.

RuhuaJiang: I agree with you.  The reason I put selected_fields_repeated in a current way, as you may already realize, is to align existing pattern of selected_fields and output_types. I have concerns around this as well, so want to make sure we agree with API first before i proceeding furthur (adding other primitive types, adding more comprehensive tests, add docs etc). 

I don't think 
```
"repeated_string[]",
"repeated_double[]",
```
is a good idea. We are using `[]` trick, which is unintuitive, to represent the semantic meaning of `is repeated`. 

Ideally,  selected_fields, output_types, selected_fields_repeated, these three should be represented as a dictionary
{
 'field': a_selected_field,  'mode': repeated_or_not,  'output_types': a_output_type,
...
}
right now these three are three lists, we rely on the index to list to reach that field:mode:output_type mapping semantic.

RuhuaJiang: the reason I didn't choose the route of making these three in the dictionary in this pull request directly primarily makes the API backwards compatible.

vlasenkoalexey: I like your proposal.
How about changing API to accept selected_fields as either array, as it is now or as a dictionary, the way you suggested?
If it is specified as a dictionary, ouput_type should not be specified.

And another question. Is it actually necessary to know whether field is repeated or not on the python side?
Would it be possible to just derive that information in C++ from AVRO schema?

RuhuaJiang: > How about changing API to accept selected_fields as either array, as it is now or as a dictionary, the way you suggested?

Sounds good. I will make the change.

> And another question. Is it actually necessary to know whether field is repeated or not on the python side?
Would it be possible to just derive that information in C++ from AVRO schema?

Good question. I gave a try a while ago, the AVRO schema from C++ do contain the information but it is a Tensor. I am not sure having a session run over there is a good idea (maybe i will give anther stab later to see how to best hide *repeated* from user). Open to suggestions :) .  Just sharing, internally we build a higher-level abstraction, say BigQueryDataSet, which calls BigQuery Python Client API to get the Avro Schema and then call this tfio API so user doesn't need to worry about setting repeated at all.

RuhuaJiang: Update: It seems I couldn't find a very elegant way at this moment to avoid treating repeated as a user input parameter for now. This repeatd_field information is used in tensorflow_io/bigquery/python/ops/bigquery_api.py to set shape for tensors in self._element_spec. 

I have updated the code to be the API as we discussed above: support both list and dict for selected_fields

tests/test_bigquery_eager.py
```diff
@@ -251,6 +320,29 @@ def tearDownClass(cls):  # pylint: disable=invalid-name
...
+            self.TABLE_ID,
+            self.DATASET_ID,
+            selected_fields={
+                "string": {"output_type": dtypes.string},
+                "boolean": {"output_type": dtypes.bool},
+                "int": {"output_type": dtypes.int32},
+                "long": {"output_type": dtypes.int64},
+                "float": {"output_type": dtypes.float32},
+                "double": {"output_type": dtypes.float64},
+                "repeated_bool": {"mode": "repeated", "output_type": dtypes.bool},
```
RuhuaJiang, line 336: leave as a todo note for myself: this test current doesn't test backwards compatibility (but I did manually rollback tests/test_bigquery_eager.py to previous version and test that selected_fields as list + output_types still pass, so the change is backwards compatible).  Let me add a test later.

cc @vlasenkoalexey

tensorflow_io/bigquery/python/ops/bigquery_api.py
```diff
@@ -115,21 +126,40 @@ def read_session(
...
+            selected_fields_repeated = [False] * len(selected_fields)
+
+        elif isinstance(selected_fields, dict):
+            _selected_fields = []
+            selected_fields_repeated = []
+            output_types = []
+            for field in selected_fields:
+                _selected_fields.append(field)
+                mode = selected_fields[field].get("mode", "nullable")
+                if mode == "repeated":
```
vlasenkoalexey, line 151: could you add a FieldMode enum similarly to DataFormat, so strings are not hardcoded.

tensorflow_io/bigquery/README.md
```diff
@@ -85,6 +85,81 @@ if __name__ == '__main__':
...
+
+```
+        { "field_a_name": {"mode": "repeated", output_type: dtypes.int64},
+          "field_b_name": {"mode": "nullable", output_type: dtypes.string},
+          ...
+          "field_x_name": {"mode": "repeated", output_type: dtypes.string}
+        }
+
+```
+"mode" is BigQuery column attribute concept, it can be 'repeated', 'nullable' or 'required'.The output field order is unrelated to the order of fields in
```
vlasenkoalexey, line 99: could you test that 'nullable' and 'required' work for DataFormat.ARROW and mention that 'repeated' only works for primitive types on leaf level fields for  DataFormat.AVRO

RuhuaJiang: > could you test that 'nullable' and 'required' work for DataFormat.ARROW

it looks like there is no existing unit test in tests/test_bigquery_eager.py for DataFormat.ARROW (https://github.com/tensorflow/io/pull/939/files and I also searched the repo).  It might not ideal to coupling adding a unit test for DataFormat.ARROW with this change, maybe better do it in a separate pull request? wdyt? 

I tried something like 

```
    def test_read_rows_nonrepeated_only_with_arrow(self):
        """Test for reading rows with non-repeated fields only, with DataFormat.ARROW, then selected_fields and output_types are list (backward compatible)."""
        client = BigQueryTestClient(BigqueryOpsTest.server.endpoint())
        read_session = self._get_read_session(client, nonrepeated_only=True, data_format=BigQueryClient.DataFormat.ARROW)

        streams_list = read_session.get_streams()
        self.assertEqual(len(streams_list), 2)
        dataset1 = read_session.read_rows(streams_list[0])
        itr1 = iter(dataset1)
        self.assertEqual(
            self._get_nonrepeated_only_fields(self.STREAM_1_ROWS[0]),
            self._normalize_dictionary(itr1.get_next()),
        )
        self.assertEqual(
            self._get_nonrepeated_only_fields(self.STREAM_1_ROWS[1]),
            self._normalize_dictionary(itr1.get_next()),
        )
        with self.assertRaises(errors.OutOfRangeError):
            itr1.get_next()

        dataset2 = read_session.read_rows(streams_list[1])
        itr2 = iter(dataset2)
        self.assertEqual(
            self._get_nonrepeated_only_fields(self.STREAM_2_ROWS[0]),
            self._normalize_dictionary(itr2.get_next()),
        )
        self.assertEqual(
            self._get_nonrepeated_only_fields(self.DEFAULT_VALUES),
            self._normalize_dictionary(itr2.get_next()),
        )
        with self.assertRaises(errors.OutOfRangeError):
            itr2.get_next()
```
and I am getting some issues

```
__________________________________________________________________________________________________________ BigqueryOpsTest.test_read_rows_nonrepeated_only_with_arrow __________________________________________________________________________________________________________

self = <tests.test_bigquery_eager.BigqueryOpsTest testMethod=test_read_rows_nonrepeated_only_with_arrow>

    def test_read_rows_nonrepeated_only_with_arrow(self):
        """Test for reading rows with non-repeated fields only, with DataFormat.ARROW, then selected_fields and output_types are list (backward compatible)."""
        client = BigQueryTestClient(BigqueryOpsTest.server.endpoint())
>       read_session = self._get_read_session(client, nonrepeated_only=True, data_format=BigQueryClient.DataFormat.ARROW)

tests/test_bigquery_eager.py:495:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
tests/test_bigquery_eager.py:356: in _get_read_session
    data_format=data_format
tensorflow_io/bigquery/python/ops/bigquery_api.py:183: in read_session
    row_restriction=row_restriction,
<string>:177: in io_big_query_read_session
    ???
/usr/local/lib/python3.6/dist-packages/tensorflow/python/framework/ops.py:6653: in raise_from_not_ok_status
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

value = None, from_value = None

>   ???
E   tensorflow.python.framework.errors_impl.InvalidArgumentError: AVRO schema is missing [Op:IO>BigQueryReadSession]

<string>:3: InvalidArgumentError
```
i think we need an ARROW schema in our test

> mention that 'repeated' only works for primitive types on leaf level fields for DataFormat.AVRO

Done

tensorflow_io/bigquery/python/ops/bigquery_api.py
RuhuaJiang: added.

tensorflow_io/bigquery/README.md
vlasenkoalexey: You are right, I didn't add ARROW test yet, need to setup test infrastructure to support ARROW, so test is not expected to work. Could you just try it on any existing BQ dataset.

RuhuaJiang: Tried 

```
from tensorflow.python.framework import ops
from tensorflow.python.framework import dtypes
from tensorflow_io.bigquery import BigQueryClient
from tensorflow_io.bigquery import BigQueryReadSession

GCP_PROJECT_ID = '<Ruhua's_project_id>'
DATASET_GCP_PROJECT_ID = "bigquery-public-data"
DATASET_ID = "samples"
TABLE_ID = "wikipedia"

def main():
  ops.enable_eager_execution()
  client = BigQueryClient()
  read_session = client.read_session(
      "projects/" + GCP_PROJECT_ID,
      DATASET_GCP_PROJECT_ID, TABLE_ID, DATASET_ID,
      selected_fields = ["title"],
      output_types = [dtypes.string],
      requested_streams=2,
      row_restriction="num_characters > 1000",
      data_format=BigQueryClient.DataFormat.ARROW)
  dataset = read_session.parallel_read_rows()

  row_index = 0
  for row in dataset.prefetch(10):
    print("row %d: %s" % (row_index, row))
    row_index += 1

if __name__ == '__main__':
  main()
```
with master branch by following https://github.com/tensorflow/io

```
TFIO_DATAPATH=bazel-bin python3 -m pip install .
TFIO_DATAPATH=bazel-bin python3  <file contains above_code>.py
```
I was getting 

```
2020-07-07 16:26:28.394491: I tensorflow/compiler/xla/service/service.cc:176]   StreamExecutor device (0): Host, Default Version
  File "/usr/local/lib/python3.6/dist-packages/tensorflow/python/eager/context.py", line 1986, in execution_mode
    yield
  File "/usr/local/lib/python3.6/dist-packages/tensorflow/python/data/ops/iterator_ops.py", line 655, in _next_internal
  File "/usr/local/lib/python3.6/dist-packages/tensorflow/python/ops/gen_dataset_ops.py", line 2363, in iterator_get_next
  File "/usr/local/lib/python3.6/dist-packages/tensorflow/python/framework/ops.py", line 6653, in raise_from_not_ok_status
tensorflow.python.framework.errors_impl.InternalError: Invalid: Invalid argument: arrow data type 0x7f1a48016fa0 is not supported: Type error: Arrow data type is not supported [Op:IteratorGetNext]

During handling of the above exception, another exception occurred:

  File "test_bigquery_eager.py", line 66, in <module>
    main()
  File "test_bigquery_eager.py", line 61, in main
    for row in dataset.prefetch(10):
  File "/usr/local/lib/python3.6/dist-packages/tensorflow/python/data/ops/iterator_ops.py", line 631, in __next__
    return self.next()
  File "/usr/local/lib/python3.6/dist-packages/tensorflow/python/data/ops/iterator_ops.py", line 670, in next
    return self._next_internal()
  File "/usr/local/lib/python3.6/dist-packages/tensorflow/python/data/ops/iterator_ops.py", line 661, in _next_internal
  File "/usr/lib/python3.6/contextlib.py", line 99, in __exit__
  File "/usr/local/lib/python3.6/dist-packages/tensorflow/python/eager/context.py", line 1989, in execution_mode
    executor_new.wait()
  File "/usr/local/lib/python3.6/dist-packages/tensorflow/python/eager/executor.py", line 67, in wait
    pywrap_tfe.TFE_ExecutorWaitForAllPendingNodes(self._handle)
tensorflow.python.framework.errors_impl.InternalError: Invalid: Invalid argument: arrow data type 0x7f1a48016fa0 is not supported: Type error: Arrow data type is not supported
```

RuhuaJiang: ah, nvm, i think this was due to arrow can't work with dtypes.string

replace 
```
      selected_fields = ["title"],
      output_types = [dtypes.string],
```
with 
```
      selected_fields = ["id"],
      output_types = [dtypes.int64],
```
things works perfectly fine with both master branch and my branch.

Also, as you suggested, test nullable and required with arrow in my branch

```
from tensorflow.python.framework import ops
from tensorflow.python.framework import dtypes
from tensorflow_io.bigquery import BigQueryClient
from tensorflow_io.bigquery import BigQueryReadSession

GCP_PROJECT_ID = 'twttr-cortex-mlengine-evl'
DATASET_GCP_PROJECT_ID = "bigquery-public-data"
DATASET_ID = "samples"
TABLE_ID = "wikipedia"

def main():
  ops.enable_eager_execution()
  client = BigQueryClient()
  read_session = client.read_session(
      "projects/" + GCP_PROJECT_ID,
      DATASET_GCP_PROJECT_ID, TABLE_ID, DATASET_ID,
      selected_fields = {
          "id": {"mode": BigQueryClient.FieldMode.NULLABLE, "output_type": dtypes.int64},
          "wp_namespace": {"mode": BigQueryClient.FieldMode.REQUIRED, "output_type": dtypes.int64},
          },
      requested_streams=2,
      row_restriction="num_characters > 1000",
      data_format=BigQueryClient.DataFormat.ARROW)
  dataset = read_session.parallel_read_rows()

  row_index = 0
  for row in dataset.prefetch(10):
    print("row %d: %s" % (row_index, row))
    row_index += 1

if __name__ == '__main__':
  main()
```
works great!

```
row 3829: OrderedDict([('id', <tf.Tensor: shape=(), dtype=int64, numpy=7863732>), ('wp_namespace', <tf.Tensor: shape=(), dtype=int64, numpy=0>)])
row 3830: OrderedDict([('id', <tf.Tensor: shape=(), dtype=int64, numpy=9317673>), ('wp_namespace', <tf.Tensor: shape=(), dtype=int64, numpy=0>)])
row 3831: OrderedDict([('id', <tf.Tensor: shape=(), dtype=int64, numpy=5986>), ('wp_namespace', <tf.Tensor: shape=(), dtype=int64, numpy=0>)])
row 3832: OrderedDict([('id', <tf.Tensor: shape=(), dtype=int64, numpy=15843197>), ('wp_namespace', <tf.Tensor: shape=(), dtype=int64, numpy=0>)])
```

vlasenkoalexey: Perfect, thanks for your contribution!!!

PR #1000, opened 2020-06-17
Add decoding and encoding of mu law for audio

yongtang:
This PR adds implementations of mu-law companding, `decode_mulaw`/`encode_mulaw`,
similiar to torchaudio's mu_law_encoding/mu_law_decoding.
mu-law is commonly used (together with a-law) in audio processing so they
could be useful addition to tfio.

At the moment the apis are exposed in tfio.experimental.audio.[decode_mulaw/encode_mulaw].
We could promote to tfio.audio after APIs stablizes.

This PR is part of #839.


yongtang:
G.711 is a special case that use invert (one's complement) for encoding. So a generic mu law algorithm does not necessarily translate to G.711 mulaw encoding directly. We will need two set of functions: mulaw_encode/decode (G.711), and mulaw_function

Will re-open once figure out the best representation of function names.

PR #1003, opened 2020-06-22
Add audio operations to trim/split noise and remix audio signals.

yongtang:
This PR adds several audio effect operations that might be useful:
- `tfio.experimental.audio.trim` removes beginning and trailing noise (keep if larger than e)
- `tfio.experimental.audio.split` returns the `start, stop` indices of audio segment that are not noise (later than e).
- `tfio.experimental.audio.remix` remix audio segments based on the original audio input and an indices matrix (could be from `trim` or `split` above.

The operations are more or less inspired from `librosa.effects.[remix|trim|split]`.
Though this PR allows multiple dimensional support so that it is possible to operate
on batches or vectorize.

This PR mostly utilize python for implementation, except for `OrderIndices` C++ kernel op
which translate an index to a value in one dimension out of n-D.

Some of the python implemtations may not be very performance, though the immediate goal is to add support first. Improvement could be done in the future (or convert to C++ kernels if needed).

While the above ops are placed under `tfio.experimental.audio`, they are really generic
ops that could be useful in other places (e.g, used for video to trim, split, remix clips).
It might be worth to consider placing them under `tfio.experimental.ops`?

Another potential enhancement is to add `trim/split/remix` support for `AudioIOTensor`,
which will be really useful if audio is very large (as `AudioIOTensor` is lazy loaded).


tests/test_audio_ops_eager.py
```diff
@@ -887,3 +887,64 @@ def test_fade():
...
+        (v, 1, e, i, k),
+        (
+            tf.reshape(v, [4, 120, 1]),
+            1,
+            tf.reshape(e, [4, 2, 1]),
+            tf.reshape(i, [4, 3, 2, 1]),
+            tf.reshape(k, [4, 90, 1]),
+        ),
+    ]:
+        indices = tfio.experimental.audio.trim(value, axis=axis, epsilon=4.95)
```
terrytangyuan, line 943: Move epsilon and potentially other re-appearing numbers to reusable constants?

yongtang:
@terrytangyuan Updated. Thanks.

PR #1008, opened 2020-06-24
replace tf.logging with tf.get_logger()

burgerkingeater:
replace tf.logging with tf.get_logger() as logging is no longer in tensorflow module


yongtang:
@burgerkingeater From the API and the code:

https://github.com/tensorflow/tensorflow/blob/v2.2.0/tensorflow/python/platform/tf_logging.py#L176-L178

It looks like in TF 2.0 the logging has been changed to use:
```
tf.get_logger().warning(msg, *args, **kwargs)
```

Since `get_logger` is a public 2.0 API, maybe we can use `tf.get_logger().warning` instead?

burgerkingeater:
@burgerkingeater  thanks, updated.

PR #1032, opened 2020-07-11
Add null handling by consuming default

burgerkingeater:
Why?
Dropping values when a null is encountered does not work for dense tensors which makes them variable length in a batch
What? 
Need to consume default and check types and put them in place of the null.

That is also the expected user behavior.

This change is made by @fraudies in our internal fork and backported here.


burgerkingeater:
@yongtang can i get a review? thanks

PR #1034, opened 2020-07-13
Fix parse_avro docstring so that it can be shown in API doc

burgerkingeater:


tensorflow_io/core/python/experimental/parse_avro_ops.py
```diff
@@ -32,11 +32,14 @@ def parse_avro(serialized, reader_schema, features, avro_names=None, name=None):
...
 
     :param serialized: The batched, serialized string tensors.
+
     :param reader_schema: The reader schema. Note, this MUST match the reader
                           schema from the avro_record_dataset. Otherwise,
                           this op will segfault!
+
     :param features: A map of feature names mapped to feature
                      information.
+
```
yongtang, line 42: @burgerkingeater I think that may not work to fix the argument display. Can you follow the way other APIs exposing the arg in docstring, e.g., :
```
    Args:
    serialized: A vector (1-D Tensor) of strings, a batch of binary
      serialized `Example` protos.
    reader_schema: A scalar string representing the reader schema.
```

yongtang:
@burgerkingeater Any update on this PR?

burgerkingeater:
@yongtang thanks for the reminder, will get to it today.

burgerkingeater:
@yongtang updated, thanks

tensorflow_io/core/python/experimental/parse_avro_ops.py
burgerkingeater: updated, thanks

PR #1044, opened 2020-07-20
Add GIF encoding support as tfio.image.encode_gif

yongtang:
This PR adds GIF encoding support as tfio.image.encode_gif.
While GIF is not a good fit for high resolution images,
GIF is most useful for storing animated series images.
This PR add encode_gif support so that it is possible to
construct a animated (movie like) gif from a collection of images.

This PR only add input of `[N, H, W, C]` support. Other shapes
such as `[H, W, C]` will be added in follow up PRs.


wookayin:
This looks great, awesome! Just out of curiosity, any idea how compact the encoded GIFs would be? GIFs created without a careful optimization often have a very big file size, so I wonder this is almost ready to use out-of-box.

yongtang:
@wookayin The scope of the GIF optimization could be large. In this PR some of the optimizations are not in. Would you mind to open a new issue about GIF optimization so that we can track the progress there?

PR #1047, opened 2020-07-21
Fix failing build on master due to TF C++ API changes

pshved:
I've been building against TF nightly, and my repo wasn't building due to some TF changes.  Attached are the fixes with the references to the breaking TF commits.

I've also just signed the Google Contributor Agreement.

The TF version I was building against is `2.4.0.dev20200718`

### DIFF
--- a/tensorflow_io/arrow/kernels/arrow_dataset_ops.cc
+++ b/tensorflow_io/arrow/kernels/arrow_dataset_ops.cc
@@ -18,6 +18,7 @@ limitations under the License.
 #include "arrow/util/io_util.h"
 #include "tensorflow/core/framework/dataset.h"
 #include "tensorflow/core/graph/graph.h"
+#include "tensorflow/core/public/version.h"
 #include "tensorflow_io/arrow/kernels/arrow_kernels.h"
 #include "tensorflow_io/arrow/kernels/arrow_stream_client.h"
 #include "tensorflow_io/arrow/kernels/arrow_util.h"
@@ -260,6 +261,10 @@ class ArrowDatasetBase : public DatasetBase {
       switch (element.dtype()) {
         TF_CALL_ALL_TYPES(HANDLE_TYPE);
         TF_CALL_QUANTIZED_TYPES(HANDLE_TYPE);
+#if TF_MAJOR_VERSION == 2 && TF_MINOR_VERSION == 3
+        TF_CALL_uint32(HANDLE_TYPE);
+        TF_CALL_uint64(HANDLE_TYPE);
+#endif
 #undef HANDLE_TYPE
         default:
           return errors::Unimplemented(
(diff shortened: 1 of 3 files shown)


yongtang:
@pshved Thanks for the PR. There are test failures (see https://github.com/tensorflow/io/pull/1047/checks?check_run_id=892500380)
```
self = <tensorflow.python.eager.executor.Executor object at 0x7fc4c7b7e588>
1745

1746
    def wait(self):
1747
      """Waits for ops dispatched in this executor to finish."""
1748
>     pywrap_tfe.TFE_ExecutorWaitForAllPendingNodes(self._handle)
1749
E     tensorflow.python.framework.errors_impl.UnimplementedError: CopyElementsToParent Unhandled data type: 22
1750

```

My understanding is that the failures are caused by the removal of:
```diff
-        TF_CALL_uint32(HANDLE_TYPE);		
-        TF_CALL_uint64(HANDLE_TYPE);
```

Can you add them back and give it a try?

pshved:
@yongtang that means the TF version used in the tests is not recent enough for the code to start failing.  I had to remove these lines because the build was failing with the following message:

```
tensorflow_io/arrow/kernels/arrow_dataset_ops.cc:256:3: error: duplicate case value                                                                                                               
   case DataTypeToEnum<T>::value: {                                       \                                                                                                                       
   ^                                                                                                                                                                                              
tensorflow_io/arrow/kernels/arrow_dataset_ops.cc:256:3: note: in definition of macro 'HANDLE_TYPE'                                                                                                
   case DataTypeToEnum<T>::value: {                                       \                                                                                                                   
   ^~~~ 
```

I think it's because my local TF did contain TF commit tensorflow/tensorflow@51ee632 but your build environment doesn't.  But soon enough, the build will start failing.  What would you suggest?  Some macro workaround that compiles both with the old and the new version of TF?

yongtang:
Thanks @pshved, I think macro will help in this case. I tried the following patch on top of your PR. It looks like it works for both 2.3.0rc0 and tf-nightly.


```diff
diff --git a/tensorflow_io/arrow/kernels/arrow_dataset_ops.cc b/tensorflow_io/arrow/kernels/arrow_dataset_ops.cc
index 8af66bf..3b4e419 100644
--- a/tensorflow_io/arrow/kernels/arrow_dataset_ops.cc
+++ b/tensorflow_io/arrow/kernels/arrow_dataset_ops.cc
@@ -18,6 +18,7 @@ limitations under the License.
 #include "arrow/util/io_util.h"
 #include "tensorflow/core/framework/dataset.h"
 #include "tensorflow/core/graph/graph.h"
+#include "tensorflow/core/public/version.h"
 #include "tensorflow_io/arrow/kernels/arrow_kernels.h"
 #include "tensorflow_io/arrow/kernels/arrow_stream_client.h"
 #include "tensorflow_io/arrow/kernels/arrow_util.h"
@@ -260,6 +261,10 @@ class ArrowDatasetBase : public DatasetBase {
       switch (element.dtype()) {
         TF_CALL_ALL_TYPES(HANDLE_TYPE);
         TF_CALL_QUANTIZED_TYPES(HANDLE_TYPE);
+#if TF_MAJOR_VERSION == 2 && TF_MINOR_VERSION == 3
+        TF_CALL_uint32(HANDLE_TYPE);
+        TF_CALL_uint64(HANDLE_TYPE);
+#endif
 #undef HANDLE_TYPE
         default:
           return errors::Unimplemented(
```

We can remove the macro later once TF 2.3.0 final is released.

PR #1063, opened 2020-07-30
BugFix: KafkaGroupIODataset retrieving messages from non-committed offsets.

kvignesh1420:
This PR addresses a bug while consuming messages from kafka topics using `tfio.experimental.streaming.KafkaGroupIODataset` . and adds two basic test cases.

**BugFix:**

**Current Scenario**: When a new consumer group is created, there are no committed offsets that are being maintained. In such cases, the dataset fails to read messages.

**Fix**: Before starting to retrieve the messages, I am checking for already committed offsets and based on the retrieved info, defaulting to the beginning offset if I don't find any. Setting it to latest offset will not return any messages. However, I will consider both the options when I implement a different variant of the dataset. (One where we can wait for the new messages to come in.)

**NOTE**: This is due to the following reasons:

1. Registering a rebalance_cb turns off librdkafka's automatic partition assignment/revocation and instead delegates that responsibility to the application's rebalance_cb.
2. Without a rebalance callback this is done automatically by librdkafka but registering a rebalance callback gives the application flexibility in performing other operations along with the assinging/revocation, such as fetching offsets from an alternate location (on assign) or manually committing offsets (on revoke).

**Reference**: https://docs.confluent.io/3.3.1/clients/librdkafka/classRdKafka_1_1RebalanceCb.html#a490a91c52724382a72380af621958741

**Tests:**

- [x] Start a new consumer by joining a new consumer group and read all the messages.
- [x] Start a new consumer by joining an existing consumer group and find no messages due to 0 lag.

### DIFF
--- a/tensorflow_io/core/kernels/kafka_kernels.cc
+++ b/tensorflow_io/core/kernels/kafka_kernels.cc
@@ -772,8 +772,15 @@ class KafkaRebalanceCb : public RdKafka::RebalanceCb {
   void rebalance_cb(RdKafka::KafkaConsumer* consumer, RdKafka::ErrorCode err,
                     std::vector<RdKafka::TopicPartition*>& partitions) {
     LOG(ERROR) << "REBALANCE: " << RdKafka::err2str(err);
+    LOG(ERROR) << "Retrieved committed offsets with status code: "
+               << consumer->committed(partitions, 5000);
+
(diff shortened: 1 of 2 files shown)


kvignesh1420:
@yongtang please review this PR before the release. Was able to resolve a minor bug while writing tests.

PR #1065, opened 2020-08-03
Additional tests for KafkaGroupIODataset.

kvignesh1420:
This PR adds test cases for testing the functionality of `KafkaGroupIODataset`. The following scenarios have been added to the existing collection:

- [x] Add new messages to the topic and resume the consumption of messages from stored offsets.
- [x] Fetch all messages from the beginning when a secondary consumer group is created. (initial + newly added messages)
- [x] Fetch initial messages from a new topic using the primary consumer group.
- [x] Add new messages to the new topic and resume the consumption of messages from stored offsets using the primary consumer group.
- [x] Create a tertiary consumer group and read from both the topics at once from the beginning.


kvignesh1420:
@yongtang , please review. Thanks.

Issue #947, opened 2020-05-07
[kafka-io] - KafkaStreamIODataset eof and timeout behavior


nmatare:
The first implementation of KafkaDataset (kafka_kernels_deprecated.cc) allowed the user to specify `eof` and `timeout` which blocked the calling thread if no data was present. The new `StreamIODataset` Kafka dataset ("A StreamIODataset where data generation may last forever and/or non-repeatable") immediately exits if there isn't any data, and doesn't allow for a timeout parameter. 

The first behavior allowed us to continuously poll from the Kafka cluster until data arrived to action the model. This prevented us from having to rebuild the iterator if there was a momentary lapse in arriving data. (Unfortunately, `timeout` didn't work in that case) Would it make sense to combine the behaviors?

```sh
bash -x -e tests/test_kafka/kafka_test.sh start kafka
```

The StreamIODataset case:
```python
import tensorflow_io as tfio
import tensorflow_io.kafka as kafka_io
from kafka import KafkaProducer

GROUP_ID = 'test_group'
SERVERS = 'localhost:9092'
TOPIC = 'testing_topic'
PARTITION = 0

producer = KafkaProducer(bootstrap_servers=[SERVERS])

for letter in ['A', 'B', 'C', 'D', 'E']:
  producer.send(TOPIC, letter.encode('utf-8'))

# A StreamIODataset where data generation may last forever and/or non-repeatable.
stream_ds = tfio.IODataset.stream().from_kafka(
    topic=TOPIC,
    offset=0,
    partition=PARTITION,
    stop=-1,
    servers=SERVERS,
    configuration=['auto.offset.reset=earliest', 'enable.partition.eof=false'],
)

for data in stream_ds:
  print(data.message)

```
The above immediately exits with no EOF parameter. We have to rebuild the entire iterator to start reading data again...
https://github.com/tensorflow/io/blob/f71ef1d3b56d665561fe8b10e95ca32ed46870e5/tensorflow_io/core/kernels/kafka_kernels.cc#L201

The KafkaDataset case:
```python
deprecated_ds = kafka_io.KafkaDataset(
    topics=TOPIC,
    servers=SERVERS,
    group=GROUP_ID,
    config_global=['auto.offset.reset=earliest'],
    message_key=False,
    eof=False,   # set here to block
    timeout=10,
)

# WARNING this will forever block, (timeout has no effect here)
for data in deprecated_ds:
  print(data)
```

kvignesh1420:
[...]

PR #1068, opened 2020-08-04
Added support for stream timeout in KafkaGroupIODataset

kvignesh1420:
This PR addresses the issue https://github.com/tensorflow/io/issues/947 by adding the following functionality:

- Allows users to pass a value for the `stream_timeout` argument in `KafkaGroupIODataset` which prevents the users from rebuilding the generator again when the consumer is waiting for new messages.
- Allows users to pass a value for the `message_timeout` argument in `KafkaGroupIODataset` when the default value might be less/more than needed, based on the cluster connectivity.
- Relevant tests for the wait functionality (can be extended).

**NOTE/INFO:** The functionality of `message_timeout` in Kafka works in such a way that, when a consumer is trying to fetch a new message with a timeout of let's say 5000 milliseconds and it doesn't find the message, then the consumer will always wait for 5000 milliseconds even though a new message was available to be fetched after let's say 2000 milliseconds.

### DIFF
--- a/tensorflow_io/core/python/experimental/kafka_group_io_dataset_ops.py
+++ b/tensorflow_io/core/python/experimental/kafka_group_io_dataset_ops.py
@@ -59,7 +59,7 @@ class KafkaGroupIODataset(tf.data.Dataset):
           servers: An optional list of bootstrap servers.
             For example: `localhost:9092`.
           message_timeout: An optional timeout value (in milliseconds) for retrieving messages
-            from kafka. Default value in 5000.
+            from kafka. Default value is 5000.
           stream_timeout: An optional timeout value (in milliseconds) to wait for the new messages
             from kafka to be retrieved by the consumers. Default value in 5000.
             NOTE: The stream_timeout value should always be greater than or equal to the message_timeout
--- a/tensorflow_io/core/python/experimental/kafka_group_io_dataset_ops.py
+++ b/tensorflow_io/core/python/experimental/kafka_group_io_dataset_ops.py
@@ -62,7 +62,7 @@ class KafkaGroupIODataset(tf.data.Dataset):
             from kafka. Default value is 5000.
           stream_timeout: An optional timeout value (in milliseconds) to wait for the new messages
             from kafka to be retrieved by the consumers. Default value is 5000.
-            NOTE: The stream_timeout value should always be greater than or equal to the message_timeout
+            NOTE: The `stream_timeout` value should always be greater than or equal to the `message_timeout`.
             value.
           configuration: An optional `tf.string` tensor containing
             configurations in [Key=Value] format.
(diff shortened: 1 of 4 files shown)


kvignesh1420:
@yongtang , instead of a `Status StreamNext` that I was planning to implement, this timeout behaviour within the existing `Status Next` seemed to address the issue. Please let me know your opinion. Thanks.

tensorflow_io/core/python/experimental/kafka_group_io_dataset_ops.py
```diff
@@ -49,6 +58,12 @@ def __init__(self, topics, group_id, servers, configuration, internal=True):
           group_id: The id of the consumer group. For example: cgstream
           servers: An optional list of bootstrap servers.
             For example: `localhost:9092`.
+          message_timeout: An optional timeout value (in milliseconds) for retrieving messages
+            from kafka. Default value in 5000.
```
terrytangyuan, line 62:
```suggestion
            from kafka. Default value is 5000.
```

tensorflow_io/core/python/experimental/kafka_group_io_dataset_ops.py
```diff
@@ -49,6 +58,12 @@ def __init__(self, topics, group_id, servers, configuration, internal=True):
...
+          stream_timeout: An optional timeout value (in milliseconds) to wait for the new messages
+            from kafka to be retrieved by the consumers. Default value in 5000.
```
terrytangyuan, line 64:
```suggestion
            from kafka to be retrieved by the consumers. Default value is 5000.
```

tensorflow_io/core/python/experimental/kafka_group_io_dataset_ops.py
```diff
@@ -49,6 +58,12 @@ def __init__(self, topics, group_id, servers, configuration, internal=True):
...
+            NOTE: The stream_timeout value should always be greater than or equal to the message_timeout
```
terrytangyuan, line 65:
```suggestion
            NOTE: The `stream_timeout` value should always be greater than or equal to the `message_timeout`.
```

kvignesh1420:
@yongtang @terrytangyuan thanks for the review. Now that we have this experimental feature, I was planning to write a tutorial around this API so that we get the feedback from the community and possibly move it from experimental sub-package to the main ops sub-package. We can highlight the key features of the API: 

- consumer group based message consumption using offsets.
- `stream_timeout` functionality for preventing the recreation of a new generator to handle intermittent data slumps.

However, since the `stream_timout` was added now, I can write prepare a notebook for the first feature as a part of our tutorials. What do you guys suggest?

yongtang:
@kvignesh1420 That would be fantastic 👍 ! We don't have tutorial for kafka yet so any tutorial will be greatly welcomed.

kvignesh1420:
@yongtang great! Will raise PR's for the same.

PR #1070, opened 2020-08-05
Clean up: Remove TimeoutConfig and reorder includes.

marioecd:
Removing unneeded ref to TimeoutConfigs and reordering includes alphabetically.


yongtang:
@marioecd Can you run clang lint to reformat tensorflow_io/core/kernels/gstpufs/memcached_file_block_cache.h with the following:
```
bazel run //tools/lint:lint -- clang -- tensorflow_io/core/kernels/gstpufs/memcached_file_block_cache.h
```

I think the header file include order might be causing linting errors.

PR #1075, opened 2020-08-06
Rename gstpufs to gsmemcachedfs

marioecd:
The memcached based distributed cache client in gstpufs is not specific to the tpu platform and thus should be named more generally to gsmemcachedfs.


marioecd:
Ok, I fixed the lint errors. The latest lint error is confusing and I am not sure it points to an issue in this PR.

Also when I run lint locally I get no errors.
bazel run -s --verbose_failures //tools/lint:check -- bazel pyupgrade black clang

marioecd:
Lint is fixed.

yongtang:
@marioecd As gsmemcachedfs has been merged, I am wondering if you are interested in adding a tutorial? We can push the tutorial to tensorflow doc site like (https://www.tensorflow.org/io/tutorials/bigquery).

PR #1080, opened 2020-08-08
Call Finish on reader when stream is ended

emkornfield:
This should be used to confirm there are no errors.


emkornfield:
This might need some more work, not sure if gRPC status is compatible with TF status.

emkornfield:
Mac CI failures looks like an CI issue (failed to download something).

yongtang:
ping @vlasenkoalexey would you mind to also take a look at this PR?

PR #1086, opened 2020-08-13
BigQuery reader optimizations

vlasenkoalexey:
- Doing field type validations only for first row since types won't change
- Initializing avro::GenericDatum only once
- Using separate gRPC connections per reader stream

### DIFF
--- a/tensorflow_io/bigquery/kernels/bigquery_lib.h
+++ b/tensorflow_io/bigquery/kernels/bigquery_lib.h
@@ -52,12 +52,13 @@ Status GetDataFormat(string data_format_str,
 class BigQueryClientResource : public ResourceBase {
  public:
   explicit BigQueryClientResource(
-      std::function<std::unique_ptr<apiv1beta1::BigQueryStorage::Stub>()>
+      std::function<std::unique_ptr<apiv1beta1::BigQueryStorage::Stub>(
+          const string &read_stream)>
           stub_factory)
       : stub_factory_(stub_factory) {}
 
   explicit BigQueryClientResource()
-      : BigQueryClientResource([]() {
+      : BigQueryClientResource([](const string &read_stream) {
           string server_name = "dns:///bigquerystorage.googleapis.com";
           auto creds = ::grpc::GoogleDefaultCredentials();
           grpc::ChannelArguments args;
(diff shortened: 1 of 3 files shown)


tensorflow_io/bigquery/kernels/bigquery_lib.h
```diff
@@ -50,17 +52,42 @@ Status GetDataFormat(string data_format_str,
...
-    return stub_;
+      std::function<std::unique_ptr<apiv1beta1::BigQueryStorage::Stub>()>
+          stub_factory)
+      : stub_factory_(stub_factory) {}
+
+  explicit BigQueryClientResource()
+      : BigQueryClientResource([]() {
+          string server_name = "dns:///bigquerystorage.googleapis.com";
+          auto creds = ::grpc::GoogleDefaultCredentials();
+          grpc::ChannelArguments args;
```
emkornfield, line 63: i think grpc can sometimes under the cover reuse channels with the same paremeters.  For parallelization you might try adding in a random arg to see if this can improve throughput.

vlasenkoalexey: Thanks for the pointer, double-checked BigTable SDK, looks like it is doing just that: https://github.com/googleapis/google-cloud-cpp/blob/b26c708de834acda41194b241f6912f02c0fd56f/google/cloud/bigtable/internal/common_client.cc#L31
Also unfortunately this didn't improve my benchmark numbers.

vlasenkoalexey:
This partially addresses #1066 
With these changes I see ~60-70% improvement for reads with 2048 batch size and 120 feature columns (original 10000 records/sec vs updated 17000 records/sec) on a benchmark running on my dev machine.

tensorflow_io/bigquery/kernels/bigquery_lib.h
```diff
@@ -50,17 +52,46 @@ Status GetDataFormat(string data_format_str,
...
+          args.SetMaxReceiveMessageSize(kMaxReceiveMessageSize);
+          args.SetUserAgentPrefix(
+              strings::StrCat("tensorflow-", TF_VERSION_STRING));
+          args.SetInt(GRPC_ARG_KEEPALIVE_PERMIT_WITHOUT_CALLS, 0);
+          args.SetInt(GRPC_ARG_KEEPALIVE_TIMEOUT_MS, 60 * 1000);
+          // To prevent gRPC from reusing channel
+          args.SetString("read_stream", read_stream);
```
emkornfield, line 71: you might want to hash % some_number here in case the number of streams returned is large.

vlasenkoalexey: read_stream uniquely identifies BQ Storage API stream and # of streams can't be too large, usually it is less than 10.
Besides if those custom args are persisted somewhere in gRPC logs, I'd rather have it as a stream name instead of some integer

Issue #1087, opened 2020-08-14
Discussion: Official docker image for tensorflow-io


kvignesh1420:
As new features are being added to the library, it would be great if we can have an official Docker image for ease of use and integration.  There was a discussion going on in https://github.com/tensorflow/io/issues/414 for integration with tf-serving, but kind of went stale.

We already have our development image but that is not an efficient one. I think it will be a good work item for the upcoming releases. Please let me know your opinions.

cc: @yongtang @terrytangyuan @BryanCutler @vlasenkoalexey

yongtang:
@kvignesh1420 That would be great! Some thing to consider, that tensorflow-io only depends on tensorflow and there is no other dependencies. We could just build a `minimal` image with tensorflow+tensorflow-io only with some lightweight base image.

Another thing to consider is to automatically push build to dockerhub (could be combined with GitHub Actions I guess).

kvignesh1420:
@yongtang, can we reach out to the tf folks to create a repository named `tensorflow/io` in docker hub? We can tag our images as follows:

- tensorflow/io:latest
- tensorflow/io:latest-devel

GPU supported images can be added later on.

yongtang:
Thanks @kvignesh1420. Also cc @seanpmorgan and @WindQAQ in case tensorflow-addons also want to publish docker images to tensorflow repo on GitHub.

seanpmorgan:
Please let us know how this goes. We decided to just create our own namespace since it seemed like less bureaucracy but would be happy to align:
https://hub.docker.com/repository/docker/tfaddons/dev_container

kvignesh1420:
@seanpmorgan, few question regarding the hub:

1. Who maintains this docker repository in the hub?
2. Do you plan to switch to the main TensorFlow repo in the hub?
3. What is the process that you are following to push the images?

kvignesh1420:
Thanks for the info @seanpmorgan. 

@yongtang what do you think about this?

yongtang:
@kvignesh1420 I think in the short term, we can start with a docker account (such as `tfio`) and build our containers with an automatic pipeline in GitHub Actions. Once the container stabilizes we can work with tensorflow team to see if the containers could be moved under tensorflow account gradually.

kvignesh1420:
@yongtang sounds good. If you can please create an account and store the creds as Github secrets, I can work on the actions and see how it turns out.

yongtang:
[...]

PR #1090, opened 2020-08-17
added 'import' tests and github actions for docker images

kvignesh1420:
This PR addresses the issue: https://github.com/tensorflow/io/issues/1087 and is a follow up of PR: https://github.com/tensorflow/io/pull/1089

Github actions have been added to test the docker build and package import functionality on the Linux and MacOS platforms.

The tests include the following:

- Build `tfio-cpu` image using `cpu.Dockerfile` and validate the tfio import in python.
- Build `tfio-dev` image using `devel.Dockerfile`, build tfio from source using bazel and validate it's import in python.

### DIFF
--- a/.github/workflows/build.space.sh
+++ b/.github/workflows/build.space.sh
@@ -1,9 +1,9 @@
 # Free disk space on Linux
-        #sudo swapoff /swapfile
-        #sudo rm -rf /swapfile
-        sudo rm -rf /usr/share/dotnet /usr/local/lib/android /opt/ghc
-        sudo apt-get remove php* ruby-* subversion mongodb-org -yq >/dev/null 2>&1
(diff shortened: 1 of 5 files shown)


kvignesh1420:
With the windows setup, was facing the following issue:

`no matching manifest for windows/amd64 10.0.17763 in the manifest list entries`

kvignesh1420:
The docker setup on macOS was posing issues with docker service discovery.

kvignesh1420:
cc: @yongtang for review

yongtang:
@kvignesh1420 The PR looks good! I think we might just simply remove the macOS build for docker, as macOS on GitHub Actions has not been super stable?

We only need to validate that docker container works (user will most likely use it on Linux). So removing macOS docker build on GitHub Actions should be fine.

PR #1092, opened 2020-08-18
Basic Arrow string type support for BQ reader

vlasenkoalexey:
- Fixed BQ connector Arrow implementation
- Added basic support for Arrow strings
- Will do additional benchmarking to see if it helps with #1066


vlasenkoalexey:
@BryanCutler, thanks for reply. Indeed this is the most basic string support for reading datasets. I realized that bunch of other code including MakeArrayData has to be updated for full support. It would also be nice to support nulls.

PR #1094, opened 2020-08-19
Update bigquery_lib.h

emkornfield:
Disable max message size checks for the BQ read API.  this seems to be consistent with the python client: https://github.com/googleapis/python-bigquery-storage/blob/645e65d3533cb445b542a54f4552e3c05a5b1382/google/cloud/bigquery_storage_v1beta2/gapic/transports/big_query_read_grpc_transport.py

### DIFF
--- a/tensorflow_io/bigquery/kernels/bigquery_kernels.cc
+++ b/tensorflow_io/bigquery/kernels/bigquery_kernels.cc
@@ -37,7 +37,6 @@ namespace tensorflow {
 namespace {
 
 namespace apiv1beta1 = ::google::cloud::bigquery::storage::v1beta1;
-static constexpr int kMaxReceiveMessageSize = 1 << 24;  // 16 MBytes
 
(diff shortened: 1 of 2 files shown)


vlasenkoalexey:
Run few benchmarks, setting kMaxReceiveMessageSize = -1 or removing kMaxReceiveMessageSize didn't make any difference on throughput. 
Would you mind removing kMaxReceiveMessageSize altogether? There is one more unused declaration at bigquery_kernels.cc

emkornfield:
Removing is not a good idea.  I believe the default is relatively small and would cause some messages to be undeliverable.

vlasenkoalexey:
LGTM, will merge once tests are fixed

yongtang:
The testing error seems to be one-time GitHub Actions issue. Believe it will be fine once merged.

I will try to merge (if issue persists will revert back).

Issue #1093, opened 2020-08-18
Add support for elasticsearch


kvignesh1420:
Since many organizations use the ELK stack for data analysis, I think it would be beneficial for many users if they can connect to ElasticSearch and train their ML models.

cc: @yongtang @terrytangyuan @BryanCutler @vlasenkoalexey

yongtang:
@kvignesh1420 This will be a great addition!. ELK stack is used widely in data analysis. Combining it with tfio => tf.data => tf.keras will really helps pushing the usage of ML to wider audiences.

kvignesh1420:
Will explore the integration.

kvignesh1420:
@yongtang  Since the connection with Elasticsearch is HTTP based, which approach would be the fastest based on previous client implementations? For example, the Prometheus kernel uses the external `prometheus/client_golang` package for connecting to the server. Do we want to follow the same style and use the [go-elasticsearch](https://github.com/elastic/go-elasticsearch) third_party package by elastic ? or do we write the connectors from scratch using libcurl (using TensorFlow wrapper `HttpRequest`)?

yongtang:
@kvignesh1420 We can start with anyone and incrementally improve later.

In prometheus, the reason to choose golang was that, prometheus by itself is a golang implementation, and golang has the first class citizen support. There is also no official C or C++ client support.

However, during the implementation we noticed that, golang does not play really well with C/C++ when embedding into one shared object. See #492 for more details. This is quite different from Swift embedding in C++ where we don't see similar issues. I have been considering to drop golang and in favor of C/C++ client for prometheus for sometime, haven't made this happen yet.

For that reason, I think a C++/C client for elasticsearch, or using libcurl directly (through HttpRequest) might make sense. I noticed that elasticsearch also has no official C++/C client. We might just pick up a C++/C client that is maintained better?

Note Rust could also be an option as there is an official rust client for elasticsearch. However, integrate Rust with Bazel could be really painful. (Last time I tried to get rust working for AVIF code in our repo, haven't had success yet). Rust also might be exposed to the risk of its own future, given the recent situation in Mozilla. For that reason we might want to defer rust related integration.

So overall I think a quality (even if not official) C++/C client would make sense for now. libcurl could be considered if the C++/C client does not meet our expectations.

Let me know if you run into any problems or have any issues, especially with Bazel 😃 .

kvignesh1420:
[...]

PR #1098, opened 2020-08-21
Elasticsearch init ops and kernel

kvignesh1420:
This PR addresses the issue https://github.com/tensorflow/io/issues/1093 by adding implementations for the following:

- `IO>ElasticsearchReadableInit` Op.
- `ElasticsearchReadableInitOp` Kernel.
- `tfio.experimental.elasticsearch.ElasticsearchIODataset` template implementation and few helper methods.

Also, modified the `BUILD` file for Bazel build of `elasticsearch_ops`.

### DIFF
--- a/tensorflow_io/core/BUILD
+++ b/tensorflow_io/core/BUILD
@@ -190,6 +190,7 @@ cc_library(
         "@local_config_tf//:libtensorflow_framework",
         "@local_config_tf//:tf_header_lib",
         "@rapidjson",
+        "@curl",
     ],
     alwayslink = 1,
 )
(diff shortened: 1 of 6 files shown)


yongtang:
@kvignesh1420 Thanks for the PR! Would you mind to add a test in a follow up PR?

kvignesh1420:
@yongtang sure.

kvignesh1420:
@yongtang can we return a tensor of tensors from the kernel to the OP?

yongtang:
@kvignesh1420 I would assume an easy way is to obtain the json string from dataset implementation, then using `dataset.map()` internally (could be part of the ElasticsearchIODataset class) to deserialize the json string into a structure of tensors. You can take a look at `tfio.experimental.serialization.decode_json` and see if this fits the need.

Let me know if you run into any issues.

PR #1113, opened 2020-09-12
Upgrade bazel version to fix kokoro ci issue

kvignesh1420:
This PR upgrades the bazel version to 3.1.0 for the CI jobs so that the Kokoro jobs don't fail.

The kokoro env seems to have the bazel 3.1.0 version already setup while running the CI jobs and it is conflicting with the 3.0.0 version mentioned in `.bazelversion` file.


kvignesh1420:
cc: @yongtang @terrytangyuan

yongtang (approved):
LGTM

PR #1115, opened 2020-09-14
Add Prewitt and Sobel filter

yongtang:
This PR adds Prewitt and Sobel filter under `tfio.experimental.filter[prewitt|sobel]`

### DIFF
--- a/tensorflow_io/core/python/experimental/filter_ops.py
+++ b/tensorflow_io/core/python/experimental/filter_ops.py
@@ -243,3 +243,83 @@ def gabor(
         input, tf.cast(tf.math.imag(g), input.dtype), [1, 1, 1, 1], padding="VALID"
     )
     return tf.complex(real, imag)
+
+
+def prewitt(input, mode=None, constant_values=None, name=None):
+    """
+    Apply Prewitt filter to image.
+
+    Args:
+      input: A 4-D (`[N, H, W, C]`) Tensor.
+      mode: A `string`. One of "CONSTANT", "REFLECT", or "SYMMETRIC"
+        (case-insensitive). Default "CONSTANT".
+      constant_values: A `scalar`, the pad value to use in "CONSTANT"
+        padding mode. Must be same type as input. Default 0.
+      name: A name for the operation (optional).
+
+    Returns:
+      A 4-D (`[N, H, W, C]`) Tensor.
+    """
+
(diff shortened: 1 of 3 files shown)


tests/test_filter_eager.py
```diff
@@ -63,8 +63,46 @@
...
             ),
         ),
+        pytest.param(
+            lambda: (np.reshape(np.arange(25) / 25, [1, 5, 5, 1])),
+            lambda e: tfio.experimental.filter.prewitt(e),
+            lambda e: np.reshape(
+                np.sqrt(
+                    np.square(
+                        scipy.ndimage.prewitt(
+                            np.reshape(e, [5, 5]), axis=0, mode="constant"
```
terrytangyuan, line 73: Do we want to test multiple modes here?

yongtang:
@terrytangyuan Different modes can be grouped into a separate tests with all ops suitable for padding. Maybe we can leave this PR alone? I will create a follow up PR for modes.

terrytangyuan (approved):
Sounds good.

tensorflow_io/core/python/experimental/filter_ops.py
```diff
@@ -243,3 +243,83 @@ def gabor(
...
+    """
+    Apply Prewitt filter to image.
+
+    Args:
+      input: A 4-D (`[N, H, W, C]`) Tensor.
+      mode: A `string`. One of "CONSTANT", "REFLECT", or "SYMMETRIC"
+        (case-insensitive). Default "CONSTANT".
+      constant_values: A `scalar`, the pad value to use in "CONSTANT"
+        padding mode. Must be same type as input. Default 0.
+      name: A name for the operation (optional).
```
kvignesh1420, line 258: @yongtang just wanted to know how we will be using the `name` parameter in the future. (like naming the OP etc.)

kvignesh1420 (approved):

yongtang:
@kvignesh1420 The name is normally used with `tf.name_scope` See https://www.tensorflow.org/api_docs/python/tf/name_scope. It can also be passed to C++ kernel's `name` attributes (All C++ kernels by default generates a `name` attr once compiled). In tensorflow-io we haven't systematically pass `name` as needed. We should as we gradually improve the code base.

kvignesh1420:
@yongtang thanks for the info.

PR #1119, opened 2020-09-16
Bug fix in github actions while pushing docker images to registry

kvignesh1420:
The current github action job `docker-release` is unable to find the script for building and testing the docker image. This PR fixes that issue.


kvignesh1420:
@yongtang I missed the step in the docker-release ci job which is causing it to fail. This artefact checkout step fixes the bug.

yongtang (approved):
LGTM

PR #1120, opened 2020-09-16
kafka experimental datasets streaming logic modifications

kvignesh1420:
This PR addresses the issue: https://github.com/tensorflow/io/issues/947 by implementing the following:

- The blocking logic when the `stream_timeout` is set has been modified in such a way that, polls for new messages during the data slump calls the `Next` OP periodically instead of just waiting within a single `Next` OP call. The current approach prevents frequent rebalance calls when the `stream_timeout` is very high and is unable to capture the incoming data once the waiting period starts.
- the `message_timeout` parameter has been changed to `message_poll_timeout` to prevent confusion with `stream_timeout`.
- Additional documentation has been added pertaining to the `message_poll_timeout` parameter.


yongtang (approved):
LGTM

PR #1126, opened 2020-09-18
fix docstring indentations for arrow, bigquery and ignite python api's

kvignesh1420:
This PR addressed the issue: https://github.com/tensorflow/io/issues/861 and fixes the docstrings for the arrow, bigquery and ignite python api's. 

Will raise a different PR to fix the API's in `core` as well.


yongtang (approved):
LGTM

PR #1128, opened 2020-09-19
fix kafka dataset OOM ,Update kafka_kernels_deprecated.cc

777ki:
fix kafka dataset OOM problem, when use kafka dataset in online learning, the evaluation will create session and create dataset many times ,then that's problem will be every clearly to appear!


777ki:
fix kafka dataset OOM problem, when use kafka dataset in online learning, the evaluation will create session and create dataset many times ,then that's problem will be every clearly to appear!

777ki:
wrong email

PR #1129, opened 2020-09-19
fix kafka dataset OOM

777ki:
fix kafka dataset OOM problem, when use kafka dataset in online learning, the evaluation will create session and create dataset many times ,then that's problem will be every clearly to appear!


kvignesh1420 (commented):
@777ki thanks for the PR. There seem to be lint issues. Try running `bazel run //tools/lint:lint` from the root of the directory.

yongtang (approved):
LGTM

Issue #1133, opened 2020-09-24
[kafka-io] - `KafkaGroupIODataset` rebalance callback does not obey offset policy.


nmatare:
The rebalance callback does not obey the user configuration supplied to `KafkaGroupIODataset`.

It will always default to "earliest"
https://github.com/tensorflow/io/blob/d15b659bd0fe1054cdbd22dcc31d1aeea2977736/tensorflow_io/core/kernels/kafka_kernels.cc#L782

https://docs.confluent.io/2.0.0/clients/librdkafka/classRdKafka_1_1Topic.html

Testcase:

Yup, the below makes sense:

```python
import tensorflow_io as tfio
from kafka import KafkaProducer

SERVERS = 'localhost:29092'
TOPIC = 'testing_topic'

producer = KafkaProducer(bootstrap_servers=[SERVERS])

for letter in ['A', 'B', 'C', 'D', 'E']:
  producer.send(TOPIC, letter.encode('utf-8'))

dataset = tfio.experimental.streaming.KafkaGroupIODataset(
  topics=[TOPIC],
  group_id='test_group',
  servers=SERVERS,
  stream_timeout=10,
  configuration=["auto.offset.reset=earliest"],
)

next(iter(dataset))

# 2020-09-24 16:39:34.759944: I tensorflow_io/core/kernels/kafka_kernels.cc:787] The consumer group was newly created, reading from beginning
```

Nope, this should obey the user configuration and read from the config metadata.

```python
for letter in ['A', 'B', 'C', 'D', 'E']:
  producer.send(TOPIC, letter.encode('utf-8'))

dataset2 = tfio.experimental.streaming.KafkaGroupIODataset(
  topics=[TOPIC],
  group_id='test_group2',
  servers=SERVERS,
  stream_timeout=10,
  configuration=["auto.offset.reset=latest"],
)

next(iter(dataset2))

# 2020-09-24 16:39:34.759944: I tensorflow_io/core/kernels/kafka_kernels.cc:787] The consumer group was newly created, reading from beginning
```

nmatare:
@kvignesh1420

kvignesh1420:
@nmatare the PR has been merged, please try it out.

nmatare:
Confirming normal operation. Thanks for the work on the PR! Closing.

PR #1134, opened 2020-09-24
fix auto.offset.reset bug while consuming messages using KafkaGroupIODataset

kvignesh1420:
This PR addresses the issue: https://github.com/tensorflow/io/issues/1133 on `KafkaGroupIODataset` by:

- fixing the rebalance callback function.
- modifying the order in which topic and global level configurations are set for Kafka consumers.
- adding tests to validate the `auto.offset.reset` functionality.
- adding brief descriptions of the underlying implementation which `librdkafka` follows for rebalancing.

Once this PR is merged, will update the tutorial as well.


yongtang (approved):
LGTM

Issue #1111, opened 2020-09-10
Convert file system implementations to use C file system API in tensorflow


yongtang:
As C file system API in tensorflow is available (see https://github.com/tensorflow/community/pull/101), we should start migrate the file system implementations in tensorflow-io to use the C file system API.

This issue track the progresses for the following:
- [ ] Azure Storage
- [ ] OSS Storage
- [ ] HTTP FS

yongtang:
Before we can migrate to modular file system, additional work on tensorflow side is needed. The following is the progress of related PRs:
- [x] PR https://github.com/tensorflow/tensorflow/pull/43209
- [ ] PR https://github.com/tensorflow/tensorflow/pull/43424

yongtang:
Additional fixes that need to happen on tensorflow/tensorflow:
- [ ] PR https://github.com/tensorflow/tensorflow/pull/43563

yongtang:
One more fixes (remove duplication of `Env::Default()` instance) that is needed in tensorflow/tensorflow:

- [ ] PR https://github.com/tensorflow/tensorflow/pull/43594

terrytangyuan:
cc @zouxu09 Would anyone in your team be interested in helping out with OSS storage support?

yongtang:
@terrytangyuan One remaining issue in AZFS is that we commented out logging temporarily https://github.com/tensorflow/io/pull/1143/files#r517351641

This issue can be addressed by using C logging API in TensorFlow. A PR in tensorflow repo has been created to expose the header files in pip package: https://github.com/tensorflow/tensorflow/pull/44690

yongtang:
With AZFS and HTTP file system converted to modular file system C API we can thinking about maintain forward-compatibility now. One thing we can do, is to lazy load the `libtensorflow_io.so` (non-forward-compatible) and only imperatively load `libtensorflow_io_plugins.so` (forward-compatible). In this way it will be possible to allow user to use file systems in future TF versions when they only touch the following:
```
import tensorflow_io as tfio
```

I have created a PR #1208 for that purpose.

The only outstanding issue is that, if we lazy-load `libtensorflow_io.so`, then OSS file system (still inside `libtensorflow_io.so`) will not be immediately available with `import tensorflow_io as tfio`. We can still have some other ways, e.g., expose `import tensorflow_io.experimental.oss as oss` to load `libtensorflow_io.so` for oss. Though we may also want to think about the timeline.

[...]

PR #1143, opened 2020-10-05
Convert AZFS to use C modular file system API

yongtang:
This PR converts AZFS to use modular C file system API.

This PR is part of the #1111.

This PR depends on PRs in tensorflow/tensorflow:
 - [x] PR tensorflow/tensorflow#43594
 - [x] PR tensorflow/tensorflow#43680

~~**Update:** Most of the issues have been resolved. The following PR on tensorflow/tensorflow repo will help but is not a blocking issue:~~
~~- [ ] PR tensorflow/tensorflow/pull/44507~~

~~We will be ready once tensorflow release `2.4.0rc` is out.~~

Update: All issues resolved. PR  tensorflow/tensorflow/pull/44507 is not necessary. The TensorFlow 2.4.0rc0 has been bumped as well.


yongtang:
The PR is ready for review now. Note the tensorflow version has been bumped to 2.4.0rc0 (released yesterday).

tensorflow_io/core/kernels/azfs_kernels.cc
```diff
@@ -154,20 +180,21 @@ std::shared_ptr<azure::storage_lite::storage_credential> get_credential(
...
 azure::storage_lite::blob_client_wrapper CreateAzBlobClientWrapper(
-    const std::string &account) {
+    const std::string& account) {
   azure::storage_lite::logger::set_logger(
-      [](azure::storage_lite::log_level level, const std::string &log_msg) {
+      [](azure::storage_lite::log_level level, const std::string& log_msg) {
         switch (level) {
           case azure::storage_lite::log_level::info:
-            _TF_LOG_INFO << log_msg;
+            // _TF_LOG_INFO << log_msg;
```
terrytangyuan, line 190: Is there any reason why these logs are commented out?

yongtang: @terrytangyuan The logging will require the plugin.so depends on C++ APIs in tensorflow which will makes cross-version (e.g., 0.16.0 support TF 2.4, 2.5, 2.6). So this line is temporarily disabled. Once TF exposes C API for logging we will add it back (using C API instead).

tensorflow_io/core/BUILD
```diff
@@ -7,6 +7,19 @@ load(
...
+
+bool_flag(
+    name = "static_build",
+    build_setting_default = False,
+)
+
+config_setting(
+    name = "static_build_on",
+    flag_values = {
+        ":static_build": "True",
```
kvignesh1420, line 20: @yongtang shouldn't this be a bool?

yongtang: @kvignesh1420 I tried before and it looks like the `flag_values` will take a `key:value` of `string:string`.

tensorflow_io/core/kernels/file_system_plugins.h
```diff
@@ -0,0 +1,38 @@
...
+static void plugin_memory_free(void* ptr) { free(ptr); }
+
+namespace az {
+
+void ProvideFilesystemSupportFor(TF_FilesystemPluginOps* ops, const char* uri);
+
+}  // namespace az
+
+}  // namespace io
+}  // namespace tensorflow
```
kvignesh1420, line 36: How do we define the namespaces for httpFS or others that might come in the future?

tensorflow_io/core/python/ops/version_ops.py
```diff
@@ -14,5 +14,5 @@
 # ==============================================================================
 """version_ops"""
 
-package = "tensorflow>=2.3.0,<2.4.0"
-version = "0.16.0"
+package = "tensorflow>=2.4.0rc0,<2.5.0"
+version = "0.17.0"
```
kvignesh1420, line 18: Is the `version==0.17.0` change intentional?

tensorflow_io/core/kernels/file_system_plugins.h
yongtang: @kvignesh1420 Yes the plan is to have 
```
namespace http {

 void ProvideFilesystemSupportFor(TF_FilesystemPluginOps* ops, const char* uri);

 }  // namespace http

```

Already have code ready for HTTP migration. Will create the PR once this azfs PR merges.

tensorflow_io/core/python/ops/version_ops.py
yongtang: @kvignesh1420 Yes we want to bump the version so that it will not interference with docker build (we don't want 0.16's docker image depends on "tensorflow>=2.4.0rc0,<2.5.0" as it will not work (0.16 only works with TF 2.3).

README.md
```diff
@@ -311,6 +311,10 @@ libraries (.so). The gcc provided by Developer Toolset and rh-python36 should be
 Also, the libstdc++ has to be linked statically to avoid discrepancy of libstdc++ installed on
 CentOS vs. newer gcc version by devtoolset.
 
+Furthermore, a special flag `--//tensorflow_io/core:static_build` has to be passed to Bazel
```
kvignesh1420, line 314: @yongtang not sure if it is the right way but can we use `plugin_build` instead of `static_build`?

yongtang: @kvignesh1420 The reason to add this special flag is a little tricky.

In this PR, we build two shared libraries, one is `libtensorflow_io.so` which depends on C++ API and will not be usable for later versions of TF (e.g., 2.5, 2.6, etc as we only build against TF 2.4).

The other shared library is the newly introduced `libtensorflow_io_plugins.so` which depends on C API only and will be usable for later versions of TF (it can be used for TF 2.6, 2.7, etc). 

However, the only glitch is the CentOS 7 platform. On CentOS 7 we can only use RedHat's Software Collections to get a newer version of gcc (default versions of gcc on CentOS 7 will not work), and with static C++ linkage. Because of that, having two shared libraries `libtensorflow_io_plugins.so` and `libtensorflow_io.so` will not work and will cause segment fault.

As a result, we actually merges `libtensorflow_io_plugins.so` as part of `libtensorflow_io.so`. That is why we call it `static_build` to indicating it is for C++ static linkage.

tensorflow_io/core/kernels/azfs_kernels.cc
```diff
@@ -95,7 +120,8 @@ Status ParseAzBlobPath(StringPiece fname, bool empty_object_ok,
     *object = std::string(objectp.substr(pos + 1));
```
kvignesh1420, line 120: I think the check for `empty_object_ok == True` condition is missing.

tensorflow_io/core/kernels/azfs_kernels.cc
```diff
@@ -356,249 +454,235 @@ class AzBlobWritableFile : public WritableFile {
...
+                 "Cannot create storage accounts");
+    return;
   }
 
   // Blob storage has virtual folders. We can make sure the container exists
   auto blob_client_wrapper = CreateAzBlobClientWrapper(account);
 
   if (blob_client_wrapper.container_exists(container)) {
-    return Status::OK();
+    TF_SetStatus(status, TF_OK, "");
```
kvignesh1420, line 630: Maybe add a message saying container exists?

tensorflow_io/core/kernels/azfs_kernels.cc
terrytangyuan: Sounds good.

tensorflow_io/core/kernels/azfs_kernels.cc
```diff
@@ -356,249 +454,235 @@ class AzBlobWritableFile : public WritableFile {
...
+  if (TF_GetCode(status) != TF_OK) {
+    return;
+  }
+
+  auto blob_client = CreateAzBlobClientWrapper(account);
+
+  blob_client.delete_blob(container, object);
+  if (errno != 0) {
+    std::string error_message =
+        absl::StrCat("Failed to delete ", path, " (", errno_to_string(), ")");
```
kvignesh1420, line 662: append "ERROR_CODE" before `errno_to_string()` for clarity purposes.

tensorflow_io/core/kernels/azfs_kernels.cc
```diff
@@ -356,249 +454,235 @@ class AzBlobWritableFile : public WritableFile {
...
   // Doesn't support file delete - call GetChildren (without delimiter) and then
   // loop and delete
 
   std::string account, container, object;
-  TF_RETURN_IF_ERROR(
-      ParseAzBlobPath(dirname, false, &account, &container, &object));
+  ParseAzBlobPath(path, false, &account, &container, &object, status);
+  if (TF_GetCode(status) != TF_OK) {
+    return;
+  }
```
kvignesh1420, line 678: Better to use TF_SetStatus here to explain why the deletion has not happened.

tensorflow_io/core/kernels/azfs_kernels.cc
```diff
@@ -608,109 +692,189 @@ Status AzBlobFileSystem::DeleteDir(const std::string &dirname) {
...
+      TF_SetStatus(status, TF_INTERNAL, error_message.c_str());
+      return;
     }
   } else {
     // Delete all blobs under dirname prefix
     std::vector<std::string> children;
-    TF_RETURN_IF_ERROR(ListResources(dirname, "", blob_client, &children));
+    ListResources(path, "", blob_client, &children, status);
+    if (TF_GetCode(status) != TF_OK) {
+      return;
```
kvignesh1420, line 705: Better to use TF_SetStatus here to explain that the deletion has not happened due to an error while listing resources.

tensorflow_io/core/kernels/azfs_kernels.cc
```diff
@@ -608,109 +692,189 @@ Status AzBlobFileSystem::DeleteDir(const std::string &dirname) {
...
   // Status can be success, pending, aborted or failed
   std::string copy_status;
   do {
-    const auto target_blob_property =
-        blob_client.get_blob_property(target_container, target_object);
-    copy_status = target_blob_property.copy_status;
+    const auto dst_blob_property =
+        blob_client.get_blob_property(dst_container, dst_object);
+    copy_status = dst_blob_property.copy_status;
   } while (copy_status.find("pending") == 0 && !copy_status.empty());
```
kvignesh1420, line 772: maybe `sleep` in-between status checks?

tensorflow_io/core/kernels/azfs_kernels.cc
```diff
@@ -608,109 +692,189 @@ Status AzBlobFileSystem::DeleteDir(const std::string &dirname) {
...
 
   if (copy_status.find("success") == std::string::npos) {
-    return errors::Internal("Process of renaming resulted in status of ",
-                            copy_status, " when renaming ", src, " to ",
-                            target);
+    std::string error_message =
+        absl::StrCat("Process of renaming resulted in status of ", copy_status,
+                     " when renaming ", src, " to ", dst);
```
kvignesh1420, line 777: absl::StrCat("Process of renaming from , src, " to ", dst, " resulted in status of ", copy_status);

tensorflow_io/core/kernels/azfs_kernels.cc
```diff
@@ -608,109 +692,189 @@ Status AzBlobFileSystem::DeleteDir(const std::string &dirname) {
...
   }
 
-  return Status::OK();
+  TF_SetStatus(status, TF_OK, "");
+}
+
+static void CopyFile(const TF_Filesystem* filesystem, const char* src,
+                     const char* dst, TF_Status* status) {
+  // 128KB copy buffer
+  constexpr size_t kCopyFileBufferSize = 128 * 1024;
```
kvignesh1420, line 797: is there a way to check how the buffer size affects performance?

tensorflow_io/core/kernels/azfs_kernels.cc
```diff
@@ -608,109 +692,189 @@ Status AzBlobFileSystem::DeleteDir(const std::string &dirname) {
...
+      return;
+    }
+    offset += bytes_to_read;
+  }
+  dst_file->Close(status);
 }
 
-Status AzBlobFileSystem::RecursivelyCreateDir(const string &dirname) {
-  return CreateDir(dirname);
+static void PathExists(const TF_Filesystem* filesystem, const char* path,
```
kvignesh1420, line 835: Shouldn't this return a `bool`?

kvignesh1420: similar to `static bool IsDirectory()`

tensorflow_io/core/kernels/azfs_kernels.cc
yongtang: Thanks @kvignesh1420. Updated to capture `empty_object_ok == True`

tensorflow_io/core/kernels/azfs_kernels.cc
yongtang: @kvignesh1420 In Azure, the container is in reference to S3's bucket and we treat them as part of the `directory` path, i.e., `/container/path/to/file` is considered as one path with AZFS like other file systems (not to treat `container` part differently). Given this function is to see if the `path` already exists. I think it might be fine to not single out container.

tensorflow_io/core/kernels/azfs_kernels.cc
yongtang: @kvignesh1420 Updated.

tensorflow_io/core/kernels/azfs_kernels.cc
yongtang: @kvignesh1420 It is less obvious though the `status` returned by `ListResource` set both the code and the error message. `TF_GetCode(status)` is to check if the status is bad. If status != OK then return immediately.

tensorflow_io/core/kernels/azfs_kernels.cc
yongtang: @kvignesh1420 Done.

tensorflow_io/core/kernels/azfs_kernels.cc
yongtang: Thanks @kvignesh1420 . Updated.

tensorflow_io/core/kernels/azfs_kernels.cc
yongtang: @kvignesh1420 The function signature is defined by C modular system API so we will  not be able to change it (return bool).

tensorflow_io/core/kernels/azfs_kernels.cc
yongtang: @kvignesh1420 That would be an interesting to take a look. Since this PR is to migrate to C modular file system API and 128K buffer was copied from the existing code, maybe we can investigate later to find the optimal buffer size.

yongtang:
Thanks @kvignesh1420 for the review. The PR has been updated. Please take a look.

tensorflow_io/core/kernels/azfs_kernels.cc
kvignesh1420: okay sure.

kvignesh1420 (commented):
@yongtang I see that we have removed the tests for azfs. Will this be taken up in a different PR?  Otherwise LGTM.

yongtang:
@kvignesh1420 the old tests were based on the C++ API in TensorFlow. As we use C API now they are irrelevant now. We still have python integration tests (with azure emulator) so we are covered here. We can work on creating follow up PRs to add additional coverage and relevant to C modular file systems APIs after this PR is merged.

kvignesh1420 (approved):
LGTM.

damienpontifex:
Nice work @yongtang ! 🎉

PR #1145, opened 2020-10-07
Add bool support for hdf5

yongtang:
This PR adds bool support for hdf5.

This PR fixes #1144.

Signed-off-by: Yong Tang <yong.tang.github@outlook.com>

PR #1147, opened 2020-10-08
Update python version to 3.8 due to GitHub Actions update

yongtang:
This PR updates python version to 3.8.6 as GitHub Actions update (from 3.8.5)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -315,10 +315,10 @@ jobs:
       - uses: actions/checkout@v2
       - uses: actions/setup-python@v1
         with:
-          python-version: 3.8.5
+          python-version: 3.8.6
       - name: Bazel on Windows
         env:
-          PYTHON_VERSION: 3.8.5
+          PYTHON_VERSION: 3.8.6
           BAZEL_VC: "C:/Program Files (x86)/Microsoft Visual Studio/2019/Enterprise/VC/"
         shell: cmd
         run: |
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -315,18 +315,18 @@ jobs:
       - uses: actions/checkout@v2
       - uses: actions/setup-python@v1
         with:
-          python-version: 3.8.6
+          python-version: 3.8
       - name: Bazel on Windows
         env:
-          PYTHON_VERSION: 3.8.6
           BAZEL_VC: "C:/Program Files (x86)/Microsoft Visual Studio/2019/Enterprise/VC/"
         shell: cmd
         run: |
           @echo on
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -315,10 +315,10 @@ jobs:
       - uses: actions/checkout@v2
       - uses: actions/setup-python@v1
         with:
-          python-version: 3.8.5
+          python-version: 3.8.6
```
kvignesh1420, line 318: @yongtang why don't we leave it at 3.8?

.github/workflows/build.yml
```diff
@@ -315,10 +315,10 @@ jobs:
...
       - name: Bazel on Windows
         env:
-          PYTHON_VERSION: 3.8.5
+          PYTHON_VERSION: 3.8.6
```
terrytangyuan, line 321: What's the context for this change?

yongtang:
@terrytangyuan  @kvignesh1420 The error is caused by GitHub Actions recent change from 3.8.5 => 3.8.6, which caused the following (see logs in https://github.com/tensorflow/io/pull/1145/checks?check_run_id=1226724329) :
```
2020-10-08T15:33:41.4951411Z ##[error]Version 3.8.5 with arch x64 not found
Available versions:
2.7.18 (x86)
3.5.4 (x86)
3.6.8 (x86)
3.7.9 (x86)
3.8.6 (x86)
2.7.18 (x64)
3.5.4 (x64)
3.6.8 (x64)
3.7.9 (x64)
3.8.6 (x64)
```

yongtang:
@terrytangyuan @kvignesh1420 On windows, we need `python3` command to run. However, GitHub Actions on Windows seems to only expose `python` command. As a result I use the following to copy `python` binary as `python3`:
https://github.com/tensorflow/io/blob/master/.github/workflows/build.yml#L330

kvignesh1420:
@yongtang so was thinking if we can just leave it `3.8` and let Github actions take the latest version. What do you think?

yongtang:
@kvignesh1420 We will also need to find python's absolute exe path:
https://github.com/tensorflow/io/blob/master/.github/workflows/build.yml#L330

It might be possible without explicitly calling out python's exact version?

kvignesh1420:
@yongtang as per your ref: https://github.com/tensorflow/io/blob/master/.github/workflows/build.yml#L330

can we set the`PYTHON_VERSION` var to `python --version` before moving to python3?

yongtang:
@kvignesh1420 updated. Please take a look.

terrytangyuan (approved):
LGTM now

kvignesh1420 (approved):
@yongtang  LGTM.

PR #1149, opened 2020-10-09
Elasticsearch tutorial

kvignesh1420:
This PR adds a tutorial for the `elasticsearch.ElasticsearchIODataset` in the form of a colab notebook. Addresses: https://github.com/tensorflow/io/issues/1093


docs/tutorials/elasticsearch.ipynb
```diff
@@ -0,0 +1,757 @@
...
+      },
+      "source": [
+        "## Overview\n",
+        "\n",
+        "This tutorial focuses on streaming data from an [Elasticsearch](https://www.elastic.co/guide/en/elasticsearch/reference/current/elasticsearch-intro.html) cluster into a `tf.data.Dataset` which is then used in conjunction with `tf.keras` for training and inference.\n",
+        "\n",
+        "Elasticseach is primarily a distributed search engine which supports storing structured, unstructured along with geospatial and numeric data. For the purpose of this tutorial, a dataset with structured records is considered.\n",
+        "\n",
+        "**NOTE:** A basic understanding of [elasticsearch storage](https://www.elastic.co/guide/en/elasticsearch/reference/current/documents-indices.html) will help you in following the tutorial with ease."
+      ]
```
yongtang, line 78: This tutorial consists of installation of elasticsearch python package. I am wondering if it makes sense to mention that installation of elasticsearch python package is only uses to setup elasticsearch server and data; and in production only the installation of tensorflow-io package is needed?

docs/tutorials/elasticsearch.ipynb
```diff
@@ -0,0 +1,757 @@
...
+    },
+    {
+      "cell_type": "code",
+      "execution_count": null,
+      "metadata": {
+        "id": "48LqMJ1BEHm5"
+      },
+      "outputs": [],
+      "source": [
+        "!ps -ef | grep elasticsearch"
```
yongtang, line 232: While trying the notebook I noticed that elasticsearch may take several seconds to be ready (encountered `connection refused` error the first time. Maybe its worth to have a `wait for 10 seconds` to make sure server is ready?

docs/tutorials/elasticsearch.ipynb
yongtang: @kvignesh1420 alternatively, maybe we can put a section of `server setup and data preparation` to include everything else, and then start with a new section of `import tensorflow_io as tfio` to emphasis that using tfio for elasticsearch does not need to install anything else other than tfio and tensorflow packages?

yongtang:
@kvignesh1420 can you also add an entry in `docs/tutorials/_toc.yaml`? That will create the TOC entry.

docs/tutorials/elasticsearch.ipynb
kvignesh1420: refstructured the sections and added the relevant info in the setup and tfio-datasets section

docs/tutorials/elasticsearch.ipynb
kvignesh1420: increased it to 15 seconds.

yongtang (approved):
Overall LGTM.

MarkDaoust (approved):

docs/tutorials/elasticsearch.ipynb
```diff
@@ -0,0 +1,750 @@
...
+      "cell_type": "code",
+      "execution_count": null,
+      "metadata": {
+        "id": "YUj0878jPyz7"
+      },
+      "outputs": [],
+      "source": [
+        "!wget https://artifacts.elastic.co/downloads/elasticsearch/elasticsearch-oss-7.9.2-linux-x86_64.tar.gz\n",
+        "!wget https://artifacts.elastic.co/downloads/elasticsearch/elasticsearch-oss-7.9.2-linux-x86_64.tar.gz.sha512\n",
+        "!shasum -a 512 -c elasticsearch-oss-7.9.2-linux-x86_64.tar.gz.sha512 \n",
```
MarkDaoust, line 189: `!` commands don't cancel execution if they fail.

If you use `%%bash`, and make the `shasum` the last command in the cell, then it will stop execution on failure.

docs/tutorials/elasticsearch.ipynb
```diff
@@ -0,0 +1,750 @@
...
+    },
+    {
+      "cell_type": "code",
+      "execution_count": null,
+      "metadata": {
+        "id": "YUj0878jPyz7"
+      },
+      "outputs": [],
+      "source": [
+        "!wget https://artifacts.elastic.co/downloads/elasticsearch/elasticsearch-oss-7.9.2-linux-x86_64.tar.gz\n",
```
MarkDaoust, line 187: The wget output isn't very helpful, can you add some `-q` here?

docs/tutorials/elasticsearch.ipynb
```diff
@@ -0,0 +1,750 @@
...
+      "cell_type": "code",
+      "execution_count": null,
+      "metadata": {
+        "id": "-7DzTScGa6lW"
+      },
+      "outputs": [],
+      "source": [
+        "ds_iter = iter(train_ds)\n",
+        "batch = next(ds_iter)\n",
+        "batch"
```
MarkDaoust, line 516: Big printouts like this are hard to read. Consider using:

```
features, label = next(ds_iter)
print(f"label: {label.numpy()[0]}")
{key: value.numpy()[0] for key,value in batch.items()}
```

docs/tutorials/elasticsearch.ipynb
```diff
@@ -0,0 +1,750 @@
...
+        "\n",
+        "# design/build the model\n",
+        "x = tf.keras.layers.Dense(32, activation=\"relu\")(all_features)\n",
+        "x = tf.keras.layers.Dropout(0.5)(x)\n",
+        "x = tf.keras.layers.Dense(64, activation=\"relu\")(x)\n",
+        "x = tf.keras.layers.Dropout(0.5)(x)\n",
+        "output = tf.keras.layers.Dense(1)(x)\n",
+        "model = tf.keras.Model(all_inputs, output)\n",
+        "\n",
+        "print(model.summary())"
```
MarkDaoust, line 663: If the model isn't a simple sequential, you can't see the structure with summary. prefer `plot_model`:

```
tf.keras.utils.plot_model(model, rankdir='LR', show_shapes=True)
```

MarkDaoust (changes requested):
Looking again, I have a couple of suggestions.

I tried running this a few times. 

A full top to bottom run always works for me, but if I go back and re-executs a couple of cells I usually end up with something like:

```
FailedPreconditionError: Corrupted response from the server null
	 [[{{node IO>ElasticsearchReadableNext}}]]
```

or

```
ConnectionError: No healthy node available for the index: train, please check the cluster config
```

So this feature isn't very stable yet.

kvignesh1420:
> Looking again, I have a couple of suggestions.
> 
> I tried running this a few times.
> 
> A full top to bottom run always works for me, but if I go back and re-executs a couple of cells I usually end up with something like:
> 
> ```
> FailedPreconditionError: Corrupted response from the server null
> 	 [[{{node IO>ElasticsearchReadableNext}}]]
> ```
> 
> or
> 
> ```
> ConnectionError: No healthy node available for the index: train, please check the cluster config
> ```
> 
> So this feature isn't very stable yet.

@MarkDaoust thanks for the review. I ran the same notebook on my mac and the cells which throw the error in colab are working fine locally.

For example:

local env
```ipython
In [20]: train_ds = tfio.experimental.elasticsearch.ElasticsearchIODataset(
    ...:         nodes=[ES_NODES],
    ...:         index="train",
    ...:         doc_type="pet",
    ...:         headers=HEADERS
    ...:     )
    ...:
    ...: # Prepare a tuple of (features, label)
    ...: train_ds = train_ds.map(lambda v: (v, v.pop("target")))
    ...: train_ds = train_ds.batch(BATCH_SIZE)
Connection successful: http://localhost:9200/_cluster/health

In [21]: train_ds = tfio.experimental.elasticsearch.ElasticsearchIODataset(
    ...:         nodes=[ES_NODES],
    ...:         index="train",
    ...:         doc_type="pet",
    ...:         headers=HEADERS
    ...:     )
    ...:
    ...: # Prepare a tuple of (features, label)
    ...: train_ds = train_ds.map(lambda v: (v, v.pop("target")))
    ...: train_ds = train_ds.batch(BATCH_SIZE)
Connection successful: http://localhost:9200/_cluster/health
```

However, when you run this cell multiple times in colab, it throws:

```console
ConnectionError: No healthy node available for the index: train, please check the cluster config
```

Not sure why this is the case. Other comments have been addressed.

docs/tutorials/elasticsearch.ipynb
kvignesh1420: addressed in the commit

docs/tutorials/elasticsearch.ipynb
kvignesh1420: addressed in the commit

docs/tutorials/elasticsearch.ipynb
kvignesh1420: addressed in the commit

docs/tutorials/elasticsearch.ipynb
kvignesh1420: addressed in the commit

MarkDaoust (approved):
My comments are addressed.

> I ran the same notebook on my mac and the cells which throw the error in Colab are working fine locally.

I don't think this is blocking since this is clearly labeled as experimental.

PR #1151, opened 2020-10-10
Use --enable_platform_specific_config flag in bazelrc to specify macos specific rules in bazel

yongtang:
Bazel has an option of --enable_platform_specific_config which allows specification of platform specific flags. The effect is that `bazel:macos`, `bazel:windows`, `bazel:linux` inside .bazelrc will be automatically picked up bazed on the platform. So rules specified for macos, windows, and linux can co-exist in .bazelrc.

This helps in avoiding specifying different .bazelrc files for different platforms.

This PR utilize this flag to simplify the bazelrc generation.

Also, on macos, `--copt="-mmacosx-version-min=10.13"` and `--linkopt="-mmacosx-version-min=10.13"` are specified by default so that local build and GitHub Actions build will remain the same.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -128,8 +128,6 @@ jobs:
           bazel build \
             ${BAZEL_OPTIMIZATION} \
             --copt -Wunguarded-availability \
-            --copt -mmacosx-version-min=10.13 \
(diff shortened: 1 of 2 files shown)


kvignesh1420 (approved):
LGTM.

PR #1153, opened 2020-10-18
[docker] push version specific tag during docker release

kvignesh1420:
Currently, we push `latest`, `latest-devel` and `nightly` tags to the docker registry. However, we do not have a tag which captures the version of the release. This PR addresses the issue by tagging the images with version obtained from `setup.py`

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -486,6 +486,8 @@ jobs:
           docker login --username tfsigio --password ${{ secrets.DOCKER_PASSWORD }}
           bash -x -e tools/docker/tests/dockerfile_cpu_test.sh
           docker push tfsigio/tfio:latest
+          TFIO_VERSION=$(python setup.py --version)
+          docker push tfsigio/tfio:${TFIO_VERSION}
           bash -x -e tools/docker/tests/dockerfile_nightly_test.sh
           docker push tfsigio/tfio:nightly
           bash -x -e tools/docker/tests/dockerfile_devel_test.sh


terrytangyuan (approved):
LGTM. Though should we tag previously released versions as well?

kvignesh1420:
@terrytangyuan the credentials are with @yongtang. We can use the `tools/docker/cpu.Dockerfile` to pass the tfio version and build. This is a one-time thing though.

yongtang:
@kvignesh1420 @terrytangyuan Sorry for the late response as I was occupied heavily recently. I will try to see if I can push the early docker images. (I tried to see if we can have a team account so that anyone from the team can push the image, but looks like docker hub just had too many restrictions).

kvignesh1420:
@yongtang no issues. Few things to note if you are pushing the earlier versions:
- [ ] Use the version number as the tag instead of adding `v` to the beginning. (for ex: 0.15.0 instead of v0.15.0). This way we can be in sync with the [tensorflow tags](https://hub.docker.com/r/tensorflow/tensorflow/tags)
- [ ] Remove the existing tag `v0.15.0`
- [ ] Delete the old account `https://hub.docker.com/r/tfiocontainer/tfio`

yongtang:
Pushed old releases from 0.8.0 => 0.14.0 to docker hub. I didn't push earlier versions than 0.8.0 as 0.8.0 was the version tied to TF 1.15. We want to give user an option of TF 1.x but we don't want to go any earlier 1.x versions.

kvignesh1420:
@yongtang thanks!

PR #1154, opened 2020-10-20
[docker] incompatible python version fix

kvignesh1420:
While capturing the tfio version from `setup.py`, github action throws the following error:
```
 python setup.py --version
  File "setup.py", line 97, in <module>
    with open(os.path.join(here, "README.md"), encoding="utf-8") as f:
TypeError: 'encoding' is an invalid keyword argument for this function
```

This PR uses the latest python version in GH Action to handle the issue.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -486,7 +486,7 @@ jobs:
           docker login --username tfsigio --password ${{ secrets.DOCKER_PASSWORD }}
           bash -x -e tools/docker/tests/dockerfile_cpu_test.sh
           docker push tfsigio/tfio:latest
-          TFIO_VERSION=$(python setup.py --version)
+          TFIO_VERSION=$(python3 setup.py --version)
           docker push tfsigio/tfio:${TFIO_VERSION}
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -481,11 +481,15 @@ jobs:
     runs-on: ubuntu-18.04
     steps:
       - uses: actions/checkout@v2
+      - uses: actions/setup-python@v1
+        with:
+          python-version: 3.8
```
terrytangyuan, line 486: Is this just for obtaining the version of TFIO? Perhaps a simple bash would be more reasonable than requiring a specific version of Python?

kvignesh1420: @terrytangyuan any python3 version should address the issue I suppose (need to check after merging). I chose 3.8 as it was the latest. 

Also, if we are using bash to get the version, we would have to grep for version in the `tensorflow_io/core/python/ops/version_ops.py` file and perform some split operations. I felt just fetching the version from setup.py would be simpler. Let me know if I am missing anything. Thanks.

terrytangyuan (approved):

PR #1155, opened 2020-10-20
[docker] add version specific local tag before pushing to registry

kvignesh1420:
This PR adds the missing local tag to the docker image before pushing to the registry.
Error from GH Action:

```
+ docker push tfsigio/tfio:0.15.0
The push refers to repository [docker.io/tfsigio/tfio]
tag does not exist: tfsigio/tfio:0.15.0
```


kvignesh1420:
Merging as of now, since the action is effective only on merges.

terrytangyuan (approved):

Issue #1161, opened 2020-10-24
Issues with building IODataset from truly hierarchical hdf5 


Antymon:
Hi, I have hdf5 files with datasets (features) nested under groups (samples) in the following way: `/sample_id/feature_column_id`. Is it possible to make a dataset of this with `IODataset.from_hdf5`? I have been trying in eager mode but failed. Keep getting `tensorflow.python.framework.errors_impl.InvalidArgumentError: unsupported data class for /eps0_t0/label_done: 8 [Op:IO>HDF5ReadableInfo]`
by simply calling `features = tfio.IODataset.from_hdf5(path,"/")`

Dependencies:
tensorflow-io Version: 0.15.0
tensorflow Version: 2.3.0

Also, whatever knowledge I get out of this question might be also worth putting as docs as I haven't found anything on how to use this part of API.

kvignesh1420:
@Antymon I think you have to give the path to each feature column (based on the current implementation of the `IODataset`).

For example:

```python
for s_id in sample_ids:
  for c_id in feature_column_ids:
    feature_ds = tfio.IODataset.from_hdf5(path,"/{}/{}".format(s_id, c_id))
```

You can then zip multiple of these `feature_ds` datasets into a single dataset for a `sample_id`.

Antymon:
Thanks, so I would use an `IODataset` object for something that is merely a feature of an example. I do wonder what would be performance implications, coz this feels like an approach that is going around the original intent of how hdf5 should be structured Perhaps my problem is the file itself as it is not really meant to be hierarchical?

kvignesh1420:
@Antymon maybe you can represent your dataset in another format? Check the PR https://github.com/tensorflow/io/pull/1163 for a discussion on this.

Antymon:
Interesting. How about if I stored each feature for all samples in a single group? Wouldn't that convention make my h5 columnar? Then I would have as many datasets as features and perhaps could zip those together. Does that sound like an improvement over my current structure?

kvignesh1420:
@Antymon can this issue be closed?

Antymon:
Well, it can, of course. As a user, I would wish for some instructions on h5 to appear in the docs, but I leave it to your judgment. I am speculating that one could mimic TFRecord in serializing all tensors to strings and dump everything as one string matrix under a single h5 dataset, but I haven't benchmarked either this or what was originally suggested. As for my problem, I will simply go for now with TFRecord.

kvignesh1420:
[...]

PR #1163, opened 2020-10-26
[hdf5] extend tests to cover grouped data

kvignesh1420:
This PR extends the test case for hdf5 dataset based on the discussion in https://github.com/tensorflow/io/issues/1161.

@yongtang is there a way to choose multiple columns (like we can in json by giving a spec) for the hdf5 dataset? The current implementation requires us to create a dataset for each hierarchy in hdf5.

### DIFF
--- a/tests/test_hdf5_eager.py
+++ b/tests/test_hdf5_eager.py
@@ -62,6 +62,43 @@ def test_hdf5():
     shutil.rmtree(runpath)
 
 
+def test_hdf5_grouped():
+    """test_hdf5 with grouped data: https://github.com/tensorflow/io/issues/1161"""
+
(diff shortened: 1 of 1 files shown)


yongtang (approved):
LGTM

yongtang:
@kvignesh1420 multiple columns could be challenging, as HDF5 itself is not really a columnar storage (like parquet or CSV) and each `dataset` in HDF5 could have any shape or length. That will makes tf.data.Dataset pipeline uneven in case dataset in HDF5 are not of the same length.

kvignesh1420:
@yongtang thanks for the info.

terrytangyuan (approved):

PR #1167, opened 2020-10-31
[github-action] fix macOS task

kvignesh1420:
This PR addresses the currently failing macOS task in GitHub Actions.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -51,9 +51,9 @@ jobs:
           echo "Bring /usr/bin to front as GitHub does not use system python3 by default"
           export PATH=/usr/bin:$PATH
           echo $PATH
-          # echo "Note the following is to fix a bug in Apple's python 3.7.3"
-          # ls /Applications/Xcode.app/Contents/Developer/Library/Frameworks/Python3.framework/Versions
-          # sudo sed -i.bak 's/sys.modules.values()/list(sys.modules.values())/g' /Applications/Xcode.app/Contents/Developer/Library/Frameworks/Python3.framework/Versions/3.7/lib/python3.7/linecache.py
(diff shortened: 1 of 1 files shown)


yongtang:
@kvignesh1420 The intention of the "macOS test" on README.md is to make sure if an user only uses `system default` python on macOS (not installed from various other python such as pyenv etc), the build can still succeed. As such you might noticed the previous python 3.7.3 fix that was intended to `macOS 10.15`'s system default path.

Note also as the intention of `macOS test` on README.md is to only ensure that it allows developers to build tensorflow-io on macOS, it does not perform a full test. Instead, only `test_serialization.py` has been invoked for sanity purposes.  The `true` build we use to release pip package for macOS, on the other hand, is based on GitHub Actions' python install.

Now it looks like some recent update on macOS switched from python `3.7.3` to `3.8.2` for its system default. That breaks our `README.md` build. However, I think we still want to use `system default` for "macOS test" on README.md. I think it might be possible to fix the README.md test while still pointing to system default python on macOS.

kvignesh1420:
@yongtang Yes, the default python version on `macos-latest` in github actions has changed from `3.7.3` to `3.8.6`.
[macos-latest(macOS 11)](https://github.com/actions/virtual-environments/blob/main/images/macos/macos-11.0-Readme.md). The same applies to [macOS 10.15](https://github.com/actions/virtual-environments/blob/main/images/macos/macos-10.15-Readme.md) as well. So as of now the system default python (3.8.6) in GitHub actions is different from what users might have. (3.7.x). Not sure, if there is a way to address this other than using `actions/setup-python`.

yongtang:
@kvignesh1420 I think the issue is related to requirements.txt, created a PR #1168, please take a look.

kvignesh1420:
Addressed in https://github.com/tensorflow/io/pull/1168

PR #1168, opened 2020-10-31
Fix macOS build failure by adding `export ARCHFLAGS="-arch x86_64"`

yongtang:
This is an attempt to fix macOS build with system default python.

See https://github.com/giampaolo/psutil/issues/1832 for similar issues.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -51,8 +51,6 @@ jobs:
           echo "Bring /usr/bin to front as GitHub does not use system python3 by default"
           export PATH=/usr/bin:$PATH
           echo $PATH
-          echo "Note the following is to fix a bug in Apple's python 3.7.3"
-          sudo sed -i.bak 's/sys.modules.values()/list(sys.modules.values())/g' /Applications/Xcode.app/Contents/Developer/Library/Frameworks/Python3.framework/Versions/3.7/lib/python3.7/linecache.py
           python3 --version
           python3 -c 'import site; print(site.getsitepackages())'
           python3 .github/workflows/build.instruction.py --sudo=true README.md "#### macOS" > source.sh
--- a/README.md
+++ b/README.md
@@ -180,20 +180,25 @@ $ bazel run //tools/lint:lint -- black pyupgrade --  tensorflow_io/core/python/o
 
 #### macOS
 
-On macOS Catalina or higher, it is possible to build tensorflow-io with
-system provided python 3 (3.7.3). Both `tensorflow` and `bazel` are needed.
+On macOS Catalina 10.15.7, it is possible to build tensorflow-io with
+system provided python 3.8.2. Both `tensorflow` and `bazel` are needed.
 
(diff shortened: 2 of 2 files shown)


kvignesh1420 (commented):
@yongtang here are the things I have observed:

- Since you added `/usr/bin` to `$PATH` (`export PATH=/usr/bin:$PATH`) the github action is using python `3.8.2`. However, the one readily available is `3.8.6` in `/usr/local/bin`. 
```console
+ which python3
/usr/local/bin/python3
+ python3 --version
+ python3 -c 'import site; print(site.getsitepackages())'
Python 3.8.6
```
Based on https://github.com/tensorflow/io/pull/1167.

- Also, I am not sure why the `requirements.txt` modification is required, as this PR: https://github.com/tensorflow/io/pull/1167 passes without those changes. Is this because users have python3.7 as system default as of now? Also, this change might affect users trying to install the lint dependencies while using python3.8. 

Please let me know if I am missing something here. Thanks.

yongtang:
@kvignesh1420 As mentioned in https://github.com/tensorflow/io/pull/1167#issuecomment-719948844 the system default python3 (comes with macOS itself, not with Python.org download, brew, or pyenv, etc) is 3.8.2.

In GitHub Actions machines, brew was used to install additional versions of python3. So that is why you see `/usr/local/bin/python3` as 3.8.6 (through brew), and `/usr/bin/python3` as 3.8.2 (macOS system default).

If your macOS machine only comes from a clean OS install (no brew), then you will notice there is only `/usr/bin/python3` (no `/usr/local/bin/python3` existence)

Since there are too many ways to install python3, our README.md is trying to stay with system default python (user does not need brew or download Python3 from python.org). If instead we support python3 with brew install, users may very well ask for additional support with other python installations as well (e.g., download from python.org). We want to avoid ending up testing too many scenarios.

Thus `export PATH=/usr/bin:$PATH` is needed to bring system default python back on GitHub Action's machines.

Now there is a small glitch on system default python 3.8.2: In order to make our lint work, bazel will try to install `regex` package which needs to compile C/C++ code.

However, as of today the latest clang with XCode will try to build with options of `-arch arm64 -arch x86_64`. Obviously this is Apple's pre-support for the upcoming `Apple Silicon` (ARM based processor) macBook that is expected to release mid-november. However, this will cause a bug similar to giampaolo/psutil#1832

So some additional fix to avoid installation of `regex` is needed. In the early attempt I added additional conditions in requirements.txt to skip the `regex` pip package.

After looking through giampaolo/psutil#1832 it looks like merely adding `export ARCHFLAGS="-arch x86_64"` will be enough.

Note this measure is only temporarily. I expect Apple's macOS 11.0 (Big Sur) will fully address the issue mentioned above. (I am guessing Big Sur will be released mid-november as well when Apple Silicon ARM macBook is released).

Please take a look at the updated PR and see if that helps.

kvignesh1420 (approved):
Thanks for the info @yongtang. It makes more sense now. I will close  https://github.com/tensorflow/io/pull/1167 in favour of this.

Issue #1156, opened 2020-10-20
tensorflow-io 0.16.0 release


yongtang:
With TensorFlow 2.4 release candidate out soon (this week or next), we can consider having a v0.16.0 release for 2.3, as was discussed in our last meeting.

Here is the list of items for 0.16.0:

- [x] Bump tensorflow-io version to 0.16.0 (PR #1157)
- [x] Update RELEASE.md to capture 0.16.0 changes (PR #1157)
- [x] Update README.md to update 0.16.0 release date (PR #1159)
- [x] When GitHub Actions complete the build against the master, download the artifact with the name tensorflow-io-release. Uncompress the content and all binaries for release are inside.
- [x] Do minimal sanity test to make sure at least the first section of getting started example code works locally.
- [x]  Push all binaries to PyPI.org.
- [x] Address the points related to docker releases in https://github.com/tensorflow/io/pull/1153#issuecomment-713163786
- [x] Add an overview description to our [docker releases](https://hub.docker.com/r/tfsigio/tfio).  Currently, it is blank.

And follow up immediately after 0.16.0 release:
- [x] Update docs/tutorials/colorspace.ipynb to use pip install tensorflow-io (instead of tensorflow-io-nightly). (PR: https://github.com/tensorflow/io/pull/1160)
- [x] Update docs/tutorials/kafka.ipynb to use pip install tensorflow-io (instead of tensorflow-io-nightly). (PR: https://github.com/tensorflow/io/pull/1160)
- [x] Update docs/tutorials/elasticsearch.ipynb to use pip install tensorflow-io (instead of tensorflow-io-nightly). (PR: https://github.com/tensorflow/io/pull/1160)
- [x] Push to tensorflow.org/io (with the help from docs team).

Finally, post 0.16.0 release:
- [x] Bump TF version to 2.4.0 (rc) (PR #1143)
- [x] Resolve any issues in #1143 to use modular file systems C API (as discussed in our last meeting). (PR #1143)
        Note since tensorflow/tensorflow#43680 has been merged, we are finally unblocked.
- [x] Send email to SIG IO

/cc @vlasenkoalexey @terrytangyuan @kvignesh1420 @BryanCutler @dmitrievanthony

yongtang:
@kvignesh1420 Do you have a PyPI.org account? If you do I can add you to the maintainer of tensorflow-io repo https://pypi.org/project/tensorflow-io/ if you prefer.

kvignesh1420:
[...]

PR #1172, opened 2020-11-04
[notebooks] use saved output to prevent site builder issues for elasticsearch.ipynb

kvignesh1420:
This PR addresses the comment: https://github.com/tensorflow/io/issues/1156#issuecomment-721905578 by using the saved outputs to prevent docs builder issues.

Issue reference: https://gist.github.com/MarkDaoust/bd7d027969b0d24b506e690a8d03b7a6#file-elastic-search-new-fail-L111


MarkDaoust (changes requested):
Test-running this version in the site-builder environment still gives me the same error. 

It works in Colab. 

So I think at this point we should just save the outputs in this notebook so the site builder doesn't need to execute it.

kvignesh1420:
@MarkDaoust thanks for the tests. Can you point me to any reference notebook or any documentation which saves the outputs of cells in the notebook itself? I am not sure about the practices around these fixes. Please let me know.

MarkDaoust:
I just mean open it with colab, and select "runtime > Run all" then when it's done "file > download ipynb", and replace the notebook with the resulting file.

After that the notebook will have "output"s attached to to each code cell.

Here's an example of a file saved that way:

https://github.com/tensorflow/tensorflow/blob/master/tensorflow/lite/g3doc/performance/post_training_float16_quant.ipynb
https://colab.sandbox.google.com/github/tensorflow/tensorflow/blob/master/tensorflow/lite/g3doc/performance/post_training_float16_quant.ipynb

tensorflow.org shows the saved outputs instead of executing the notebook.

https://www.tensorflow.org/lite/performance/post_training_float16_quant

kvignesh1420:
@MarkDaoust Thanks for the info. I ran all the cells and committed the changes. Updated the PR title and description as well. Please take a look.

MarkDaoust (approved):
Thanks!

PR #1182, opened 2020-11-08
[fs tests] refactor and extend test cases

kvignesh1420:
This PR:
- Refactors and extends the test cases for the HTTP File System related functionality.
- Fixes the docstrings for the AZFS test cases.
- Switches to compatible tf public api's for AZFS and HTTP FS tests.

### DIFF
--- a/tests/test_azure.py
+++ b/tests/test_azure.py
@@ -40,11 +40,11 @@ class AZFSTest(tf.test.TestCase):
 
     def setUp(self):  # pylint: disable=invalid-name
         super().setUp()
-        if not tf.io.gfile.IsDirectory(self.path_root):
-            tf.io.gfile.MakeDirs(self.path_root)
+        if not tf.io.gfile.isdir(self.path_root):
+            tf.io.gfile.makedirs(self.path_root)
 
     def test_exists(self):
-        self.assertTrue(tf.io.gfile.IsDirectory(self.path_root))
+        self.assertTrue(tf.io.gfile.isdir(self.path_root))
 
     def test_also_works_with_full_dns_name(self):
         """Test the file system also works when we're given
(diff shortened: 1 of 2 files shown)


tests/test_http_eager.py
```diff
@@ -15,38 +15,72 @@
...
-        i += 1
-    assert i == len(lines)
-
-    remote = tf.io.gfile.GFile(filename)
-    assert remote.size() == len(local)
-    offset_start = list(range(0, len(local), 100))
-    offset_stop = offset_start[1:] + [len(local)]
-    for (start, stop) in zip(offset_start, offset_stop):
-        assert remote.read(100) == local[start:stop]
+class HTTPFSTest(test.TestCase):
```
yongtang, line 23: It looks like `tf.test.TestCase` is public API. I think we can use that directly like below?
```
class HTTPFSTest(tf.test.TestCase):
```


(The `from tensorflow.python.platform import test` is not publicly exposed endpoint so they might be subject to change in the future.

tests/test_azure.py
```diff
@@ -26,9 +26,9 @@
 class AZFSTest(test.TestCase):
```
yongtang, line 26: That was from the past code but I think we can use `tf.test.TestCase` as well.

Also, `tf.io.gfile` can be used directly. So at the beginning of the file:
```
from tensorflow.python.platform import test
from tensorflow.python.platform import gfile
```

can be replaced with:
```
import tensorflow as tf
```

tests/test_http_eager.py
kvignesh1420: addressed in the commit

tests/test_azure.py
kvignesh1420: addressed in the commit

kvignesh1420:
@yongtang I have observed that the public API for tf.io.gfile has changed a bit.

For example:

in current test_azure.py
```python
def setUp(self):  # pylint: disable=invalid-name
        super().setUp()
        if not tf.io.gfile.IsDirectory(self.path_root):
            tf.io.gfile.MakeDirs(self.path_root)
```

But, the prompts for the API's are a bit different:

image

yongtang:
@kvignesh1420 The APIs that are listed on tensorflow.org's api_docs (e.g., https://www.tensorflow.org/api_docs/python/tf/io/gfile/isdir) are considered as `public API` and have the guarantee of API-compatibility.

Alternatively, the symbols inside tensorflow's code base with `@tf_export("...")` decorated are considered as `public API` (tensorflow.org only extracts APIs defined with `@tf_export("...")`):
https://github.com/tensorflow/tensorflow/blob/fcc4b966f1265f466e82617020af93670141b009/tensorflow/python/lib/io/file_io.py#L499

Note in the code base you may notice some API specified with `v1` (e.g., `@tf_export(v1=["gfile.Copy"])`):
https://github.com/tensorflow/tensorflow/blob/v2.3.1/tensorflow/python/lib/io/file_io.py#L483

Those APIs are only for TF 1.x and will not be exposed with TF 2.0.

yongtang:
Relevant to this PR we should use the ones that are exposed, i.e., instead of `tf.io.gfile.MakeDirs`, we should use `tf.io.gfile.makedirs` (public APIs), as `tf.io.gfile.MakeDirs` does not guarantee no future changes.

kvignesh1420:
@yongtang thanks for the info. Made the applicable changes. PTAL.

yongtang (approved):

Issue #1183, opened 2020-11-08
Migration of modular file system plugins in tensorflow (s3/hdfs/gcs) to tensorflow-io


yongtang:
As was discussed in our September monthly meetings, we are looking into migrate modular file system plugins in tensorflow (s3/hdfs/gcs) to tensorflow-io package.

Once the migration is done the size and build time of tensorflow repo could be substantially improved, this can greatly help external tensorflow contributors experience as they could selectively decide to only build some components (e.g., tensorflow core, tensorflow-io, tensorflow-addons, etc).

Currently it takes at least 8 hours (or even more) for many external tensorflow contributors to build complete tensorflow package due to the lack of bazel cache that is only internally available in google.

Now as we already have a couple of modular file systems in tensorflow-io (azure blob file system and http file system. see #1111), it might be time to start looking into migration of plugins in tensorflow to tensorflow-io.

/cc @mihaimaruseac @vnvo2409

vnvo2409:
Let me check if there is any incompatible between the core implementation and the plugins.

There are some points which I would like to make it clear.

---

Documentation: There are some issues of TF related to the behavior of the environment variables: 
- `S3`:  https://github.com/tensorflow/tensorflow/issues/44564 https://github.com/tensorflow/tensorflow/issues/42389 
- `GCS`: https://github.com/tensorflow/tensorflow/issues/15530

Up till now, I don't see much documentation about those filesystems. I would like to have a working example and documentation about other envs. Where should I make a PR to ?

---

Issues: Could we transfer the issues related to those filesystems from TF repo to this repo ?

---

Testing: What we should do about testing ? I think we could set up a `s3` and `gcs` emulator in `github actions`. Another possibility is tagging those tests `manual` and run it locally.

yongtang:
Thanks @vnvo2409 !

For testing: I have created a PR #1184 which setup the AWS Localstack emulator in GitHub Actions. See https://github.com/tensorflow/io/blob/master/tests/test_aws/aws_test.sh for localstack emulator setup.

The test is working for Kinesis dataset in tensorflow-io. I configured the localstack emulator to have s3 and a bare-bone stub test has been created as well as a starting. Please take a look (https://github.com/tensorflow/io/blob/master/tests/test_s3_eager.py)

I will take a look at gcs emulator and see if a similar setup could be done as well.

For issue transfer: Will take a look to see if issue can be transferred here.

[...]

PR #1184, opened 2020-11-09
Re-enable AWS Localstack emulator and add a s3 stub test

yongtang:
The AWS Localstack setup was disabled at one point due to the API endpoint port change in Localstack (4568 => 4566).

Now as we are looking into AWS S3 file systems, this PR re-enables localstack setup so that we can test AWS services through this emulator.

This PR also add a stub s3 test case so that we can continue working on s3 file system.

This PR is a prerequisite of #1183.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -182,7 +182,7 @@ jobs:
         python: ['3.6', '3.7']
     steps:
       - uses: actions/checkout@v2
-      - uses: docker-practice/actions-setup-docker@master
+      - uses: docker-practice/actions-setup-docker@v1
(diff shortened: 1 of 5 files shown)


.github/workflows/build.yml
```diff
@@ -182,6 +182,7 @@ jobs:
         python: ['3.6', '3.7']
     steps:
       - uses: actions/checkout@v2
+      - uses: docker-practice/actions-setup-docker@master
```
terrytangyuan, line 185: Should we pin a version here?

terrytangyuan (approved):

PR #1186, opened 2020-11-11
Migrated curl_http_request.[h|cc] so that HTTP file system to not depends on TensorFlow C++ API

yongtang:
This PR migrated curl_http_request.[h|cc] in tensorflow to tensorflow-io, so that the HTTP file system does not depends on TensorFlow C++ APIs anymore. With this change, it should be possible to load the plugin so file in any future versions of tensorflow and removes the version match requirement.

### DIFF
--- a/tensorflow_io/core/kernels/http_kernels.cc
+++ b/tensorflow_io/core/kernels/http_kernels.cc
@@ -125,7 +125,7 @@ uint64_t Length(const TF_ReadOnlyMemoryRegion* region) { return 0; }
 
 // SECTION 4. Implementation for `TF_Filesystem`, the actual filesystem
 // ----------------------------------------------------------------------------
-namespace tf_azfs_filesystem {
+namespace tf_http_filesystem {
 
 static void Init(TF_Filesystem* filesystem, TF_Status* status) {
   TF_SetStatus(status, TF_OK, "");
@@ -275,7 +275,7 @@ static char* TranslateName(const TF_Filesystem* filesystem, const char* uri) {
   return strdup(uri);
 }
 
-}  // namespace tf_azfs_filesystem
+}  // namespace tf_http_filesystem
 
 }  // namespace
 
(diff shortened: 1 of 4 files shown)


tensorflow_io/core/kernels/http_kernels.cc
```diff
@@ -13,16 +13,527 @@ See the License for the specific language governing permissions and
...
+// Set to 1 to enable verbose debug output from curl.
+constexpr uint64_t kVerboseOutput = 0;
+
+static absl::Mutex mu;
+static bool uninitialized(false);
+void CurlInitialize() {
+  absl::MutexLock l(&mu);
+  if (!uninitialized) {
+    curl_global_init(CURL_GLOBAL_ALL);
+    uninitialized = true;
```
kvignesh1420, line 41: @yongtang I think using `initialized` instead of `uninitialized` is more relatable.

```cpp
static absl::Mutex mu;
static bool initialized(false);
void CurlInitialize() {
  absl::MutexLock l(&mu);
  if (!initialized) {
    curl_global_init(CURL_GLOBAL_ALL);
    initialized = true;
  }
}
```

tensorflow_io/core/kernels/http_kernels.cc
```diff
@@ -13,16 +13,527 @@ See the License for the specific language governing permissions and
...
+      return;
+    }
+    if ((s = curl_easy_setopt(curl_, CURLOPT_WRITEFUNCTION,
+                              &CurlHttpRequest::WriteCallbackDirect)) !=
+        CURLE_OK) {
+      std::string error_message =
+          absl::StrCat("Unable to set CURLOPT_WRITEDATA: ", s);
+      TF_SetStatus(status, TF_INTERNAL, error_message.c_str());
+      return;
+    }
```
kvignesh1420, line 195: @yongtang these two conditions are covered in `Initialize()`. Any specific reason for adding here again?

tensorflow_io/core/kernels/http_kernels.cc
```diff
@@ -13,16 +13,527 @@ See the License for the specific language governing permissions and
...
+      const auto connect_time_status =
+          curl_easy_getinfo(that->curl_, CURLINFO_CONNECT_TIME, &connect_time);
+
+      double pretransfer_time = -1;
+      const auto pretransfer_time_status = curl_easy_getinfo(
+          that->curl_, CURLINFO_PRETRANSFER_TIME, &pretransfer_time);
+
+      double starttransfer_time = -1;
+      const auto starttransfer_time_status = curl_easy_getinfo(
+          that->curl_, CURLINFO_PRETRANSFER_TIME, &starttransfer_time);
```
kvignesh1420, line 469: should be `CURLINFO_STARTTRANSFER_TIME` ?

tensorflow_io/core/kernels/http_kernels.cc
yongtang: @kvignesh1420 Updated. Thanks.

tensorflow_io/core/kernels/http_kernels.cc
```diff
@@ -13,16 +13,527 @@ See the License for the specific language governing permissions and
...
+      std::string error_message =
+          absl::StrCat("Unable to set CURLOPT_WRITEDATA: ", s);
+      TF_SetStatus(status, TF_INTERNAL, error_message.c_str());
+      return;
+    }
+    if ((s = curl_easy_setopt(curl_, CURLOPT_WRITEFUNCTION,
+                              &CurlHttpRequest::WriteCallbackDirect)) !=
+        CURLE_OK) {
+      std::string error_message =
+          absl::StrCat("Unable to set CURLOPT_WRITEDATA: ", s);
```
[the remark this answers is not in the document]
yongtang: Updated.

tensorflow_io/core/kernels/http_kernels.cc
yongtang: Thanks. Updated.

tensorflow_io/core/kernels/http_kernels.cc
yongtang: @kvignesh1420 There are two cases. One is to set the result buffer with caller provided buffer (`SetResultBufferDirect`). Another is the non-caller provided buffer. To make it clear I have split out the part with new function `SetResultBuffer`.

yongtang:
Thanks @kvignesh1420 for the review. The PR has been updated. Please take a look.

kvignesh1420 (approved):
LGTM.

kvignesh1420:
@yongtang now that we have the http support in place. I will be making the switch for elasticsearch's use of curl_http_request as well.

PR #1188, opened 2020-11-11
Re-organize the structure of the az and http plugins

yongtang:
This PR re-organize the structure of the az and http plugins
so that it is easy to manage. Now each file system is on its own
directory `tensorflow_io/core/plugins/xxx`. This is for the upcoming
s3/gcs/hdfs plugins migration.


kvignesh1420 (approved):
@yongtang

terrytangyuan (approved):

PR #1189, opened 2020-11-12
Bump AWS CPP SDK to 1.7.336 

yongtang:
This PR bumpds AWS CPP SDK to 1.7.336, so that it matches the verison
in tensorflow (and in preparation of upcoming s3 file system migration).

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -382,11 +382,11 @@ http_archive(
 http_archive(
     name = "aws-sdk-cpp",
     build_file = "//third_party:aws-sdk-cpp.BUILD",
-    sha256 = "8724a2c3f00478e5f9af00d1d9bc67b7a0a557cc587a10f7097b1257008c2e68",
-    strip_prefix = "aws-sdk-cpp-1.7.270",
+    sha256 = "758174f9788fed6cc1e266bcecb20bf738bd5ef1c3d646131c9ed15c2d6c5720",
+    strip_prefix = "aws-sdk-cpp-1.7.336",
     urls = [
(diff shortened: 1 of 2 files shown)


WORKSPACE
```diff
@@ -382,11 +382,14 @@ http_archive(
 http_archive(
     name = "aws-sdk-cpp",
     build_file = "//third_party:aws-sdk-cpp.BUILD",
-    sha256 = "8724a2c3f00478e5f9af00d1d9bc67b7a0a557cc587a10f7097b1257008c2e68",
-    strip_prefix = "aws-sdk-cpp-1.7.270",
+    patch_cmds = [
+        """sed -i.bak 's/UUID::RandomUUID/Aws::Utils::UUID::RandomUUID/g' aws-cpp-sdk-core/source/client/AWSClient.cpp""",
```
kvignesh1420, line 386: @yongtang any reason for this replacement?

yongtang: @kvignesh1420 On Windows UUID is a defined symbol so a build failure will occur:
```
external/aws-sdk-cpp/aws-cpp-sdk-core/source/client/AWSClient.cpp(215): error C2872: 'UUID': ambiguous symbol
```

As such we have to declare `Aws::Utils::UUID::RandomUUID` in full namespace (not starting with `UUID::RandomUUID`)

kvignesh1420 (approved):
LGTM.

WORKSPACE
kvignesh1420: Ok

PR #1190, opened 2020-11-12
[fs plugins] modify pre-processor directives

kvignesh1420:
This PR modifies the pre-processor directives of the file_system_plugins as per the new directory structure.

### DIFF
--- a/tensorflow_io/core/plugins/file_system_plugins.h
+++ b/tensorflow_io/core/plugins/file_system_plugins.h
@@ -13,8 +13,8 @@ See the License for the specific language governing permissions and
 limitations under the License.
 ==============================================================================*/
 
-#ifndef TENSORFLOW_IO_CORE_KERNELS_FILE_SYSTEM_PLUGINS_H
-#define TENSORFLOW_IO_CORE_KERNELS_FILE_SYSTEM_PLUGINS_H
+#ifndef TENSORFLOW_IO_CORE_PLUGINS_FILE_SYSTEM_PLUGINS_H
+#define TENSORFLOW_IO_CORE_PLUGINS_FILE_SYSTEM_PLUGINS_H
 
 #include <stdlib.h>
 
--- a/tensorflow_io/core/plugins/file_system_plugins.h
+++ b/tensorflow_io/core/plugins/file_system_plugins.h
@@ -41,4 +41,4 @@ void ProvideFilesystemSupportFor(TF_FilesystemPluginOps* ops, const char* uri);
 }  // namespace io
 }  // namespace tensorflow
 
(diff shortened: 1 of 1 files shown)


kvignesh1420:
@yongtang I have observed that you have defined the `CurlHttpRequest` class within `http_file_system.cc` file itself. Can I move the `CurlHttpRequest` definitions to a different header file so that other kernels can also use it? 
If there is a better solution, please let me know.

tensorflow_io/core/plugins/file_system_plugins.h
```diff
@@ -13,8 +13,8 @@ See the License for the specific language governing permissions and
 limitations under the License.
 ==============================================================================*/
 
-#ifndef TENSORFLOW_IO_CORE_KERNELS_FILE_SYSTEM_PLUGINS_H
-#define TENSORFLOW_IO_CORE_KERNELS_FILE_SYSTEM_PLUGINS_H
+#ifndef TENSORFLOW_IO_CORE_PLUGINS_FILE_SYSTEM_PLUGINS_H
```
yongtang, line 16: Since we are here, maybe we can also update the very last line of the file as well?
```
#endif  // TENSORFLOW_IO_CORE_KERNELS_FILE_SYSTEM_PLUGINS_H
```

kvignesh1420: @yongtang done !

yongtang:
@kvignesh1420 We can always split the `CurlHttpRequest` into a separate lib so that it could be used by different components (e.g., with kernels or plugins).

We can also continue to use TensorFlow C++ APIs in kernels as well, as they are built in different shared objects so it will not impact for now:
- `libtensorflow_io_plugins.so` for plugins and using TensorFlow C API only (cross-version compatible)
- `libtensorflow_io.so` for kernels and using TensorFlow C++ API (**not** cross-version compatible)

At some point we may want to build `libtensorflow_io.so` with C API as well. Though I would suggest defer converting kernels to use C APIs. There are several effort going on in tensorflow teams (one from the TF Runtime and another one from TF MLIR) to change kernels implementations in tensorflow core. I think we may want to wait until the future kernel implementations in Tensorflow core has been finalized before adopting out version.

kvignesh1420:
@yongtang If this is the case, then it would be better to keep the plugins isolated and continue to use tensorflow's curl_http_request as of now for the elasticsearch kernel.

yongtang:
> @yongtang If this is the case, then it would be better to keep the plugins isolated and continue to use tensorflow's curl_http_request as of now for the elasticsearch kernel.

Sounds good to me 👍

yongtang (approved):

PR #1195, opened 2020-11-16
Add ffmpeg support for Ubuntu 20.04 and switch to python 3.8 for tests

yongtang:
This PR does two things:

1) Add ffmpeg support for Ubuntu 20.04
   In the past only Ubuntu 16.04 and 18.04 are supported for ffmpeg audio/video Dataset.
   In this PR ffmpeg support for Ubuntu 20.04 has been added as Ubuntu 20.04 has been released this year.
   Also fix `Invalid return value 0 for stream protocol` issue when LIBAVFORMAT_VERSION_MAJOR > 57

2) Switch to python 3.8 for tests
   In the past we use python 3.6/3.7 as the default version for testing as some packages for testing are not available for 3.8 yet, even though many python systems have been upgraded to 3.8. For example, Ubuntu 20.04 and macOS's default system pythons are already 3.8.
   This PR switches to use python 3.8 for testings. This should also reduce GitHub Actions jobs a little bit.
   This PR also updates ffmpeg video samples (as there is a small different between Ubuntu 18.04 and Ubuntu 20.04 ffmpeg)


.github/workflows/build.yml
```diff
@@ -179,7 +179,7 @@ jobs:
     runs-on: macos-latest
     strategy:
       matrix:
-        python: ['3.6', '3.7']
+        python: ['3.8']
```
kvignesh1420, line 182: @yongtang maybe we can test with 3.7 and 3.8? Just to be sure of the compatibility as users might still be using 3.7.

kvignesh1420 (approved):

terrytangyuan (approved):

PR #1197, opened 2020-11-17
Initial commit of hadoop file system (credit vnvo2409)

yongtang:
This PR is an initial commit of Hadoop modular file system that was moved from tensorflow to here. Credit goes to author @vnvo2409.

This PR consists of mostly mechanic changes, though there are some small change to create a wrapper so that the loading of libhdfs.so will only be done lazily (not loaded when file system is initialized but loaded at the first usage). This avoids the case of the following error in case user does not even want to use hdfs:
```
Library (libhdfs.dylib) not found: dlopen(libhdfs.dylib, 6): image not found']
```

It is a little challenging to setup Hadoop for tests so this PR defers. Will create a follow up PR to set up the tests.


tensorflow_io/core/plugins/hdfs/hadoop_filesystem.cc
```diff
@@ -0,0 +1,919 @@
...
+
+HadoopFileSystemImplementation::HadoopFileSystemImplementation(
+    TF_Status* status)
+    : libhdfs(new LibHDFS(status)),
+      connection_cache_lock(),
+      connection_cache() {}
+
+typedef struct HadoopFileSystem {
+  absl::Mutex mu;
+  std::unique_ptr<HadoopFileSystemImplementation> ptr ABSL_GUARDED_BY(mu);
```
vnvo2409, line 460: I think the test failed because you forgot to add `#include<memory>`

tensorflow_io/core/plugins/hdfs/hadoop_filesystem.cc
```diff
@@ -0,0 +1,919 @@
...
+  TF_SetStatus(status, TF_OK, "");
+  return symbol;
+#endif
+}
+
+// Implementation of a filesystem for HADOOP environments.
+// This filesystem will support `hdfs://`, `viewfs://` and `har://` URI schemes.
+
+static void* plugin_memory_allocate(size_t size) { return calloc(1, size); }
+static void plugin_memory_free(void* ptr) { free(ptr); }
```
vnvo2409, line 87: These two functions are unused. Likely because I copied from `posix`. Let just merge this PR and I will make a follow up PR ( mostly cleanup and better memory management ).

tensorflow_io/core/plugins/hdfs/hadoop_filesystem.cc
```diff
@@ -0,0 +1,919 @@
...
+
+namespace tensorflow {
+namespace io {
+namespace hdfs {
+
+void* LoadSharedLibrary(const char* library_filename, TF_Status* status) {
+#if defined(_MSC_VER)
+  TF_SetStatus(status, TF_UNIMPLEMENTED, "LoadSharedLibrary not implemented");
+  return nullptr;
+#else
```
vnvo2409, line 47: Is there any reason not to use `TF_LoadSharedLibrary` as we did in experiemental plugin. And does it mean that we drop support for `hdfs` on Windows ?

tensorflow_io/core/plugins/hdfs/hadoop_filesystem.cc
```diff
@@ -0,0 +1,919 @@
...
+    constexpr char kLibHdfsDso[] = "libhdfs.so";
+#endif
+    char* hdfs_home = getenv("HADOOP_HDFS_HOME");
+    if (hdfs_home != nullptr) {
+      auto JoinPath = [](std::string home, std::string lib) {
+#if defined(_WIN32)
+        if (home.back() != '\\') home.push_back('\\');
+        return home + "lib\\native\\" + lib;
+#else
```
vnvo2409, line 207: If we drop support for `hdfs` on Windows, I think this isn't necessary anymore.

vnvo2409:
Thank you for this PR @yongtang !

In addition, I am thinking of a loading dynamically mechanics for the plugins. Maybe we could change 
https://github.com/tensorflow/tensorflow/blob/fa6c1bb0fab42d153fa375ad31713fbda99aba0c/tensorflow/core/platform/env.cc#L112-L120
```cpp
Status Env::GetFileSystemForFile(const std::string& fname,
                                 FileSystem** result) {
  StringPiece scheme, host, path;
  io::ParseURI(fname, &scheme, &host, &path);
  FileSystem* file_system = file_system_registry_->Lookup(std::string(scheme));
  if (!file_system) {
    if (scheme.empty()) {
      scheme = "[local]";
    }

    return errors::Unimplemented("File system scheme '", scheme,
                                 "' not implemented (file: '", fname, "')");
  }
  *result = file_system;
  return Status::OK();
}
```
`Lookup` to `LookupOrLoad` so we don't have to load all the plugins into TF in the first place.

tensorflow_io/core/plugins/hdfs/hadoop_filesystem.cc
yongtang: @vnvo2409 Thanks. The PR has been updated.

tensorflow_io/core/plugins/hdfs/hadoop_filesystem.cc
yongtang: @vnvo2409 Inside tensorflow while the functions are defined, they somehow are not part of the pip package so we cannot include them directly. I tried to modify the tensorflow repo but had some trouble due to the big circular dependency. Give we only need those two functions and they are small, I think it makes sense to just include them here.

Also, I have added the Windows counterpart as well so now hdfs should be supported on Windows.

tensorflow_io/core/plugins/hdfs/hadoop_filesystem.cc
yongtang: @vnvo2409 See my other comments. The Windows part has been added.

yongtang:
Thanks @vnvo2409 for the review. The PR has been updated now to address the review feedback.

vnvo2409:
LGTM. Thank you :smile:

PR #1200, opened 2020-11-18
Use system default python3 for bazel building on macOS

yongtang:
GitHub Actions recently changed the python3 version on macOS to 3.9.
However, as tensorflow does not have 3.9 support yet, this is failing
our build.

This PR changes to use system default python3 instead.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -117,6 +117,10 @@ jobs:
       - name: Bazel on macOS
         run: |
           set -x -e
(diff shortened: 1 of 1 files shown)


kvignesh1420:
@yongtang the change from 3.8 to 3.9 was pretty fast!

Reference: [macos-latest](https://github.com/actions/virtual-environments/blob/main/images/macos/macos-11.0-Readme.md)

kvignesh1420 (approved):

PR #1202, opened 2020-11-18
Convert to C API logging

yongtang:
This PR convert to C API logging in modular file system plugins. The `tensorflow/c/logging.h` is already available, however, the library didn't link the API (will be fixed by https://github.com/tensorflow/tensorflow/pull/44993).

For the moment this PR converts the C++ `LOG` call which will be needed very soon anyway.

### DIFF
--- a/tensorflow_io/core/plugins/az/az_file_system.cc
+++ b/tensorflow_io/core/plugins/az/az_file_system.cc
@@ -31,6 +31,12 @@ limitations under the License.
 #include "storage_account.h"
 #include "storage_credential.h"
 #include "storage_errno.h"
+// TODO: Restore logging.h
+#define TF_Log(...)
+#define TF_VLog(...)
+// #include "tensorflow/c/logging.h"
+// TODO: Restore logging.h
+#include "tensorflow/c/tf_status.h"
 #include "tensorflow_io/core/plugins/file_system_plugins.h"
 
 namespace tensorflow {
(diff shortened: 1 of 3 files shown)


tensorflow_io/core/plugins/az/az_file_system.cc
```diff
@@ -31,6 +31,12 @@ limitations under the License.
 #include "storage_account.h"
 #include "storage_credential.h"
 #include "storage_errno.h"
+// TODO: Restore logging.h
+#define TF_Log(...)
+#define TF_VLog(...)
```
kvignesh1420, line 36: @yongtang the `TF_VLog` macro isn't being used here, is this just to bring in the existing logging macro's ?

terrytangyuan (approved):
LGTM

tensorflow_io/core/plugins/az/az_file_system.cc
yongtang: @kvignesh1420 The `TF_VLog` is part of the C logging API in tensorflow (https://github.com/tensorflow/tensorflow/blob/master/tensorflow/c/logging.h#L34-L36). Currently it is defined as empty as we need https://github.com/tensorflow/tensorflow/pull/44993 . Once the PR is merged those defines here can be safely removed.

kvignesh1420: okay!

kvignesh1420 (approved):
LGTM

PR #1203, opened 2020-11-18
Initial commit of gcs modular file system plugin (vnvo2409)

yongtang:
This is the initial commit of gcs modular file system plugin authored
by @vnvo2409. This is the first step with code moved here.

Several things that needs to be done in follow up PRs:
1. The thread-related functions was using TensorFlow APIs. However, they are not exposed in TF 2.4 so we may need to convert them here first.
2. The testing has not been touched yet. It will be covered in follow up PRs.

This PR is a first step, as the content is pretty big so split into steps.


tensorflow_io/core/plugins/gs/cleanup.h
```diff
@@ -0,0 +1,117 @@
...
+//     }
+//   }
+//
+// You can use Cleanup<F> directly, instead of using MakeCleanup and auto,
+// but there's rarely a reason to do that.
+//
+// You can call 'release()' on a Cleanup object to cancel the cleanup.
+
+#ifndef TENSORFLOW_C_EXPERIMENTAL_FILESYSTEM_PLUGINS_GCS_CLEANUP_H_
+#define TENSORFLOW_C_EXPERIMENTAL_FILESYSTEM_PLUGINS_GCS_CLEANUP_H_
```
kvignesh1420, line 43: @yongtang the guards can be renamed.

vnvo2409:
LGTM. Thank you !

tensorflow_io/core/plugins/gs/cleanup.h
yongtang: @kvignesh1420 Thanks. The PR has been updated with guard renamed.

kvignesh1420 (approved):

kvignesh1420:
@yongtang thank you for migrating the plugins. Also, credit to @vnvo2409 for writing the plugins.

PR #1206, opened 2020-11-24
Bump TF to 2.4.0rc3, and enable TF C API logging

yongtang:
This PR bumps TF to 2.4.0rc3, and enable TF C API logging


terrytangyuan (commented):
I remember there are some code that uses logging but temporarily commented out. Perhaps those can be incorporated into this PR as well?

yongtang:
@terrytangyuan Yes some of the logging has already been `pre-converted` to C logging Api (like 47a3c25). They have been uncommented before this PR (but not in effect due to the `#define TF_Log(...)`. This PR removed the `#define TF_Log(...)` so logging is in place now.

kvignesh1420 (approved):

terrytangyuan (approved):

PR #1207, opened 2020-11-25
Add support for mongoDB

kvignesh1420:
This PR adds support for `mongoDB` which is a document-oriented NoSQL database. 

The PR exposes the following API's:

- `tfio.experimental.mongodb.MongoDBWriter` for writing documents to mongo collections
- `tfio.experimental.mongodb.MongoDBIODataset` for reading documents from mongo collections.

In addition to writing messages to mongo, the `MongoDBWriter` also has a helper method to delete documents. (A wrapper around mongo's  `DeleteMany()` operation)


kvignesh1420:
@yongtang while building on Linux systems, I get the following error during `import tensorflow_io as tfio`

```console
NotImplementedError: unable to open file: /v/tensorflow_io/core/python/ops/__init__.py, from paths: ['/v/tensorflow_io/core/python/ops/libtensorflow_io.so', 'bazel-bin/tensorflow_io/core/python/ops/libtensorflow_io.so']
2020-11-25T20:00:54.2783446Z E   caused by: ['/v/tensorflow_io/core/python/ops/libtensorflow_io.so: cannot open shared object file: No such file or directory', 'bazel-bin/tensorflow_io/core/python/ops/libtensorflow_io.so: undefined symbol: __dn_expand'
```

and for macOS

```console
NotImplementedError: unable to open file: /Users/runner/work/io/io/tensorflow_io/core/python/ops/__init__.py, from paths: ['/Users/runner/work/io/io/tensorflow_io/core/python/ops/libtensorflow_io.so', 'bazel-bin/tensorflow_io/core/python/ops/libtensorflow_io.so']
2020-11-25T20:02:17.5867580Z E   caused by: ['dlopen(/Users/runner/work/io/io/tensorflow_io/core/python/ops/libtensorflow_io.so, 6): image not found', 'dlopen(bazel-bin/tensorflow_io/core/python/ops/libtensorflow_io.so, 6): Symbol not found: __mongoc_rand_bytes\n  Referenced from: bazel-bin/tensorflow_io/core/python/ops/libtensorflow_io.so\n  Expected in: flat namespace\n in bazel-bin/tensorflow_io/core/python/ops/libtensorflow_io.so'
```

Please let me know if you have any suggestions.

yongtang:
@kvignesh1420 It looks like the `_mongoc_rand_bytes` seems to be a file from mongoldb's source code? I will take a look.

kvignesh1420:
@yongtang did you build `openssl` using bazel before? Please let me know as SSL support is compulsory for mongo auth to function.

kvignesh1420:
**UPDATE:**  `libmongoc` needs either openssl/libressl for providing auth support while connecting to mongo. However, since we already have `boringssl` the symbols are conflicting with `openssl` build.

for example:

```console
duplicate symbol '_X509_free' in:
    bazel-out/darwin-fastbuild/bin/external/openssl/libcrypto.a(x_x509.o)
    bazel-out/darwin-fastbuild/bin/external/boringssl/libcrypto.a(x_x509.pic.o)
```

cc: @yongtang @terrytangyuan @BryanCutler @vlasenkoalexey

kvignesh1420:
**UPDATE:**

- The symbol conflict was addressed by building a separate `libtensorflow_io_mongodb.so` file for mongo ops. 
- Additionally, by referring to the `ffmpeg` ops, the centos7 build issue was addressed by loading the `.so` file only on demand.

yongtang:
Thanks @kvignesh1420 for the PR and sorry for the slow response, as last week was mostly holiday in us.

yongtang:
@kvignesh1420 I think we may want to either use openssl only (and remove boringssl), or use boringssl only. The reason we stayed with boringssl was because boringssl was used in TF core and we don't want to have a (slightly) different library when core and tensorflow-io's C++ symbols may interleave. (In the future if we can stay with TF C API then openssl will be an option, though we are not completely there yet.)

 I take a look at boringssl and found out there are several APIs in boringssl are missing comparing with openssl. I will take a look first and come back.

kvignesh1420:
@yongtang sure. As you said there are some missing API's in boringssl that are causing the build's to fail. Also, when I tried using only OpenSSL, the windows builds were failing. 

Thanks for helping with this issue.

yongtang:
@kvignesh1420 I give it a try. Looks like if you can apply the commit https://github.com/tensorflow/io/pull/1212/commits/8fd28e677b265e4a862d8c727747039ccaac731d (see #1212) on top of you PR the boringssl issue might be resolved. Would you like to give it a try?

kvignesh1420:
@yongtang sure. Will try it out and let you know

kvignesh1420:
@yongtang thanks for the help. The builds and test cases are now passing with `boringssl`.

yongtang (approved):
LGTM. This is a great added feature, thanks for the PR! Ideally a follow up PR for a tutorial would be nice.

kvignesh1420:
@yongtang yes, few follow up PR's are required for updating docs, extending test cases and tutorials.

PR #1208, opened 2020-11-25
Use lazy load to avoid loading libtensorflow_io.so until needed.

yongtang:
In tensorflow-io we have two shared libraries:
- `libtensorflow_io.so` is linked with TF C++ API and does not have forward compatibility
- `libtensorflow_io_plugins.so` is linked with TF C API and does have forward compatibility

The plan is to eventually have libtensorflow_io.so linked to TF C API only
but that will take a long time and requires rewriting C++ kernel (after modular kernel API is available in TF, unknown status now).

With both `libtensorflow_io.so` and `libtensorflow_plugins.so` loaded in `import tensorflow_io as tfio`,
the tensorflow-io package is not forward compatible even though our file systems implementations are forward compatible already.

In the meantime, one way to allow forward compatible for file systems, is to lazy load `libtensorflow_io.so`:
- When user tries to run some function like `decode_json` they still encounter errors if they have newer versions of TF.
- However, thwn user tries to just use the file system feature (e.g., azfs/http/etc), it works with newer versions of TF with tensorflow-io through `import tensorflow_io as tfio` (not touching any other submodules in `tensorflow_io`).

This PR is part of the #1111.

### DIFF
--- a/tensorflow_io/core/python/api/experimental/__init__.py
+++ b/tensorflow_io/core/python/api/experimental/__init__.py
@@ -28,3 +28,9 @@ from tensorflow_io.core.python.api.experimental import audio
 from tensorflow_io.core.python.api.experimental import streaming
 from tensorflow_io.core.python.api.experimental import filter
 from tensorflow_io.core.python.api.experimental import elasticsearch
+
+
+def oss():
+    from tensorflow_io.core.python.ops import core_ops as _core_ops
+
+    return _core_ops._load()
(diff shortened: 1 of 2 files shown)


yongtang:
/cc @terrytangyuan I exposed a `tfio.exeprimental.oss` which can be called through:
```
tfio.exeprimental.oss()
```
to invoke OSS support explicitly. This can help us go through transition until OSS is converted to C file system API.

terrytangyuan (approved):
Thanks! Sounds good.

PR #1218, opened 2020-12-09
Add missed function RecursivelyCreateDir for hdfs file system implementation

yongtang:
This PR adds missed function RecursivelyCreateDir for hdfs file system implementation

### DIFF
--- a/tensorflow_io/core/plugins/hdfs/hadoop_filesystem.cc
+++ b/tensorflow_io/core/plugins/hdfs/hadoop_filesystem.cc
@@ -774,6 +774,11 @@ void CreateDir(const TF_Filesystem* filesystem, const char* path,
     TF_SetStatus(status, TF_OK, "");
 }
 
+static void RecursivelyCreateDir(const TF_Filesystem* filesystem,
+                                 const char* path, TF_Status* status) {
+  CreateDir(filesystem, path, status);
+}
+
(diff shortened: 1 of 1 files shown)


terrytangyuan (approved):

kvignesh1420 (approved):

vnvo2409:
@yongtang 
I am not sure it is a good idea. AFAIK, hadoop behaves exactly like posix in term of [directory](https://stackoverflow.com/questions/40143528/hdfs-dfs-mkdir-no-such-file-or-directory). So in the current plugin implementation, I let it fall back to the default [implementation](https://github.com/tensorflow/tensorflow/blob/2f7251d79b9562aeb5f8cab30a5848fd03120637/tensorflow/c/experimental/filesystem/modular_filesystem.cc#L264-L274) of TensorFlow. Please let me know if I am missing anything.

yongtang:
@vnvo2409 Without RecursivelyCreateDir explicitly added I encountered the issue of `RecursivelyCreateDir` will try to call itself and eventually running out of stack (tested with gdb). I think that might be caused by the `RecursivelyCreateDir` not behave correctly with fallback?

yongtang:
@vnvo2409 Given the 2.4 is already out and the issue in TF side will not be fixed until next release, I will merge this PR for now. We can always remove the extraneous function once the fix in TF side is fixed.

PR #1219, opened 2020-12-09
Setup GitHub Actions for API Compatibility

yongtang:
This PR setup GitHub Actions for API Compatibility
(file system only due to C vs. C++ API in TensorFlow Core)
on different combinations of tensorflow vs tensorflow-io.

As a first step this PR adds the following API Compatibility tests:
- tensorflow==2.4.0rc4 vs. tensorflow-io-nightly
- tf-nightly vs. tensorflow-io-nightly

In the future addtional combinations might be added.

This will give us a better idea in case tensorflow updates breaks API
or tensorflow-io breaks API.

As we needs to test against TF's nightly build, this PR setup the GitHub
Actions trigger to daily (on 8AM EST).

### DIFF
--- a/.github/workflows/api.yml
+++ b/.github/workflows/api.yml
@@ -1,0 +1,118 @@
+name: API Compatibility
+
+on:
+  schedule:
+    - cron: "0 12 * * *"
+
+env:
+  BAZEL_OPTIMIZATION: --copt=-msse4.2 --copt=-mavx --compilation_mode=opt
+
+jobs:
+
+  macos:
+    name: macOS ${{ matrix.python }} + ${{ matrix.version }}
+    runs-on: macos-latest
+    strategy:
+      matrix:
+        python: ['3.8']
+        version: ['tensorflow==2.4.0rc4:tensorflow-io-nightly', 'tf-nightly:tensorflow-io-nightly']
+    steps:
+      - uses: actions/checkout@v2
+      - uses: docker-practice/actions-setup-docker@v1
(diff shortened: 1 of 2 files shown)


terrytangyuan (approved):

.github/workflows/api.yml
```diff
@@ -0,0 +1,118 @@
...
+  macos:
+    name: macOS ${{ matrix.python }} + ${{ matrix.version }}
+    runs-on: macos-latest
+    strategy:
+      matrix:
+        python: ['3.8']
+        version: ['tensorflow==2.4.0rc4:tensorflow-io-nightly', 'tf-nightly:tensorflow-io-nightly']
+    steps:
+      - uses: actions/checkout@v2
+      - uses: docker-practice/actions-setup-docker@v1
```
kvignesh1420, line 21: @yongtang last time we checked, docker on macOS was a bit unstable. Did they fix the issues in their latest releases? If yes, then we can use it to test additional datasets in macOS env of `build.yml` as well.

yongtang: @kvignesh1420 From the experience it looks like docker on GitHub Actions' macOS is much more stable. We can gradually re-enable the docker on macOS for dataset tests. It is a little slow compared with on Ubuntu though.

yongtang:
@kvignesh1420 The PR has also been rebased to include the commit from #1220 for lint failure.

kvignesh1420 (approved):

PR #1221, opened 2020-12-09
add test for filesystem plugins

vnvo2409:
I can not merge this PR because of #1222 .

I would like to share with you how I think we should test the plugins. The idea is inspired by [modular_filesystem_test](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/c/experimental/filesystem/modular_filesystem_test.cc). I used parameterized tests for control the plugins and environements variable. Please let me know what you think !

Part of #1183

### DIFF
--- a/tests/test_filesystem.py
+++ b/tests/test_filesystem.py
@@ -1,0 +1,357 @@
+# Copyright 2021 The TensorFlow Authors. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may not
+# use this file except in compliance with the License.  You may obtain a copy of
+# the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.  See the
+# License for the specific language governing permissions and limitations under
+# the License.
+# ==============================================================================
+
+import os
+import posixpath
+import time
+from urllib.parse import urlparse
+
+import boto3
+import pytest
+import tensorflow as tf
+import tensorflow_io as tfio  # pylint: disable=unused-import
+
+# pytestmark = pytest.mark.skipif(
+#     sys.platform in ("win32", "darwin"),
+#     reason="TODO emulator not setup properly on macOS/Windows yet",
+# )
+
+ROOT_PREFIX = f"tf-io-root-{int(time.time())}/"
+S3_URI = "s3"
+
+
+def mock_patchs(monkeypatch, patchs):
+    if isinstance(patchs, dict):
+        for key, value in patchs.items():
+            if value is None:
+                monkeypatch.delenv(key, raising=False)
+            else:
+                monkeypatch.setenv(key, value)
+    elif callable(patchs):
+        patchs(monkeypatch)
+    else:
+        pass
+
+
+@pytest.fixture(autouse=True)
+def reload_filesystem():
+    # We need to find a way to reload `tensorflow` or filesystems since
+    # the envs are read in the first time the filesystems are called.
+    pass
+
+
+@pytest.fixture(scope="module")
+def s3_fs():
+    monkeypatch = pytest.MonkeyPatch()
+    bucket_name = os.environ.get("S3_TEST_BUCKET")
+    client = None
+
+    # This means we are running against emulator.
+    if bucket_name is None:
+        endpoint_url = "http://localhost:4566"
+        monkeypatch.setenv("AWS_REGION", "us-east-1")
+        monkeypatch.setenv("AWS_ACCESS_KEY_ID", "TEST")
+        monkeypatch.setenv("AWS_SECRET_ACCESS_KEY", "TEST")
+        monkeypatch.setenv("S3_ENDPOINT", endpoint_url)
+
+        bucket_name = f"tf-io-bucket-s3-{int(time.time())}"
+        client = boto3.client("s3", endpoint_url=endpoint_url)
+        client.create_bucket(Bucket=bucket_name)
+    else:
+        # TODO(vnvo2409): Implement for testing against production scenario
+        pass
+
+    client.put_object(Bucket=bucket_name, Key=ROOT_PREFIX, Body="")
+
+    def parse(path):
+        res = urlparse(path, scheme=S3_URI, allow_fragments=False)
+        return res.netloc, res.path[1:]
+
+    def path_to(*args):
+        return f"{S3_URI}://{bucket_name}/{posixpath.join(ROOT_PREFIX, *args)}"
+
+    def read(path):
+        bucket_name, key_name = parse(path)
+        response = client.get_object(Bucket=bucket_name, Key=key_name)
+        return response["Body"].read()
+
+    def write(path, body):
+        bucket_name, key_name = parse(path)
+        client.put_object(Bucket=bucket_name, Key=key_name, Body=body)
+
+    def mkdirs(path):
+        if path[-1] != "/":
+            path += "/"
+        write(path, b"")
+
+    yield path_to, read, write, mkdirs, posixpath.join
+    monkeypatch.undo()
+
+
+@pytest.fixture
+def fs(request, s3_fs):
+    if request.param == S3_URI:
+        return s3_fs
+
+
+@pytest.mark.parametrize("fs, patchs", [(S3_URI, None)], indirect=["fs"])
+def test_init(fs, patchs, monkeypatch):
+    path_to, _, _, _, _ = fs
+    mock_patchs(monkeypatch, patchs)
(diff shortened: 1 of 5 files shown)


tests/test_fs_plugins.py
```diff
@@ -0,0 +1,106 @@
...
+
+@pytest.mark.parametrize("envs", [None])
+def test_init(uri_builder, envs):
+    uri, _, get_uri = uri_builder
+    setup_env(uri, envs)
+    assert tf.io.gfile.exists(get_uri("")) is True
+
+
+@pytest.mark.parametrize(
+    "envs", [None, (S3_URI, {"S3_DISABLE_MULTI_PART_DOWNLOAD": "1"})]
```
vnvo2409, line 87: `(S3_URI, {"S3_DISABLE_MULTI_PART_DOWNLOAD": "1"})` means this env is specified for `s3e` only. If it is passed into another filesystems ( because of `uri_builder` ) this test will be skipped.

yongtang:
@vnvo2409 parameterized pytest looks good! This can help standardize the file system implementations.

vnvo2409:
@yongtang 
Could you please take a look ? I wrote it with a super flexibility. However, I am not sure if I am overdoing it or not. Thank you !

vnvo2409:
One more thing. I find it hard to test the plugins with environment variables. Because most filesystems like `s3` will only check for the envs when initialization. I am thinking of introduce a new env `S3_FORCE_REINITIALIZATION`, everytime this env is present, everything will be reinitialized by `GetTransferManager`, `GetExecutor` and `GetS3Client` and so on for other filesystem. WDYT ?

vnvo2409:
@yongtang 

Do you have any thought about this approach ? Thank you !

yongtang:
@vnvo2409 I think one concern is that when you change the configuration/initialization, the change applies process wise - meaning all current ongoing pytests (within the same process) will be impacted. Unless we can force pytest to run in a sequence order then there might have unintended effect. One test belonging to one group may get into the way of another test belonging to another group (with different intended settings)?

tests/test_fs_plugins.py
```diff
@@ -0,0 +1,181 @@
...
+        pytest.param(VIEWFS_URI, marks=pytest.mark.skip(reason="TODO")),
+        pytest.param(HAR_URI, marks=pytest.mark.skip(reason="TODO")),
+        pytest.param(GCS_URI, marks=pytest.mark.skip(reason="TODO")),
+    ],
+)
+def uri_init(request):
+    uri = request.param
+    monkeypatch = pytest.MonkeyPatch()
+    yield uri, *MAP_URI_HELPER[uri](monkeypatch)
+    monkeypatch.undo()
```
vnvo2409, line 148: Because the scope of this fixture is `module`, it will be called once for each `uri`. With [monkeypatch](https://docs.pytest.org/en/stable/monkeypatch.html), all the envs for each `uri` will be unset when we call `monkeypatch.undo()`. [`yield`](https://docs.pytest.org/en/stable/fixture.html#fixture-finalization-executing-teardown-code) is a technique allows us to setup and teardown fixture in one function ( and that where we need to call `monkeypatch.undo()` to unset the envs for that `uri` )

So one layer of isolation, env for a `uri` won't be clash with other `uri`.

tests/test_fs_plugins.py
```diff
@@ -0,0 +1,181 @@
...
+
+
+# ----------------------------------- TEST ----------------------------------- #
+
+
+@pytest.mark.parametrize("envs", [("+all", None)])
+def test_init(uri_init, envs, monkeypatch):
+    uri, get_uri, _, _ = uri_init
+    check_test_condition_and_setup_env(uri, envs, monkeypatch)
+    assert tf.io.gfile.exists(get_uri("")) is True
```
vnvo2409, line 158: Another layer of isolation, each test has their own `monkeypatch` ( not the `monkeypatch` inside `uri_init` ). That `monkeypatch` is resposible for unset the env in `envs` ( it will automatically unset these env after each parametrized test ends, note that it is a fixture, not a variable like inside `uri_init` )

Combine with `cond_str` in `check_uri_skip_condition`, we could effectively isolate each test by isolate common envs of each uri and isolate the addition envs that are passed to each tests.

vnvo2409 (commented):
@yongtang 

I am trying to elaborate what I am understand. In summary, there are 2 layers of isolation ( explained below ). So I think these tests won't clash with each other. Please tell me if there are something unclear.

vnvo2409:
@yongtang. Since #1222 is done. We could test the plugins with TF nightly, could you take a look at it ? If it is ok, I will expand for all the operations ( not only reading/writing )

tests/test_fs_plugins.py
yongtang: This sounds good to me 👍

yongtang:
@vnvo2409 Sounds good to me 👍. Let me know when the PR draft is ready

vnvo2409:
@yongtang Sorry for putting this PR on hold for too long. Could you take a look please ? I will add some more tests ( mostly directory-related ) in later PRs. Thank you !

In addition, I've already exposed `env` through https://github.com/tensorflow/tensorflow/pull/46302. Could we update `gcs` right now or we have to wait for `tensorflow 2.5rc0` ?

yongtang:
@vnvo2409 The PR looks good! There are a couple of test failures on Linux, can you take a look (or disable it temporarily for now)?

yongtang:
@vnvo2409 For gcs the plan is to enable modular file system (and use TF_ENABLE_LEGACY_FILESYSTEM=1 to switch on/off) before the 2.5 branch and 2.5.0rc0 is out. Though I think we want to send out the email to community (tf developer google group) first so that people is prepared. /cc @mihaimaruseac

vnvo2409:
@yongtang with `gcs`, I means updating the `gcs_env` https://github.com/tensorflow/io/blob/ac75e1c0b0c89f5d25db48a3c6218ceeee0c1c6a/tensorflow_io/core/plugins/gs/gcs_env.cc#L82-L85

Nothing is related to the release plan I think. But we need the `env` header right now is exposed in `tf-nightly` only so I am not sure if we have to wait for `tf-2.5rc0`

yongtang:
@vnvo2409 tensorflow-io is already using tf-nightly now (see https://github.com/tensorflow/io/blob/master/tensorflow_io/core/python/ops/version_ops.py), which was added as part of PR #1309

vnvo2409:
@yongtang I am so sorry that I will have to put it on hold for much longer. It has some weird errors and I don't have a linux  machine right now. I will probably try it again when tf 2.5 is out and `aws` and `google-cloud-cpp` are removed from TensorFlow.

vnvo2409:
@yongtang It seems that the errors are from the upstream. I would like to ask if it is possible to switch to use `google-cloud-cpp` `bigtable` and `pubsub` client instead of implement our own client with our own proto client ? I could give it a try if you think it's ok.

yongtang:
>  I would like to ask if it is possible to switch to use google-cloud-cpp bigtable and pubsub client instead of implementing our own client with our own proto client ? I could give it a try if you think it's ok.

@vnvo2409 That will be great and preferred 👍

vnvo2409:
@yongtang This PR is ready to merge. Turn out `azurite` was too old and it broke the tests.

The philosophy of the test is still the same. But the filesystem are added manually instead of using parameterized tests which allows greater control and less generated tests.

Still, testing with different envs right now aren't possible as the envs are read only once when the filesystem are loaded for the first time. It will be addressed along with #1342 and #1354.

tests/test_filesystem.py
```diff
@@ -0,0 +1,434 @@
...
+from urllib.parse import urlparse
+
+import boto3
+import pytest
+import tensorflow as tf
+import tensorflow_io as tfio  # pylint: disable=unused-import
+from azure.storage.blob import ContainerClient
+
+pytestmark = pytest.mark.skipif(
+    sys.platform in ("win32", "darwin"),
```
yongtang, line 29: Wondering if we can enable macOS/Darwin?

yongtang: Actually, looks like this part is not needed? Maybe we can remove ln 28-31?

tests/test_filesystem.py
```diff
@@ -0,0 +1,434 @@
...
+
+    assert read(fname) == body
+
+
+@pytest.mark.parametrize(
+    "fs, patchs",
+    [
+        (
+            S3_URI,
+            # `use_multi_part_download` does not work with `seakable`.
```
yongtang, line 206: `seaaable` => `seekable`

tests/test_filesystem.py
vnvo2409: I will send a follow up PR for this and some other improvements.

yongtang: @vnvo2409 Sounds good!

yongtang (approved):
LGTM. A couple of minor cosmetic comments though those can be updated in follow up PRs.

PR #1226, opened 2020-12-14
Bump `aws-sdk-cpp` to 1.18.105

vnvo2409:
This PR:

- Bump `aws-sdk-cpp` to 1.18.105
- Remove unused code for getting s3 region since it will be done automatically by [`aws-sdk-cpp`](https://github.com/aws/aws-sdk-cpp/wiki/What’s-New-in-AWS-SDK-for-CPP-Version-1.8#client-configuration-now-reads-environment-variables-configuration-file-and-ec2-metadata-to-get-default-aws-region)
- Enable [`ENABLE_CURL_LOGGING`](https://github.com/aws/aws-sdk-cpp/wiki/What’s-New-in-AWS-SDK-for-CPP-Version-1.8#turn-on-enable_curl_logging-by-default)
- Support more http code for `TF_SetStatusFromAWSError`
- I am introducing a breaking change here. `RecursivelyCreateDir` now calls `CreateDir` instead of relying on TensorFlow fallback. ( Not really a breaking change though )

Part of #1183

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -385,11 +385,11 @@ http_archive(
     patch_cmds = [
         """sed -i.bak 's/UUID::RandomUUID/Aws::Utils::UUID::RandomUUID/g' aws-cpp-sdk-core/source/client/AWSClient.cpp""",
     ],
-    sha256 = "758174f9788fed6cc1e266bcecb20bf738bd5ef1c3d646131c9ed15c2d6c5720",
-    strip_prefix = "aws-sdk-cpp-1.7.336",
+    sha256 = "e8223d59c4774afa9c1ef40ec68fbb7e9e080725cde024ce0cd8351f98dba00d",
+    strip_prefix = "aws-sdk-cpp-1.8.105",
     urls = [
-        "https://storage.googleapis.com/mirror.tensorflow.org/github.com/aws/aws-sdk-cpp/archive/1.7.336.tar.gz",
-        "https://github.com/aws/aws-sdk-cpp/archive/1.7.336.tar.gz",
+        "https://storage.googleapis.com/mirror.tensorflow.org/github.com/aws/aws-sdk-cpp/archive/1.8.105.tar.gz",
+        "https://github.com/aws/aws-sdk-cpp/archive/1.8.105.tar.gz",
     ],
 )
 
--- a/tensorflow_io/core/plugins/BUILD
+++ b/tensorflow_io/core/plugins/BUILD
@@ -15,9 +15,13 @@ cc_library(
     copts = tf_io_copts(),
     linkstatic = True,
     deps = [
-        "@local_config_tf//:libtensorflow_framework",
         "@local_config_tf//:tf_c_header_lib",
-    ],
+    ] + select({
+        "@bazel_tools//src/conditions:windows": [
+            "@local_config_tf//:libtensorflow_framework",
+        ],
+        "//conditions:default": [],
+    }),
     alwayslink = 1,
 )
 
--- a/third_party/aws-sdk-cpp.BUILD
+++ b/third_party/aws-sdk-cpp.BUILD
@@ -54,10 +54,10 @@ cc_library(
         "aws-cpp-sdk-core/include/aws/core/SDKConfig.h",
     ],
     defines = [
-        'AWS_SDK_VERSION_STRING=\\"1.7.366\\"',
+        'AWS_SDK_VERSION_STRING=\\"1.8.105\\"',
         "AWS_SDK_VERSION_MAJOR=1",
-        "AWS_SDK_VERSION_MINOR=7",
-        "AWS_SDK_VERSION_PATCH=366",
+        "AWS_SDK_VERSION_MINOR=8",
+        "AWS_SDK_VERSION_PATCH=105",
         "ENABLE_OPENSSL_ENCRYPTION=1",
         "ENABLE_CURL_CLIENT=1",
         "OPENSSL_IS_BORINGSSL=1",
--- a/third_party/curl.BUILD
+++ b/third_party/curl.BUILD
@@ -112,7 +112,7 @@ genrule(
         "#  define HAVE_SYS_FILIO_H 1",
         "#  define HAVE_SYS_SOCKIO_H 1",
         "#  define OS \"x86_64-apple-darwin15.5.0\"",
-        "#  define USE_DARWINSSL 1",
+        "#  define USE_SECTRANSP 1",
         "#else",
         "#  define CURL_CA_BUNDLE \"/etc/ssl/certs/ca-certificates.crt\"",
         "#  define GETSERVBYPORT_R_ARGS 6",


tensorflow_io/core/plugins/s3/s3_filesystem.cc
```diff
@@ -125,36 +157,10 @@ static Aws::Client::ClientConfiguration& GetDefaultClientConfig() {
   if (!init) {
     const char* endpoint = getenv("S3_ENDPOINT");
     if (endpoint) cfg.endpointOverride = Aws::String(endpoint);
-    const char* region = getenv("AWS_REGION");
     // TODO (yongtang): `S3_REGION` should be deprecated after 2.0.
```
vnvo2409, line 160: @yongtang 
Maybe we should mark it as deprecated in `2.5rc` and remove it in `2.5` ?

tensorflow_io/core/plugins/s3/s3_filesystem.cc
```diff
@@ -62,29 +62,61 @@ constexpr size_t kUploadRetries = 3;
...
     const Aws::Client::AWSError<Aws::S3::S3Errors>& error, TF_Status* status) {
-  switch (error.GetResponseCode()) {
-    case Aws::Http::HttpResponseCode::FORBIDDEN:
-      TF_SetStatus(status, TF_FAILED_PRECONDITION,
-                   "AWS Credentials have not been set properly. "
-                   "Unable to access the specified S3 location");
-      break;
-    case Aws::Http::HttpResponseCode::REQUESTED_RANGE_NOT_SATISFIABLE:
-      TF_SetStatus(status, TF_OUT_OF_RANGE, "Read less bytes than requested");
-      break;
```
vnvo2409, line 78: I am not sure if this is correct or not since the old code already supported `REQUESTED_RANGE_NOT_SATISFIABLE = 416`. WDYT @yongtang ?
https://github.com/tensorflow/io/blob/9b5e23391dd5fce214ca68fad1ad345e52ce3cac/tensorflow_io/core/plugins/s3/s3_filesystem.cc#L507-L512

tensorflow_io/core/plugins/s3/s3_filesystem.cc
```diff
@@ -62,29 +62,61 @@ constexpr size_t kUploadRetries = 3;
 
 constexpr size_t kS3ReadAppendableFileBufferSize = 1024 * 1024;  // 1 MB
 
-static void* plugin_memory_allocate(size_t size) { return calloc(1, size); }
-static void plugin_memory_free(void* ptr) { free(ptr); }
+static inline char* TF_MakeStatusMessage(const Aws::String& str) {
+  char* msg = new char[str.length() + 1];
+  strcpy(msg, str.c_str());
```
yongtang, line 67: @vnvo2409 Will that cause memory leak as I don't see they are being freed?

vnvo2409: [`Status`](https://github.com/tensorflow/tensorflow/blob/fbcb024d7f107700a35226cd83f7ef25a476036e/tensorflow/core/platform/status.cc#L92-L101) will take the ownership of that `char*` so there won't be any leak.

yongtang: @vnvo2409 Inside `Status` a copy is made:

https://github.com/tensorflow/tensorflow/blob/fbcb024d7f107700a35226cd83f7ef25a476036e/tensorflow/core/platform/status.cc#L97

`string` will take the input of `const char *` or `char *` and make a copy internally. I think it does not take ownership. The input will not be modified and it is up to the caller to free.

The `c_str()` returns a `const char*` which will be deleted when it is out of scope, but once the `const char*` passed to `TF_SetStatus` and returned from `TF_SetStatus`, `TF_SetStatus`  will not touch passed `const char*` anymore (as `TF_SetStatus` make a copy).

Can you point me to the place of the issue you were encountered?

vnvo2409: If so, I am just overthinking about it. Thank you !

tensorflow_io/core/plugins/s3/s3_filesystem.cc
yongtang: @vnvo2409 The switch is controlled in:
https://github.com/tensorflow/io/blob/9b5e23391dd5fce214ca68fad1ad345e52ce3cac/tensorflow_io/core/plugins/s3/s3_filesystem.cc#L519

You can change the default value to `true` and see if the issue is resolved.

tensorflow_io/core/plugins/s3/s3_filesystem.cc
yongtang: @vnvo2409 The comment was copied from tensorflow's source code. I think we can just remove it in the code, and add a note to the documentation. That should be enough (as it was supposed to be deprecated from 2.0 and we are in 2.4/2.5 already now).

vnvo2409: Done

tensorflow_io/core/plugins/s3/s3_filesystem.cc
vnvo2409: I am investigating this issue https://github.com/tensorflow/tensorflow/issues/44564. Let's leave it for another PR.

vnvo2409:
@yongtang 
I could not understand why the test keeps falling with segfault and I could not reproduce it locally. Do you have some hints why it happens ? Thank you very much !

yongtang:
@vnvo2409 The crash could be caused by duplicate symbols between aws and arrow library? Though I am not quite sure. It seems to be happening on Linux only.

Could you try split the PR into two PRs with one PR for aws-sdk-cpp to 1.18.105 and another PR for other changes like `Enable ENABLE_CURL_LOGGING`, `Support more http code for TF_SetStatusFromAWSError`? This can help us locate the cause of the issue better.

In the meantime, I am trying to update the arrow library from 0.16.0 to 2.0.0 in #1231. We need to update arrow library anyway so I think it makes sense to see if updating arrow will resolve the aws+arrow issue.

vnvo2409:
I already rebase to pickup the arrow update but it still failed. Do you have any idea what I should do next @yongtang ? Thank you very much !

yongtang:
@vnvo2409 The Linux build on GitHub Actions is in:
https://github.com/tensorflow/io/blob/master/.github/workflows/build.yml#L222-L227

There are two difference in between the GitHub Actions and the normal `bazel build`.
1. GitHub Actions utilize the optimization mode with `--copt=-msse4.2 --copt=-mavx --compilation_mode=opt` passed to bazel
2. GitHub Actions utilize the docker image from tensorflow's build image which is Ubuntu16.04+ customized gcc from developer toolset.

I tends to believe optimization (1) is the factor though I cannot exclude (2) yet. Will take a look.

tensorflow_io/core/plugins/s3/s3_filesystem.cc
vnvo2409: @yongtang

https://github.com/tensorflow/tensorflow/issues/44564 is resolved. We could enable this option

vnvo2409:
The problem is [`CurlHttpClient.cpp`](https://github.com/aws/aws-sdk-cpp/blob/ec45db72a754a672f6663e9c8849861847fdfe25/aws-cpp-sdk-core/source/http/curl/CurlHttpClient.cpp#L368-L373) throws segfault because `curlVersionData->ssl_version` is `NULL`. Passing `USE_SSL` to `curl` solves the problem. 

cc7c401 also fixes a compilation problem on Mac because `USE_DARWINSSL` is renamed to `USE_SECTRANSP` https://github.com/curl/curl/blob/ccbdbe13c43f15a637afdd1ac4bf5df62c25a235/CMakeLists.txt#L365-L367

vnvo2409:
@yongtang 

I build it [separately](https://github.com/vnvo2409/s3-bazel) and it seems fine without any issue. I doubt that it somehow conflicts with the `s3` inside TensorFlow. On my machine, sometime it works and sometime it throws segfaul 😕. I am building a custom wheel without `s3` to see if the conflict is the problem.

yongtang:
@vnvo2409 Ah that might explain the segment as there are actually two copies of AWS C++ sdk. One is in tensorflow and another is in tensorflow-io. And they are with different versions. If the AWS C++ SDK on both tensorflow-io and tensorflow are updated to the same version then it might be OK (as two copies with same version).. Or if we remove the one from tensorflow (so that there is only one version) then it might also be OK I guess.

kvignesh1420:
@vnvo2409 the PR by @yongtang : (https://github.com/tensorflow/io/pull/1241) to exclude external libraries has been merged. Can you please check and proceed with this PR?

vnvo2409:
Sorry. I'm still stuck at the segfault. I think I will try this PR again when `aws` is completely removed out of Tensorflow.

PR #1229, opened 2020-12-14
Bring back GCS ops.

michaelbanfield:
This was original deleted because 

https://pypi.org/project/tensorflow-gcs-config/

exists, however the source code of this is actually built from tensorflow/io . The reason we have the dedicated package is to have a smaller .so for disk space constrained environments.

We would like to leave these ops in tensorflow/io also.


vnvo2409:
I would like to ask if it has some overlaps with [`gcs`](https://github.com/tensorflow/io/tree/master/tensorflow_io/core/plugins/gs) filesystem. In addition, we now use `google-cloud-cpp` as a dependency, I think you could use that library instead of writing your own function 😄

kvignesh1420 (commented):
@michaelbanfield Currently, the file system plugins are being migrated from `tensorflow` to `tensorflow-io` and the progress is being tracked here https://github.com/tensorflow/io/issues/1183. The goal is to have the file system extensions use a common modular approach and have a similar python API usage for all the schemes.

yongtang:
My understanding is that the GCS ops is to be able to make adjustment (e.g., `max_cache_size`, `block_size`, `service account`, `refresh token`, etc), in real time within the same graph session. This is useful in situation where tensorflow has already been initialized when user get access.

vnvo2409:
I didn't expect there are many dependencies related to `gcs`. We will have to think about the design of `gcs` right now. There are some problems:
- As I am using `google-cloud-cpp`, we lost control to some parameters ( auth, token, ... ).
- The inheritance ( like `RetryingGcsFileSystem*` ) will not be possible. I am thinking about introducing a new env to control it. https://github.com/tensorflow/io/pull/1221#issuecomment-744355685 )

Apologize for all the back and forth.

yongtang:
@vnvo2409 I think the issue with env variable is that you have to pass the env at the initialization time. This can post some limitations where tensorflow might have already been imported/initialized.

An alternative approach is to hold a global variable in C/C++ for all the gcs related configurations. This will allow having an extra ops to modify the gcs configurations in runtime within the same graph session. This is more or less the same way gcs ops works in this PR.

michaelbanfield:
Thanks for the comments, moving gcs filesystem to use `google-cloud-cpp` makes a lot of sense.

As discussed above these ops allow clients to modify gcs auth while tensorflow is already initialized. There are a few use cases that depend on this so we would want to retain this functionality when splitting to the modular filesystem.

Is there any concerns merging in these ops for the current filesystem? They used to live within tensorflow but were moved out to IO. Once the modular GCS filesystem supports configuration after initialization we can move to that.

yongtang:
@michaelbanfield That should be fine. Though the test is failing now. Can you apply the following patch:
```diff
diff --git a/tensorflow_io/core/BUILD b/tensorflow_io/core/BUILD
index 3563b04..e661149 100644
--- a/tensorflow_io/core/BUILD
+++ b/tensorflow_io/core/BUILD
@@ -724,6 +724,7 @@ cc_binary(
         "//tensorflow_io/core:text_ops",
         "//tensorflow_io/core:ignite_ops",
         "//tensorflow_io/core:mongodb_ops",
+        "//tensorflow_io/gcs:gcs_config_ops",
         "@local_config_tf//:libtensorflow_framework",
         "@local_config_tf//:tf_header_lib",
     ] + select({
diff --git a/tests/test_gcs_config_ops.py b/tests/test_gcs_config_ops.py
index 7d01409..291986c 100644
--- a/tests/test_gcs_config_ops.py
+++ b/tests/test_gcs_config_ops.py
@@ -29,7 +29,7 @@ tf_v1 = tf.version.VERSION.startswith("1")
 class GcsConfigOpsTest(test.TestCase):
     """GCS Config OPS test"""
 
-    @pytest.mark.skipif(sys.platform == "darwin", reason=None)
+    @pytest.mark.skipif(sys.platform == "win32", reason="Windows not working yet")
     def test_set_block_cache(self):
         """test_set_block_cache"""
         cfg = gcs.BlockCacheParams(max_bytes=1024 * 1024 * 1024)
```

Once the patch is applied I think all tests will pass.

tensorflow_io/gcs/kernels/gcs_config_op_kernels.cc
```diff
@@ -0,0 +1,206 @@
...
+  if (filesystem == nullptr) {
+    return errors::FailedPrecondition("The GCS file system is not registered.");
+  }
+
+  *fs = dynamic_cast<RetryingGcsFileSystem*>(filesystem);
+  if (*fs == nullptr) {
+    return errors::Internal(
+        "The filesystem registered under the 'gs://' scheme was not a "
+        "tensorflow::RetryingGcsFileSystem*.");
+  }
```
kvignesh1420, line 61: @michaelbanfield I think the `dynamic_cast<RetryingGcsFileSystem*>(filesystem)` is returning a null pointer which in turn is raising the error. I think this is because the gs file system plugin has already been registered. Can we remove this typecast operation?

PR #1231, opened 2020-12-16
Bump Apache Arrow to 2.0.0

yongtang:
This PR bumps Apache Arrow to 2.0.0, also bumps Apache Thrift to 0.13.0, and flatbuffers to 1.12.0.


tests/test_feather_eager.py
```diff
@@ -37,9 +38,7 @@ def test_feather_format():
     }
     df = pd.DataFrame(data).sort_index(axis=1)
     with tempfile.NamedTemporaryFile(delete=False) as f:
-        df.to_feather(f)
-
-    df = pd.read_feather(f.name)
+        pa.feather.write_feather(df, f, version=1)
```
BryanCutler, line 41: version 2 is the includes compression by default, is only version 1 working when reading reading back in?

BryanCutler (approved):
This look great, thanks for doing it @yongtang !  I've been wanting to upgrade Arrow for a while, but been too busy with other projects right now. The changes LGTM.

tests/test_feather_eager.py
yongtang: Thanks @BryanCutler for the review. It only works for version 1 now as version 2 changed the magic string from `FEA1` to `ARROW1` (and a few other changes in place). I am thinking of adding version 2 support in follow up PRs. This PR is just to bring arrow library to the latest version.

BryanCutler: Ok great, SGTM

chengji77:
Hi @yongtang , do you also need to bump the thrift version here? https://github.com/tensorflow/io/blob/318aab95d48598e62cf1d197307c3e12e72bfb1d/third_party/thrift.BUILD#L43

PR #1234, opened 2020-12-17
Add emulator for gcs

vnvo2409:
In this PR:

- Bump com_github_googleapis_google_cloud_cpp to `1.21.0`. Because there is a change in the xml endpoint of `google-cloud-cpp` https://github.com/googleapis/google-cloud-cpp/pull/5097

- Copy testbench from `google-cloud-cpp 1.19.0`. Note that this is the old emulator. I prefer it over the new emulator because it is much simpler. The new emulator is more complicated and supports `gRPC` API, which we don't need. In addition, I delete many functions releated to `acl` and keep only data operation ( upload, download, compose, ... ).

- Bump `libcurl` to `7.69.1` to fix `EasyPause() - CURL error [43]=A libcurl function was given a bad argument`

### DIFF
--- a/.github/workflows/api.yml
+++ b/.github/workflows/api.yml
@@ -50,6 +50,7 @@ jobs:
           python -m pytest -s -v tests/test_http_eager.py
           python -m pytest -s -v tests/test_s3_eager.py
           python -m pytest -s -v tests/test_azure.py
+          python -m pytest -s -v tests/test_gcs_eager.py
 
   linux:
     name: Linux ${{ matrix.python }} + ${{ matrix.version }}
@@ -85,6 +86,7 @@ jobs:
           python -m pytest -s -v tests/test_http_eager.py
           python -m pytest -s -v tests/test_s3_eager.py
           python -m pytest -s -v tests/test_azure.py
+          python -m pytest -s -v tests/test_gcs_eager.py
 
   windows:
     name: Windows ${{ matrix.python }} + ${{ matrix.version }}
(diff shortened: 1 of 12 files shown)


vnvo2409:
@yongtang 

It seems that I did not setup the emulator correctly. Could you take a look please ?

vnvo2409:
@yongtang 
`GitHub CI / Test 3.8 Linux` is passing. I think we could merge this PR. Could you take a look ? Thank you !

yongtang (approved):
PR looks good to me. The macOS is a known issue as this week GitHub Actions is experiencing degradation of macOS.

yongtang:
@vnvo2409 I noticed you already enabled the read of gcs test and it passes. Just wondering if `StartThread` is even invoked since we haven't added the code to StartThread yet?
https://github.com/tensorflow/io/blob/master/tensorflow_io/core/plugins/gs/gcs_env.cc#L149-L155

vnvo2409:
No indeed. It is invoked only when `max_staleness_ > 0`
https://github.com/tensorflow/io/blob/4a73f6051b3d4a7da66a200454994833407cb99c/tensorflow_io/core/plugins/gs/ram_file_block_cache.h#L67-L75

But the default value for staleness is `0`
https://github.com/tensorflow/io/blob/4a73f6051b3d4a7da66a200454994833407cb99c/tensorflow_io/core/plugins/gs/gcs_filesystem.cc#L50

yongtang:
@vnvo2409 Ah thanks for the explanation.

PR #1238, opened 2020-12-18
fix nightly build because of missing `google-cloud-storage`

vnvo2409:
@yongtang 

Sorry for the back and forth. I did not notice it since it was skipped in the last PR.


yongtang (approved):
LGTM

vnvo2409:
@yongtang 
The `API Compatibility` still fails on e2bb58f0882a214343d49264be1bd33ead041f37. AFAIK, nightly build will use the wheel package from yesterday so it could result in a regression between master and nightly. I am watiting for other builds to see if they pass or not.

yongtang:
Thanks @vnvo2409 for the update

vnvo2409:
@yongtang 

FYI, both `Linux 3.7` and `Linux 3.8` are passing in 4a73f6051b3d4a7da66a200454994833407cb99c. So there are nothing to worry about.

Issue #1240, opened 2020-12-21
import tf-io failed when using custom tf wheel


vnvo2409:
I've built a custom tf wheel (OS: Ubuntu 20.10, commit https://github.com/tensorflow/tensorflow/commit/b19ef9bf1f7caca22576f3cdd01b9d2aa0e13653)  with `--config=noaws --config=nogcp --config=nohdfs --config=nonccl` ( basically no `boringssl`, `curl` stuff and no filesystem ). When I tried to import `tf-io` ( both nightly and `0.17` ), I got this error
```
Traceback (most recent call last):
  File "/home/vovannghia2409/programming/.venvs/tf-mini/lib/python3.8/site-packages/tensorflow_io/core/python/ops/__init__.py", line 95, in <module>
    plugin_ops = _load_library("libtensorflow_io_plugins.so", "fs")
  File "/home/vovannghia2409/programming/.venvs/tf-mini/lib/python3.8/site-packages/tensorflow_io/core/python/ops/__init__.py", line 68, in _load_library
    raise NotImplementedError(
NotImplementedError: unable to open file: libtensorflow_io_plugins.so, from paths: ['/home/vovannghia2409/programming/.venvs/tf-mini/lib/python3.8/site-packages/tensorflow_io/core/python/ops/libtensorflow_io_plugins.so']
caused by: ['/home/vovannghia2409/programming/.venvs/tf-mini/lib/python3.8/site-packages/tensorflow_io/core/python/ops/libtensorflow_io_plugins.so: undefined symbol: curl_easy_cleanup']

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/vovannghia2409/programming/.venvs/tf-mini/lib/python3.8/site-packages/tensorflow_io/__init__.py", line 17, in <module>
    from tensorflow_io.core.python.api.v0 import *  # pylint: disable=wildcard-import
  File "/home/vovannghia2409/programming/.venvs/tf-mini/lib/python3.8/site-packages/tensorflow_io/core/python/api/v0/__init__.py", line 18, in <module>
    from tensorflow_io.core.python.ops.io_dataset import IODataset
  File "/home/vovannghia2409/programming/.venvs/tf-mini/lib/python3.8/site-packages/tensorflow_io/core/python/ops/__init__.py", line 98, in <module>
    core_ops = _load_library("libtensorflow_io.so")
[...]

PR #1241, opened 2020-12-21
Remove AWS headers from tensorflow, and use headers from third_party …

yongtang:
…in tensorflow-io instead

This PR removes AWS headers from tensorflow and use headers
from third_party packages in tensorflow-io instead

### DIFF
--- a/third_party/toolchains/tf/tf_configure.bzl
+++ b/third_party/toolchains/tf/tf_configure.bzl
@@ -147,7 +147,7 @@ def _symlink_genrule_for_dir(
     if src_dir != None:
         src_dir = _norm_path(src_dir)
         dest_dir = _norm_path(dest_dir)
-        files = "\n".join(sorted(_read_dir(repository_ctx, src_dir).splitlines()))
+        files = "\n".join(sorted([e for e in _read_dir(repository_ctx, src_dir).splitlines() if ("/external/" not in e) and ("/absl" not in e)]))
 
         # Create a list with the src_dir stripped to use for outputs.
         if tf_pip_dir_rename_pair_len:
--- a/third_party/toolchains/tf/tf_configure.bzl
+++ b/third_party/toolchains/tf/tf_configure.bzl
@@ -147,7 +147,7 @@ def _symlink_genrule_for_dir(
     if src_dir != None:
         src_dir = _norm_path(src_dir)
         dest_dir = _norm_path(dest_dir)
-        files = "\n".join(sorted([e for e in _read_dir(repository_ctx, src_dir).splitlines() if ("/external/" not in e) and ("/absl" not in e)]))
+        files = "\n".join(sorted([e for e in _read_dir(repository_ctx, src_dir).splitlines() if ("/external/" not in e) and ("/absl/" not in e)]))
 
         # Create a list with the src_dir stripped to use for outputs.
         if tf_pip_dir_rename_pair_len:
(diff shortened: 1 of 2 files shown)


kvignesh1420 (approved):
LGTM. @yongtang lint check is failing though!

terrytangyuan (approved):

vnvo2409:
@yongtang Do you think that we should remove `curl`, `boringssl` as well because in #1240, we are missing `curl_easy_cleanup`

yongtang:
@kvignesh1420 @terrytangyuan  @vnvo2409 The PR has been updated with lint fixed and curl, boringssl headers removed as well.

vnvo2409:
@yongtang. LGTM. Thank you!

vnvo2409:
@yongtang I think there is one more leak with `absl::StrCat`. I haven't tried this patch yet, but when I loaded `libtensorflow_io_plugins.so` , it said that there was a mismatch with the symbol of `absl::StrCat`. I think we have to exclude some more headers :confused:

vnvo2409:
@yongtang 

Here is my list of excluded headers:
```
("/external/aws" not in e) and
("external/boringssl" not in e) and
("external/curl" not in e) and
("absl" not in e)
```

One thing I notice is `hadoop` do use `absl` but does not include `absl` in `deps`https://github.com/tensorflow/io/blob/e2bb58f0882a214343d49264be1bd33ead041f37/tensorflow_io/core/plugins/hdfs/BUILD#L17-L20

I think `hadoop` use `absl` of `tensorflow` which cause an ABI incompatibility. Let me take a look at `tf_configure.bzl`, I will try to extract only `tensorflow/c/` headers and put it into a separate rule ( `@local_config_tf//:tf_c_header_lib` ). 

Afterward, we could replace this https://github.com/tensorflow/io/blob/e2bb58f0882a214343d49264be1bd33ead041f37/tensorflow_io/core/plugins/BUILD#L17-L20 with `@local_config_tf//:tf_c_header_lib`.

kvignesh1420:
@yongtang how can we proceed with this PR?

yongtang:
@kvignesh1420 The PR has been updated. Now all external headers (not just aws/etc) have been excluded except for protobuf/grpc, which seems to be too embedded with tensorflow's headers.

Please take a look at the changes.

third_party/toolchains/tf/tf_configure.bzl
```diff
@@ -147,7 +147,7 @@ def _symlink_genrule_for_dir(
     if src_dir != None:
         src_dir = _norm_path(src_dir)
         dest_dir = _norm_path(dest_dir)
-        files = "\n".join(sorted(_read_dir(repository_ctx, src_dir).splitlines()))
+        files = "\n".join(sorted([e for e in _read_dir(repository_ctx, src_dir).splitlines() if ("/external/" not in e) and ("/absl" not in e)]))
```
kvignesh1420, line 150: can we change it from `/absl` to `/absl/` just to be safe and consistent with the other condition?

kvignesh1420 (changes requested):
@yongtang sorry for the delay. I was travelling and couldn't review it earlier. Can you please check the minor change request?

yongtang:
@kvignesh1420 The PR has been updated. Please take a look.

kvignesh1420 (approved):
thanks @yongtang

PR #1242, opened 2020-12-22
Remove redundant output of dataset.element_spec in PostgreSQL tutorial

terrytangyuan:
The output is already shown as part of the previous code block `print(dataset.element_spec)`.


kvignesh1420 (approved):
LGTM.

PR #1244, opened 2020-12-22
add tf-c-header rule

vnvo2409:
Related to https://github.com/tensorflow/io/pull/1241#issuecomment-749376381

@yongtang This is my approach for dividing the header from TensorFlow into smaller rules. Using the same approach, we could effectively have `tf_core_header_lib` and `tf_cc_header_lib` and remove all the `third_party` and `external` headers from TensorFlow which cause a lot of problems.

There is a lot of `#include "tensorflow/core/..."` in the codebase right now. I believe introducing `tf_core_header_lib` and removing `tf_header_lib` could help us to prevent a lot of problem related to symbol leak.

### DIFF
--- a/third_party/toolchains/tf/BUILD.tpl
+++ b/third_party/toolchains/tf/BUILD.tpl
@@ -8,6 +8,14 @@ cc_library(
 )
 
 cc_library(
+    name = "tf_c_header_lib",
+    hdrs = [":tf_c_header_include"],
+    include_prefix = "tensorflow/c",
+    strip_include_prefix = "include_c",
+    visibility = ["//visibility:public"],
+)
+
+cc_library(
     name = "libtensorflow_framework",
     srcs = [":libtensorflow_framework.so"],
     #data = ["lib/libtensorflow_framework.so"],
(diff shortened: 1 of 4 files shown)


third_party/toolchains/tf/BUILD.tpl
```diff
@@ -7,6 +7,14 @@ cc_library(
     visibility = ["//visibility:public"],
 )
 
+cc_library(
+    name = "tf_c_header_lib",
+    hdrs = [":tf_c_header_include"],
+    include_prefix = "tensorflow/c",
+    strip_include_prefix = "include_c",
+    visibility = ["//visibility:public"],
+)
```
vnvo2409, line 16: Should we add 
```
deps = [
        "@local_config_tf//:libtensorflow_framework",
    ],
```

here since they always go with each other ?

yongtang (approved):
LGTM

vnvo2409:
@yongtang 

I would like to send another PR for replacing `tf_header_lib` with `tf_core_header_lib`. What do you think ?

yongtang:
@vnvo2409 makes sense to me 👍

PR #1246, opened 2020-12-22
Add `fail-fast: false` to API Compatibility GitHub Actions

yongtang:
This PR adds `fail-fast: false` to API Compatibility GitHub Actions.
The main reason is to make sure if any job fails, the parallel jobs
within the same matrix of the workflow can continue. The API Compatibility
is to see how our plugin binaries match with different versions
and as such we want to see the whole compatibility matrix results.

### DIFF
--- a/.github/workflows/api.yml
+++ b/.github/workflows/api.yml
@@ -16,6 +16,7 @@ jobs:
     name: macOS ${{ matrix.python }} + ${{ matrix.version }}
     runs-on: macos-latest
     strategy:
+      fail-fast: false
       matrix:
         python: ['3.8']
         version: ['tensorflow==2.4.0:tensorflow-io-nightly', 'tf-nightly:tensorflow-io==0.17.0', 'tf-nightly:tensorflow-io-nightly']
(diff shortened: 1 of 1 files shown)


terrytangyuan (commented):
The downside is that contributors would no longer be able to see all remaining errors and additional build time might be wasted since they need to push a commit to fix this first before running the rest of the tests.

yongtang:
@terrytangyuan Yes that is why we set `fail-fast` to `false`.

terrytangyuan (approved):
Oh I see. I thought we were doing the opposite in this PR. LGTM then.

vnvo2409:
@yongtang 

`Linux 3.8 + tf-nightly:tensorflow-io==0.17.0` failed because gcs filesystem does not work correctly with tensorflow_io==0.17.0 ( because `google-cloud-cpp` does not work with `libcurl <= 7.68.0` ). If we want to have a green HEAD, I think we have to remove `test_gcs_eager.py` in the API compatibility for now.

PR #1248, opened 2020-12-23
S3 Improvements

vnvo2409:
It seems that there are still some problems with symbol leaks and #1226.

So this PR:
- adds some improvements for `s3` without bumping to newer version.
- fixes a ssl support problem on Mac because USE_DARWINSSL is renamed to USE_SECTRANSP https://github.com/curl/curl/blob/ccbdbe13c43f15a637afdd1ac4bf5df62c25a235/CMakeLists.txt#L365-L367.


yongtang (approved):
LGTM, Thanks!

Issue #1245, opened 2020-12-22
decode_vorbis causing RAM leak


marno1d:
I'm seeing an issue where using tfio.audio.decode_vorbis is causing a RAM memory leak. If I call decode_vorbis in a loop the RAM increases. However, when I call decode_wav in a loop the RAM stays constant. 

TF Version: 2.4
TF IO Version 0.17.0
OS: Ubuntu 18.04

I created a colab notebook that reproduces this issue: https://colab.research.google.com/drive/1W0ubBFr2lEEHeqAYDguXgAyFZgjuZOvl?usp=sharing

There was a similar issue where decode video was causing a memory leak #680. Perhaps a similar problem is happening here.

kvignesh1420:
@marno1d PR: https://github.com/tensorflow/io/pull/1249 has been opened to address this issue.

marno1d:
@kvignesh1420 Great, thanks for getting to this so quickly. I really appreciate the whole tfio.audio library.

kvignesh1420:
@marno1d I didn't observe any obvious mem leaks while trying the `decode_vorbis` functionality on my mac. However, I have addresses the file cleanup in the PR. Folks might be on vacation so it might take some time to get this reviewed.

marno1d:
@kvignesh1420 I had a chance to test these changes in the new tensorflow-io-nightly build and it seemed to get rid of the memory leak. Thanks for making this fix. I'll close this issue now.

PR #1249, opened 2020-12-24
[audio] cleanup vorbis file after usage

kvignesh1420:
This PR aims to address the issue: https://github.com/tensorflow/io/issues/1245 by cleaning up the `ogg_vorbis_file_` pointer after decoding the content. Also, it initializes the output buffer only once, instead of reinitializing it in a loop.

Reference: https://xiph.org/vorbis/doc/vorbisfile/ov_clear.html

@yongtang not sure if this is the root cause of the mem leak. I didn't find any other obvious places where the leak might be possible. However, since the cleanup was missing, I hope this helps in a minor way at least.


kvignesh1420:
The internal CI builds are failing due to Bazel version mismatch.

terrytangyuan (approved):

yongtang (approved):
LGTM

PR #1253, opened 2020-12-30
[s3] add support for testing on macOS

kvignesh1420:
This PR addresses #1183 by modifying the`localstack` docker setup using a docker-compose file and modifying the `$TMPDIR` mount. 

I was able to run the `tests/test_s3_eager.py` tests on my mac after this change.

Reference: https://hub.docker.com/r/localstack/localstack (see `using docker-compose` section)


yongtang (approved):
LGTM, Thanks!

PR #1262, opened 2021-01-06
Fix incomplete row reading issue in parquet files

### LINKED ISSUE #1254: "InvalidArgumentError: null value in column" when loading Parquet into tfio.IODataset when column has no actual null values
Loading movielens data from Parquet I pre-generated, into an IODataset, yields error as below (stack included).

I'm attaching sample parquet files. I do not see any records where the `movie_title` column value would be null.

Code is also attached. The logic flow is:
1. Load the CSV movielens data from AWS S3.
2. Convert this data to Parquet and store Parquet in S3. This generates the Parquet we can test with.
3. Download the Parquet files from S3 to the local temp dir. Have to do this as a workaround for #1252 (can't stream Parquet directly into TF IO to load a dataset).
4. Load a TF IO dataset from the Parquet data.
5. Observe the error tensorflow.python.framework.errors_impl.InvalidArgumentError: null value in column: movie_title.
6. Note: I tried Parquet with the compression option (snappy) and without, same error.

Check either of the attached .parquet files, esp. the ratings file, e.g. with `parquet-tools cat --json part-00000-b1706a65-3a09-4e0b-a7fb-33e5144d046c-c000.snappy.parquet`.  No null/empty movie_title values.

The error appears to be coming from ./tensorflow_io/core/kernels/parquet_kernels.cc.

Trace:

> 2020-12-30 20:33:28.965622: W tensorflow/core/framework/op_kernel.cc:1763] OP_REQUIRES failed at io_kernel.h:78 : Invalid argument: null value in column: movie_title
  File "/home/hadoop/.local/lib/python3.7/site-packages/tensorflow/python/eager/context.py", line 2113, in execution_mode
    yield
  File "/home/hadoop/.local/lib/python3.7/site-packages/tensorflow/python/data/ops/iterator_ops.py", line 733, in _next_internal
  File "/home/hadoop/.local/lib/python3.7/site-packages/tensorflow/python/ops/gen_dataset_ops.py", line 2579, in iterator_get_next
  File "/home/hadoop/.local/lib/python3.7/site-packages/tensorflow/python/framework/ops.py", line 6862, in raise_from_not_ok_status
tensorflow.python.framework.errors_impl.InvalidArgumentError: null value in column: movie_title
	 [[{{node IO>ParquetReadableRead}}]] [Op:IteratorGetNext]
>
> During handling of the above exception, another exception occurred:
>
> Traceback (most recent call last):
  File "/mnt/tmp/spark-e0c9234e-cfad-460a-8c83-1d4f2b86aef8/my_app.py", line 367, in <module>
    main(sys.argv)
  File "/mnt/tmp/spark-e0c9234e-cfad-460a-8c83-1d4f2b86aef8/my_app.py", line 81, in main
    movies_ds, test, train, unique_movie_titles, unique_user_ids = prepare_data(movies_ds, ratings_ds)
  File "/mnt/tmp/spark-e0c9234e-cfad-460a-8c83-1d4f2b86aef8/my_app.py", line 197, in prepare_data
    print(">> SIZE(ratings_no_titles_ds) = " + str(ds_size(ratings_no_titles_ds)))
  File "/mnt/tmp/spark-e0c9234e-cfad-460a-8c83-1d4f2b86aef8/my_app.py", line 176, in ds_size
    for elem in ds:
  File "/home/hadoop/.local/lib/python3.7/site-packages/tensorflow/python/data/ops/iterator_ops.py", line 747, in __next__
    return self._next_internal()
  File "/home/hadoop/.local/lib/python3.7/site-packages/tensorflow/python/data/ops/iterator_ops.py", line 739, in _next_internal
  File "/usr/lib64/python3.7/contextlib.py", line 130, in __exit__
  File "/home/hadoop/.local/lib/python3.7/site-packages/tensorflow/python/eager/context.py", line 2116, in execution_mode
    executor_new.wait()
  File "/home/hadoop/.local/lib/python3.7/site-packages/tensorflow/python/eager/executor.py", line 69, in wait
    pywrap_tfe.TFE_ExecutorWaitForAllPendingNodes(self._handle)
tensorflow.python.framework.errors_impl.InvalidArgumentError: null value in column: movie_title
	 [[{{node IO>ParquetReadableRead}}]]

[tf_io_issue_parquet_to_ds_null_column.zip](https://github.com/tensorflow/io/files/5755806/tf_io_issue_parquet_to_ds_null_column.zip)

yongtang:
This PR tries to address the issue raised in #1254 where reading parquet
files will results in `InvalidArgumentError: null value in column`

The issue comes from the fact that parquet's ColumnReader C++ API
`ReadBatch(...)` does not necessarily respect the number of rows
requested and may return less instead.

This PR fixes #1254.


tensorflow_io/core/kernels/parquet_kernels.cc
```diff
@@ -162,6 +162,9 @@ class ParquetReadableResource : public ResourceBase {
           row_group_reader->Column(column_index);
 
       // buffer to fill location is value.data()[row_to_read_start - start]
+      // Note: ReadBatch may not be able to read the elements requested
+      // (row_to_read_count) in one shot, as such we use while loop of
+      // `while (row_left > 0) {...}` to read until complete.
```
kvignesh1420, line 167: @yongtang the puzzling thing was that this happened only for a particular column in the parquet dataset that was provided. Any reference as to why this might happen?

dgoldenberg-audiomack: Yes, it'd be great to understand how general this error is. Any workaround for now? When might this fix get into a release?

kvignesh1420: @dgoldenberg-audiomack this PR should fix the issue. Also, can we use your sample parquet dataset in our CI tests (if that is fine with you)?

dgoldenberg-audiomack: @kvignesh1420 Use the parquet dataset - yes, you can absolutely. So the fix will be in the next release; do you know when that's slated for?

kvignesh1420: @dgoldenberg-audiomack I am not sure about the next release, but you can always use `tensorflow-io-nightly` for using these immediate fixes.

kvignesh1420: Will merge this after the tests pass.

yongtang: Thanks @kvignesh1420 @dgoldenberg-audiomack. The `ReadBatch` was only encountered for ByteArray. This is related to the internal implementation of Arrow's Parquet cpp. One discrepancy I can think of, is that ByteArray are different from other types where other types are more uniform with size known before hand (e.g., the size can be preallocated for float/bool/int as long as the number of rows are known). For ByteArray (variable size) the allocation might be hard to pre-allocate.

kvignesh1420 (approved):
LGTM. Thanks, @yongtang for the info and fix.

dgoldenberg-audiomack:
Could folks pls provide any info on what version of TF this fix is/will be available in? Thanks.

PR #1280, opened 2021-01-21
Super Serial- automatically save and load TFRecords from Tensorflow datasets

markemus:
TFRecords are notoriously difficult to work with, but no longer! Super Serial provides a simple and easy to use way to save (dictionary-style) datasets as TFRecords and load them back into memory. The saver will create a .header file that stores the metadata needed to reconstruct the dataset in a human-readable YAML format.

Users only need to deal with 2 functions:
`save(dataset)`
`dataset = load(tfrecord_path, header_path)`

Yes, it's really that easy! If you've worked with TFRecords before you'll know this is a big improvement.

I added super_serial.py to the "api" subfolder- not sure if that's the best place for it or not. Please let me know if there's anything more I need to do to make this acceptable. It also should not be too difficult to extend this to tuple-style datasets if you guys want me to.


markemus:
@googlebot I signed it!

markemus:
@googlebot I fixed it.

markemus:
@googlebot I fixed it.

markemus:
I have no idea why Lint is failing- Pycharm's linter indicates that my code is perfectly conformant to pep8. I tried with 2 space indents instead of 4 (as in the base Tensorflow contributor guidelines) but that didn't help, so I switched it back to 4. Can anyone advise me on what the issue is?

yongtang:
@markemus You can run `bazel run //tools/lint:lint` and that will fix most of the lint issues automatically.

markemus:
@yongtang I ran lint myself from the command line and apparently it's failing on the fstrings (line 60 of super_serial and when that's "fixed" it fails on line 60, also an fstring). Does that make sense? Should I go with old school format strings instead?

I will try bazel after work tonight.

yongtang:
@markemus Which python version are you using? Wondering if the issue is tied to the python version (we support 3.6 or higher).

markemus:
@yongtang I ran it again in python 3.8.3 (there was a conda env active last time), and this time I got a bunch of warnings. The fstring issue seems to have disappeared so that must have had something to do with the python version in conda.

The warnings are:

line-too-long
fix-me
import-error (for tensorflow- that's a weird one, prob safe to ignore)
invalid-name
too-many-function-args

code rating: 6.74/10

Okay, point taken. Is it necessary to fix all of these (ie I like to use "x" when looping and "y" if looping again) or are only some necessary? I'd like to keep the TODOs if possible- they're not bugs but they are features I'd like to add in future, and Pycharm tracks TODO comments.

yongtang:
@markemus Are those errors `pylint` related? The `tensorflow/io` repo does not require pylint check. We only do `black` and `pyupgrade` check instead.

What is the output when you run:
```
bazel run //tools/lint:lint -- black pyupgrade
```

locally?

markemus:
@yongtang are you able to see the checks on the pull request? It lists `GitHub CI / Lint (pull_request)` as a failing check. Bazel is listed as successful for macOS, Linux, and Windows, but I don't have bazel installed on my local system to test it out.

yongtang:
@markemus you can run `black` and `pyupgrade` locally to re-format the `.py` file. 

We use:
```
black==19.10b0
pyupgrade==2.1.0
```

in our CI lint setup.

markemus:
@yongtang ok I appreciate it- I will set up bazel tonight and hopefully that will get things working. Thanks!

yongtang:
@markemus You can install bazel which will not require black or pyupgrade installation explicitly, or install black or pyupgrade through pip or conda and run black and pyupgrade locally without bazel.

markemus:
@yongtang thank you, I ran bazel with `bazel run //tools/lint:lint -- black pyupgrade` and that fixed the lint issues. All the other checks broke though! From the log it looks as though that's a problem with the google servers, is that correct or do I need to fix something else?

yongtang:
@markemus I think that was an intermittent error caused by the connection from GitHub Actions to nasm. The issue should have been gone by now. Can you try bump the commit and push again? An empty commit will re-trigger the GitHub Actions to re-rerun.

yongtang:
@markemus I will take a detailed review over the weekend. I think some mechanic changes to conform to tensorflow-io's layout will be needed.

yongtang:
@markemus Can you move the test to tests directory, e.g., with name `tests/test_xxx.py`? That will trigger the GitHub CI to perform the test automatically. You can use `tests/test_s3_eager.py` as an example.

burgerkingeater:
@markemus thanks for the contribution. I wonder why a file of metadata to reconstruct the dataset is needed? TFRecord doesn't need schema to read, right?

markemus:
@burgerkingeater as far as I know it is necessary to define a loading function as well- see the tutorial [here](https://www.tensorflow.org/tutorials/load_data/tfrecord) or this separate Keras tutorial [here](https://keras.io/examples/keras_recipes/tfrecord/). If there is already an automated way to do this I would love to know about it!

BTW are the issues with the Bazel checks failing resolved? 

@yongtang I will hopefully fix the test tomorrow- thank you for taking a look at this and for all your help this week.

markemus:
@yongtang I moved the test as you requested- can you confirm that it's what you had in mind? All checks are passing now.

yongtang:
Thanks @markemus. I think this feature is fine to be put into experimental package first. There are some mechanics that may need to be done:
1. Add License headers to `super_serial.py` and `test_super_serial.py`
2. Move `super_serial.py` to `tensorflow_io/core/python/experimental/serial_ops.py` (let's remove the `super_`)
3. Import the exposed functions (`save` and `load`) to `tensorflow_io/core/python/api/experimental/serialization.py`

The above will correctly expose the API in a way so that docs in tensorflow.org can handle.

After that, in `test_super_serial.py` the import should be changed from
```
import tensorflow_io.core.python.api.v0.super_serial as ser
```
to
```
import tensorflow_io as tfio
```

and use `tfio.experimental.serialization.save`/`tfio.experimental.serialization.load` directly.

Additionally, there might be some minor changes that is worth considering, e.g., maybe it makes sense to rename the API as `save_dataset` and `load_dataset`?

tensorflow_io/core/python/experimental/serial_ops.py
```diff
@@ -0,0 +1,192 @@
+# Copyright 2020 The TensorFlow Authors. All Rights Reserved.
```
kvignesh1420, line 1: The year should be 2021. Applicable to other newly created files as well.

markemus: Done. I didn't change it in the older file I modified (tfio.core.python.api.experimental.serialization.py) but I changed it in the two I created.

tensorflow_io/core/python/experimental/serial_ops.py
```diff
@@ -0,0 +1,192 @@
...
+    WARNING: Only compatible with "dictionary-style" datasets {key: val, key2:val2,..., keyN, valN}.
+    WARNING: Must run in eager mode!"""
+    # TODO handle tuples and flat datasets as well.
+    for x in ds:
+        # Each individual tensor is converted to a known serializable type.
+        features = {key: np_value_to_feature(value.numpy()) for key, value in x.items()}
+        # All features are then packaged into a single Example object.
+        example = tf.train.Example(features=tf.train.Features(feature=features))
+
+        yield example.SerializeToString()
```
kvignesh1420, line 164: If only eager mode is supported, then maybe you can check whether tf is executing in eager mode and raise an exception if it isn't.

markemus: Hmm. It will raise an exception already if tf is not in eager mode, and if no exceptions are raised then it will run fine. In general I'm not in favor of explicitly adding exceptions like that because I've seen it cause unnecessary exceptions in other packages, but if that's the practice here I'm happy to add it.

kvignesh1420: Can we just ensure this in a different test case? The test would switch to graph mode and save the dataset and assert that the respective exception is raised.

markemus: Sure, I'll add that.

markemus: @kvignesh1420 I added the test you requested as `test_ensure_graph_fail()` in `tests/test_serial_ops.py`.

markemus:
@yongtang it looks like some of the checks are failing because it can't import yaml, which makes sense- I never added pyyaml as a dependency anywhere. How do I do that? I didn't see anything in setup.py and there doesn't seem to be a requirements.py anywhere.

yongtang:
@markemus I think `yaml` should not be a hard requirement for users. In tensorflow-io we purposely limit the requirement to only tensorflow itself with no other dependencies. You can place import yaml inside the function like:
https://github.com/tensorflow/io/blob/master/tensorflow_io/core/python/ops/audio_ops.py#L195-L201

so that module will only be loaded as needed.

yongtang:
@markemus Also, it looks like you are using an email that is not registered with your GitHub handle (` <c...@jhmi.edu>`). This is not a requirement though that causes the commit not being correctly linked by GitHub:
https://github.com/tensorflow/io/pull/1280/commits

See the author of the commits is not linked to your GitHub handle. When the PR is merged, you may not see your name appears on the contributors list correctly.

You can either change the git config to use the email you registered with GitHub, or add the email (` <c...@jhmi.edu>`) to GitHub account to allow correct linking of the author/commit.

markemus:
@yongtang I can do that, but the tests require yaml in order to run as well.

yongtang:
@markemus Test-specific modules can be added to the following place like other packages:
https://github.com/tensorflow/io/blob/master/.github/workflows/build.wheel.sh#L9

They are for test-only and tensorflow-io itself does not require those packages to run.

markemus:
@yongtang I made the changes you requested- the save_dataset and load_dataset functions import yaml, and I added yaml to the test build. Anything else we need?

yongtang:
@markemus The build is failing though it is caused by a separate issue and have been fixed in #1320. Can you rebase to the latest master? I think that will fix the build issue.

markemus:
@yongtang I rebased- unfortunately the tests are still failing, and it doesn't _seem_ to be caused by my code. Hard to tell though.

markemus:
@kvignesh1420 I'm not sure what's going on with the `test_ensure_graph_fail` test- it seems to be raising a different exception in Linux/macOS than in Windows. But in any case it does fail in graph mode on all systems, like expected, and I've included a warning in the module that graph mode is not supported. Do we really need a test to confirm that it's failing? If it doesn't fail, great, unsupported behavior can succeed sometimes. 

Would it be alright if I just removed the `test_ensure_graph_fail` test?

kvignesh1420:
@markemus I see that the warning is only in the form of a docstring. To avoid flaky behaviour we need in case of graph mode, it is better if we check for the value of `tf.executing_eagerly` and then immediately raise the exception with the message that graph mode is not supported.

markemus:
@kvignesh1420 okay, that seems reasonable, particularly since the natural exception seems to be OS dependent- I'll raise that exception and change the test to check for it.

markemus:
@kvignesh1420 I added the exception- the build is failing now but I see the main build is failing too for the same reason. I'll check back and try again in a day or two.

kvignesh1420:
@markemus can you please rebase with the master branch so that CI jobs are triggered again?

markemus:
@kvignesh1420 @yongtang rebase is done- tests aren't fully complete yet but look ok. Are we ready to merge?

markemus:
@kvignesh1420 @yongtang all tests pass now! Sorry for the false alarm yesterday.

kvignesh1420 (commented):
LGTM. @yongtang can you please take a look.

yongtang:
@markemus LGTM. Thanks for the contribution! 👍 

As a side note, have you seen my comment in https://github.com/tensorflow/io/pull/1280#issuecomment-791577572 ? If the email is not linked, then after the merge I think your author information will not show up in contributor's list. 

If you are ok with that we can go ahead and merge. However, I would suggest you to fix this, as we do respect contributor's hard work and want to make sure your contribution are correctly recognized in the repo.

markemus:
@yongtang thank you for the heads up! I checked the commit log and added both version of that email address to my account. Ready to merge when you are.

yongtang:
@markemus Merged. Thanks for your contribution! 👍

markemus:
@yongtang Awesome! Thank you and @kvignesh1420 for all your help :)

PR #1282, opened 2021-01-22
trigger benchmarks workflow only on commits

kvignesh1420:
The scheduled benchmarks workflow is failing with the following error:

```
Error: No commit information is found in payload: {
  "schedule": "0 12 * * *"
}
```

This PR removes the scheduler so that the workflow is triggered only when commits are made to master branch.


yongtang (approved):
LGTM

PR #1283, opened 2021-01-22
Adds AVRO_PARSER_NUM_MINIBATCH to override num_minibatches and logs the parsing time

StanfordMCP:
Added AVRO_PARSER_NUM_MINIBATCH to override num_minibatches. This is recommended to be set equal to the vcore request.


burgerkingeater (commented):
thanks. The change also includes logging the parsing time. Can you update the PR title/description?

ashahab:
@googlebot I consent.

yongtang:
@StanfordMCP @burgerkingeater Is there any additional information about the `AVRO_PARSER_NUM_MINIBATCH` and its impact? From the PR, the `AVRO_PARSER_NUM_MINIBATCH` and the aforementioned ` vcore request` are not very clear.

StanfordMCP:
@ashahab please comment

ashahab:
@yongtang .
AVRO_PARSER_NUM_MINIBATCH changes the number of minibatches an avro parser groups the batches for parallelism.
Avro parser's parallelism is identical to tfrecord's: https://github.com/tensorflow/tensorflow/blob/v2.2.0/tensorflow/core/util/example_proto_fast_parsing.cc#L1195
Pasting the comment below. Feel free to give us feedback on how best to communicate the intent of the environment variable.  
The intent is the users/running processes 
```
// TODO(lew): A big performance low-hanging fruit here is to improve
  //   num_minibatches calculation to take into account actual amount of work
  //   needed, as the size in bytes is not perfect. Linear combination of
  //   size in bytes and average number of features per example is promising.
  //   Even better: measure time instead of estimating, but this is too costly
  //   in small batches.
  //   Maybe accept outside parameter #num_minibatches?
  ```

**Comment by yongtang** [comment]
Thanks @ashahab @StanfordMCP for the background. I think in that case an attribute would be preferable than an environmental variable, as we rarely use env for parameter fine tuning for kernel ops.

Would you mind updating the PR to use attribute instead? The attribute can be passed in similar way as other attributes:

https://github.com/tensorflow/io/blob/a200cfb4ccb6ee61a7a3c1c422c055be713168a8/tensorflow_io/core/kernels/avro/parse_avro_kernels.cc#L406-L411

You can set the default value of the attribute, e.g.:
https://github.com/tensorflow/io/blob/a200cfb4ccb6ee61a7a3c1c422c055be713168a8/tensorflow_io/core/ops/avro_ops.cc#L276

so that user does not need to pass the attribute explicitly.

**Comment by kvignesh1420** [comment]
@ashahab @StanfordMCP any update on this PR?

**Comment by burgerkingeater** [comment]
@StanfordMCP can you fix the build issue?

**Comment by yongtang** [comment]
I think the build issue is unrelated and has been fixed in #1320. Can you rebase to the latest master? I think that will fix the build issue.

**Comment by ashahab** [comment]
@googlebot I consent.

File: tensorflow_io/core/python/experimental/parse_avro_ops.py [hunk 1]
```diff
@@ -130,6 +130,7 @@ def _parse_avro(
     dense_defaults=None,
     dense_shapes=None,
     name=None,
+    avro_num_minibatches=0,
```
burgerkingeater, line 133: 0 as default for avro_num_minibatches? doesn't it invalidate the check of avro_num_minibatches >= 0?

tensorflow_io/core/kernels/avro/parse_avro_kernels.cc
```diff
@@ -204,8 +207,13 @@ Status ParseAvro(const AvroParserConfig& config,
         minibatch_bytes = 0;
       }
     }
-    // 'special logic'
-    const size_t min_minibatches = std::min<size_t>(8, serialized.size());
+    if (avro_num_minibatches_) {
```
burgerkingeater, line 210: where does avro_num_minibatches_ get set?

burgerkingeater: nvm, found it.

burgerkingeater, line 210: if avro_num_minibatches_ is not set, what's the default value? can you comment it in the code?

tensorflow_io/core/kernels/avro/parse_avro_kernels.cc
```diff
@@ -204,8 +207,13 @@ Status ParseAvro(const AvroParserConfig& config,
...
+      VLOG(5) << "Overriding num_minibatches with " << avro_num_minibatches_;
+      result = avro_num_minibatches_;
+    }
+    // This is to ensure users can control the num minibatches all the way down
+    // to size of 1(no parallelism).
+    const size_t min_minibatches = std::min<size_t>(1, serialized.size());
```
burgerkingeater, line 216: can you also comment that avro_num_minibatches_ is [1, 64] in the op interface?

burgerkingeater (commented):
thanks for making the change! left few comments.

tensorflow_io/core/kernels/avro/parse_avro_kernels.cc
```diff
@@ -401,6 +421,11 @@ class ParseAvroOp : public OpKernel {
           dense_shapes_[d].dims() > 1 && dense_shapes_[d].dim_size(0) == -1;
     }
 
+    // Check that avro_num_minibatches is positive
```
burgerkingeater, line 424: nit: Check that avro_num_minibatches is **non-negative**

tensorflow_io/core/python/experimental/parse_avro_ops.py
burgerkingeater: nvm, i thought avro_num_minibatches needs to be positive based on the comment there, didn't read it through.

tensorflow_io/core/kernels/avro/parse_avro_kernels.cc
StanfordMCP: Updated

tensorflow_io/core/kernels/avro/parse_avro_kernels.cc
StanfordMCP: Updated

tensorflow_io/core/kernels/avro/parse_avro_kernels.cc
StanfordMCP: Updated

ashahab:
@googlebot I consent

burgerkingeater (approved):
Thanks, LGTM.

StanfordMCP:
@googlebot I consent.

burgerkingeater:
@yongtang the failure `GitHub CI / Test 3.9 Linux (pull_request) Failing after 22m — Test 3.9 Linux` is not related to this PR, but it cannot be merged until all check passes. Do you have any idea how we can proceed with this?

ashahab:
@googlebot I consent

yongtang:
@burgerkingeater I can override the merge. We will see if after the merge the issue still happens.

Issue #1236, opened 2020-12-17
Bazel caching for GitHub Actions workflows


wchargin:
It looks like your CI jobs build from scratch on every build, and thus
take many hours of workflow time (e.g., [7 hours for this workflow][w]).
These cause builds for the entire TF organization to queue.

Bazel has built-in support for [incremental remote caching][c]. This is
not a Travis-style giant opaque cache that needs to be sent to and from
the VM on each run. Instead, it’s a fine-grained key-value store that
stores the results for individual actions on a remote server. It has
transparent integration with GCS and can be used with just a backing
bucket.

Could you please investigate using this? It would make you happy, and it
would make us and everyone else happy.

We use this on TensorBoard. It speeds up our builds from ~75 minutes
(clean slate) to ~4 minutes (no-op). You can look at our [workflow file]
and our [bazelrc] for guidance.

[w]: https://github.com/tensorflow/io/actions/runs/427023817
[c]: https://docs.bazel.build/versions/master/remote-caching.html

[workflow file]: https://github.com/tensorflow/tensorboard/blob/48c0a2d814c755937f5950e251e8071199c99d38/.github/workflows/ci.yml#L55-L74
[bazelrc]: https://github.com/tensorflow/tensorboard/blob/48c0a2d814c755937f5950e251e8071199c99d38/ci/bazelrc#L27-L34

yongtang:
@wchargin We will be happy to use bazel cache as this will help speed up our CI process as well. At the moment the biggest blocker is that we don't have a GCS account/credential.

@angerson @ewilderj @martinwicke Wondering if you can help with GCS account/credential to speed up the build?

angerson:
FYI @theadactyl as well. I'm looking into the technical setup for a TF sub-project for this kind of thing.

angerson:
@yongtang I started our internal "official" process for setting up a new GCP project to contain a bucket for SIGs to use, but unfortunately there's a production freeze and my request won't be processed until the 5th of January. This should cover due-diligence for any privacy/security related policy issues, although it's still possible we'll run into other policy issues.

FYI @perfinion

yongtang:
@angerson Thanks for the update 👍

angerson:
I'm still working on this. It's going slow because there's a complex internal approval process for things like the ability to export service account credentials for bazel.

angerson:
Another update: I've been working on this with the following assumptions:

1. `bazel build` can only cache via a GCP service account credential file
[...]

PR #1287, opened 2021-01-29
Add bazel cache to speed up build for git push

yongtang:
This PR is part of the effort for #1236. It adds bazel cache to speed up build for push. We added a gcs bucket provided by tensorflow team so that the bazel build will point to the cache to gcs bucket.

The speed for Linux build is now 20 min (vs 1 hour 10 min in the past):
https://github.com/tensorflow/io/runs/1787903245?check_suite_focus=true

One limitation for this PR is that the cache is only available for git push, not for forked PRs (by outside contributors). 

To enable speed up for forked PRs, the gcs bucket will need to give read-only access to anonymous users. This will be inline with the way tensorboard team.

We will sync up to see this will be possible for the gcs bucket that was provided. Note tensorboard is done similarly in:

https://github.com/tensorflow/tensorboard/blob/48c0a2d814c755937f5950e251e8071199c99d38/ci/bazelrc#L27-L34

Thanks @angerson for the great help!

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -9,6 +9,8 @@ on:
       - master
 
 env:
+  GCP_CREDS: ${{ secrets.GCP_CREDS }}
+  EVENT_NAME: ${{ github.event_name }}
   BAZEL_OPTIMIZATION: --copt=-msse4.2 --copt=-mavx --compilation_mode=opt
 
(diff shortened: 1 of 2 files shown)


terrytangyuan (approved):
Exciting to see the huge speedup!

PR #1293, opened 2021-02-02
Only enable bazel caching writes for tensorflow/io github actions

yongtang:
This PR updates so that only GitHub actions run on
tensorflow/io repo will be enabled with bazel cache writes.

Without the updates, a focked repo actions will cause error.

Note once bazel cache read-permissions are enabled from gcs
forked repo will be able to access bazel cache (read-only).

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -10,6 +10,7 @@ on:
 
 env:
   GCP_CREDS: ${{ secrets.GCP_CREDS }}
+  REPO_NAME: ${{ github.repository }}
   EVENT_NAME: ${{ github.event_name }}
   BAZEL_OPTIMIZATION: --copt=-msse4.2 --copt=-mavx --compilation_mode=opt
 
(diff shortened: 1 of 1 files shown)


kvignesh1420 (approved):
@yongtang thanks.

kvignesh1420:
@yongtang to use the cache we need the GCS creds. However, as they are being stored as a secret of this repo, the forked repos shouldn't have the permission to use that secret right?

yongtang:
@kvignesh1420 Yes forked repo could not use the cache as secret is not accessible and will encounter invalid cache error if they do so. That is why the condition has been chanced to:
```
if [[ "${EVENT_NAME}" == "push" && "${REPO_NAME}" == "tensorflow/io" ]]; then
```

which limits the cache to "push only", and "from tensorflow/io repo only" (forked repo will skip).

kvignesh1420:
Okay!

terrytangyuan (approved):

PR #1295, opened 2021-02-02
Enable bazel remote cache for kokoro tests

yongtang:
Also enable bazel remote cache for kokoro tests


kvignesh1420 (commented):
@yongtang we are not using the gcs creds here as this is internal to google?

yongtang:
@kvignesh1420 we are using read-only mode here (just enabled today) for gcs.

kvignesh1420 (approved):

PR #1297, opened 2021-02-03
Rename tests

yongtang:
For legacy reasons (during the time from tensorflow v1 => v2),
we splits tests into two groups:
- tests/test_xxx.py which is used in v1 mode (and `tf.compat.v1.disable_eager_execution()` called)
- tests/test_xxx_eager.py which is used in v2 mode.

The reason was to prevent the mixture of v1 and v2 mode
as tensorflow may crash when two modes are mixed.

Over time it seems these two modes are not exactly follows.
For example, tests/test_azure.py is running in v2 mode
though they are groups in v1 during tests.

This PR changes the name so that:
- tests/test_xxx.py will always be run in v2 mode.
- tests/test_xxx_v1.py will always be run in v1 mode (and `tf.compat.v1.disable_eager_execution()` called)


terrytangyuan (approved):
LGTM

Issue #1298, opened 2021-02-03
arm64 support for M1


redmlr:
Dear TensorFlow-io team and community,
are there any plans for adding support for arm64 machines? I use a M1 mac and I'm not able to install the wheel. Anyone who's reading this, do you know any alternatives that I can use? My goal is to create a mel-spectogram from a .wav file. ps: M1 cannot use Scipy at this time.

yongtang:
@redmlr In order to get M1 support I think we needs
1. The M1 version of the tensorflow package (https://github.com/tensorflow/tensorflow/issues/44751 provides a way to build tensorflow on M1, though no release package)
2. A M1 build machine (ideally with GitHub Actions) (https://github.com/actions/runner/issues/805)

I don't have access to M1 laptop yet, may take a look after I get one.

redmlr:
@yongtang Thank you for answering,
There’s a m1 accelerated version of tensorflow in apples tensorflow repo, which is in alpha version. I’m running the latest version of this and it runs fine but tensorflow-io is a separate library which won’t install.

yongtang:
@redmlr The tensorflow-io package will need to be re-compiled for all functionalities to work. Though if you are only looking for functions that does not requires C++ code (only need python code) then it should be possible to just package the source tree. 

Which function (or list of functions) you intend to use?

redmlr:
@yongtang I'm planning to use the following: 

`tfio.experimental.audio.trim
tfio.experimental.audio.spectrogram
tfio.experimental.audio.melscale
tfio.experimental.audio.dbscale
tfio.experimental.audio.freq_mask
tfio.experimental.audio.time_mask
`

redmlr:
Thank you @yongtang !
I'm using not that long.. What does just package the source tree mean and how do I do this? I already downloaded the latest release of io and changed the desired file (__init__.py) as mentioned in #1302

yongtang:
@redmlr With PR #1302 in you can build the wheel with:
```
python3 setup.py -q bdist_wheel
```

The wheel file will be located under `dist` directory.

You can then install the wheel with:
```
python3 -m pip install --no-deps <wheel-file-build-from-last-step>
```

That will at least allows you to use python only functions in tensorflow-io.

Note `--no-deps ` will ignore the dependency of `tensorflow` package (as the official tensorflow package is different from the apple's tensorflow package).

Note normally the wheel file needs to pass `--data bazel-bin`
```
[...]

PR #1302, opened 2021-02-05
Avoid error if plugins .so module is not available

yongtang:
This PR raises a warning instead of an error in case plugins .so module is not available, so that tensorflow-io package can be at least partially used with python-only functions.

This PR is related to #1298

### DIFF
--- a/tensorflow_io/core/python/ops/__init__.py
+++ b/tensorflow_io/core/python/ops/__init__.py
@@ -18,6 +18,7 @@ import os
 import ctypes
 import sys
 import inspect
+import warnings
 import types
 
 import tensorflow as tf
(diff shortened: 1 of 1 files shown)


tensorflow_io/core/python/ops/__init__.py
```diff
@@ -95,5 +96,8 @@ def __dir__(self):
     plugin_ops = _load_library("libtensorflow_io_plugins.so", "fs")
 except NotImplementedError as e:
     # Note: load libtensorflow_io.so imperatively in case of statically linking
-    core_ops = _load_library("libtensorflow_io.so")
-    plugin_ops = _load_library("libtensorflow_io.so", "fs")
+    try:
+        core_ops = _load_library("libtensorflow_io.so")
+        plugin_ops = _load_library("libtensorflow_io.so", "fs")
```
burgerkingeater, line 101: should we reload  `plugin_ops = _load_library("libtensorflow_io.so", "fs")` only?

yongtang: @burgerkingeater This part is specific to CentOS 7. In CentOS 7 we used static linking, that prompt us to combine `libtensorflow_io.so` and `libtensorflow_io_plugins.so` into one `libtensorflow_io.so` as static linking does not work with two `.so`. Notice the comment about:
```
Note: load libtensorflow_io.so imperatively in case of statically linking
```

When we have one `libtensorflow_io.so` file to load, due to tensorflow's limitations, it needs to be loaded with full symbols before file system symbols. So we use `core_ops = _load_library("libtensorflow_io.so")` before fs load.
This also means tensorflow-io on CentOS 7 can not have api backward-compatibility like other linux dist. This is a limitation that may only be resolved in CentOS 8 or higher.

kvignesh1420 (approved):

PR #1303, opened 2021-02-06
Adds Avro Benchmark Tests

StanfordMCP:
-Adds Benchmark Tests for various file formats
-Tests developed by @fraudies


fraudies:
Thanks for putting this up.

A few considerations:
- Tests may run in CLI and thus can take a lot of resources; do you know what's the total runtime for these "benchmark" tests? I can't remember but if it takes too long we may need to annotate so they don't run in CLI or only run sometimes in CLI
- This posts another project; the "avro data generator" here; to organize source code it may make more sense to put that into a separate open source project with it's own test cases and than just depend on that here -- would simplify the review as well. An alternative could be to modify the benchmarks and just take an "off-the shelf" random data generator for avro -- there are a few projects out there

StanfordMCP:
@googlebot I fixed it.

yongtang:
@StanfordMCP You may also need to fix the commit's author/email for all commits:
https://github.com/tensorflow/io/pull/1303/commits

Notice that the first 4 commits does not link with your GitHub handle -- that likely is caused by the email you used in the previous commits.

StanfordMCP:
@yongtang I updated the email but I think it changed the hashes. I'll close this PR and resubmit.

Issue #1275, opened 2021-01-16
Inconsistent documentation in tensorflow.org/io


kvignesh1420:
![image](https://user-images.githubusercontent.com/25393945/104816811-755cc800-5843-11eb-81d6-765c22339b5c.png)

The docs for https://www.tensorflow.org/io/api_docs/python/tfio/experimental/IODataset seems to be broken.
However, the [from_parquet docstring](https://github.com/tensorflow/io/blob/ccb321cbf2fc0323e9863bacc2b34b56c6b1d25d/tensorflow_io/core/python/ops/io_dataset.py#L260) seem to be fine:

kvignesh1420:
cc: @MarkDaoust

kvignesh1420:
cc: @MarkDaoust can you please take a look?

PR #1305, opened 2021-02-08
Fix docstring.

MarkDaoust:
This is breaking everything below it.

https://www.tensorflow.org/io/api_docs/python/tfio/experimental/IODataset

Fixes: https://github.com/tensorflow/io/issues/1275


kvignesh1420 (approved):
thanks @MarkDaoust

kvignesh1420:
@MarkDaoust can you please push the docs? Thanks for the fix.

PR #1308, opened 2021-02-09
Add @com_google_absl//absl/strings:cord

yongtang:
While trying to build tensorflow-io against tf-nightly (not tensorflow package), I noticed there were some additional fixes that will be needed to compile.

Since they are additional includes and will not impact existing build, and we will update to next version of tensorflow soon, it makes sense to have the update.

### DIFF
--- a/tensorflow_io/core/kernels/text_kernels.cc
+++ b/tensorflow_io/core/kernels/text_kernels.cc
@@ -16,10 +16,11 @@ limitations under the License.
 #include "tensorflow/core/framework/op_kernel.h"
 #include "tensorflow/core/lib/io/buffered_inputstream.h"
 #include "tensorflow_io/core/kernels/io_stream.h"
-
 #if defined(_MSC_VER)
 #include <io.h>
 #define STDIN_FILENO _fileno(stdin)
+#else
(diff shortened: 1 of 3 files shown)


kvignesh1420 (commented):
@yongtang can you please provide any references regarding this change?

yongtang:
@kvignesh1420 I am testing this change locally by:
```
./python3 -m pip install -U tf-nightly
./configure.sh
```

which overrides the default install of `tensorflow>=2.4.0`.

After that, run bazel like before (now headers are pointing to tf-nightly):
```
```

Without the PR change, the build will fail in several places. The build failure is likely caused by tensorflow's master branch change recently (showed up in tf-nightly).

kvignesh1420:
@yongtang I think we should add the tfio build verification step to `API-Compatibility` Github-Actions workflows as well. We can thus identify these kinds of issues without human intervention. What do you think?

yongtang:
@kvignesh1420 Yes I think we can add an additional test to test against daily tf-nightly so that any build failures could be detected. The test can run everyday.

kvignesh1420 (approved):
thanks for the info @yongtang.

PR #1311, opened 2021-02-13
Modify --plat-name for macosx wheels

kvignesh1420:
This PR updates the `--plat-name` from `macosx_10_13_x86_64` to `macosx_10_15_x86_64` while building the tfio wheel as per the github actions env.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -157,7 +157,7 @@ jobs:
           set -x -e
           python -m pip install -U wheel setuptools
           python --version
-          python setup.py --data bazel-bin -q bdist_wheel --plat-name macosx_10_13_x86_64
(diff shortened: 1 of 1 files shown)


yongtang:
@kvignesh1420 While GitHub Actions have macos 10.15, we actually limit the macOS binary compatibility to 10.14 (used to be 10.13):
https://github.com/tensorflow/io/blob/master/tools/build/configure.py#L122-L123

In GitHub Actions we limit to 10.13 in the past to allow the binaries we built could run on lower versions of macOS (not just 10.15).

If we want to change the `--plat-name`, maybe we can change to 10.14 instead (and drop the 10.13 support)?

kvignesh1420:
@yongtang yes sure, made the change.

yongtang (approved):

PR #1314, opened 2021-02-25
Update to enable python 3.9 building on Linux

yongtang:
Since tf-nightly 3.9 is available on Linux, this PR enables our testing for python 3.9 (Linux only)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -281,18 +281,10 @@ jobs:
   linux-test:
     name: Test ${{ matrix.python }} Linux
     needs: linux-wheel
-    runs-on: ${{ matrix.os }}
+    runs-on: ubuntu-20.04
     strategy:
       matrix:
-        os: [ubuntu-18.04, ubuntu-20.04]
         python: ['3.7', '3.8', '3.9']
-        exclude:
-          - os: ubuntu-18.04
-            python: '3.8'
-          - os: ubuntu-18.04
-            python: '3.9'
-          - os: ubuntu-20.04
-            python: '3.7'
     steps:
       - uses: actions/checkout@v2
       - uses: actions/download-artifact@v1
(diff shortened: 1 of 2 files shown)


burgerkingeater (approved):
thanks!

.github/workflows/build.yml
```diff
@@ -285,10 +285,12 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-18.04, ubuntu-20.04]
-        python: ['3.7', '3.8']
+        python: ['3.7', '3.8', '3.9']
         exclude:
           - os: ubuntu-18.04
             python: '3.8'
+          - os: ubuntu-18.04
+            python: '3.9'
```
kvignesh1420, line 293: @yongtang is there a reference for this unstable combination?

yongtang:
@kvignesh1420 to explain a little about the GitHub Actions setup: Previously when we do testing (when Ubuntu 20.04 was not released yet), we use 18.04 as the default test machine and test again 3.5, 3.6, 3.7. Because 18.04 does not comes with 3.5 and 3.7, we installed `deadsnake`'s PPA on ubuntu 18.04.

After ubuntu20.04 was out, we used ubuntu 20.04's system python 3.8 to test 3.8 as well. At that time, we use a matrix of (Ubuntu 18.04 + 20.04) x (python 3.6 + 3.7 + 3.8). However, GitHub Actions was not smart enough so we have to exclude several duplicate combinations (Ubuntu 18.04+3.8, Ubuntu 20.04+3.6, Ubuntu 20.04+3.7 was excluded.

That is why you see the line of `excluded`.

Now I take a look at the workflow again, and realized that actually python3.6, python 3.7, python 3.8, python 3.9 are all available on ubuntu 20.04 (either from system default python 3.8, or deadsnake PPA's 3.6, 3.7). For that reason looks like we don't need ubuntu 18.04 at all.

From the above I updated the script again and pushed the PR. Please take a look.

kvignesh1420 (approved):
Thanks for the info @yongtang.

PR #1315, opened 2021-02-26
Experimental: Add initial wavefront/obj parser for vertices

yongtang:
This PR is an early experimental implementation of wavefront obj
parser in tensorflow-io for 3D objects.
This PR is the first step to obtain raw vertices in float32
tensor with shape of `[n, 3]`.

Additional follow up PRs will be needed to handle meshs with
different shapes (not sure if ragged tensor will be a good fit
in that case)

Some background on obj file:
Wavefront (obj) is a format widely used in 3D (another is ply)
modeling (http://paulbourke.net/dataformats/obj/). It is simple
(ASCII) with good support for many softwares. Machine learning
in 3D has been an active field with some advances such as
PolyGen (https://arxiv.org/abs/2002.10880)

Processing obj files are needed to process 3D with tensorflow.

In 3D the basic elements could be vertices or faces. This PR
tries to cover vertices first so that vertices in obj file
can be loaded into TF's graph for further processing within
graph pipeline.


burgerkingeater:
do you have any introduction to wavefront format? Can you link it in the PR given it's a less known format.

yongtang:
@burgerkingeater obj file is a simple ASCII format for 3D models (http://paulbourke.net/dataformats/obj/). I have updated the PR with additional description. Please take a look.

tensorflow_io/core/python/experimental/image_ops.py
```diff
@@ -208,3 +208,17 @@ def decode_jp2(contents, dtype=tf.uint8, name=None):
...
     """
     return core_ops.io_decode_jpeg2k(contents, dtype=dtype, name=name)
+
+
+def decode_obj(contents, name=None):
+    """
+    Decode a Wavefront (obj) file into a float32 tensor.
+
+    Args:
+      contents: A `Tensor` of type `string`. 0-D. The Wavefront (obj) file.
```
kvignesh1420, line 218: @yongtang I think
`contents: A Tensor of type string. 0-D.  Wavefront (obj) file.` 

be replaced with
`contents: A 0 dimensional Tensor of type string, i.e the content of the Wavefront (.obj) file.`

kvignesh1420 (changes requested):
@yongtang thanks for adding the support for the wavefront `.obj` file format. Please check the minor docstring change.

yongtang:
@kvignesh1420 Thanks for the review. The PR has been updated. Please take a look.

kvignesh1420 (approved):
LGTM.

PR #1320, opened 2021-03-07
update `protobuf` version to `3.11.4` to match tensorflow-nightly

vnvo2409:
There is a mismatch version of protobuf between `tf-nightly` and `tf-io` ( and there are leaks symbols somewhere I think ). Commit https://github.com/tensorflow/tensorflow/commit/e71367f826ed32e8a7a3e78d5f35aded180c9c72 updated `protobuf` version of Tensorflow to `3.11.4`. This PR does the same things for `tf-io`.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -12,11 +12,14 @@ cuda_configure(name = "local_config_cuda")
 
 http_archive(
     name = "com_google_protobuf",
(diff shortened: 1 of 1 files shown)


kvignesh1420 (approved):

yongtang:
Thanks @vnvo2409 for the fix! 👍

PR #1326, opened 2021-03-13
WIP: switch `pubsub_kernels` to use `google_cloud_cpp` instead of raw `grpc`

vnvo2409:


vnvo2409:
We have to wait for Tensorflow to remove `google-cloud-cpp` before continuing.

PR #1327, opened 2021-03-14
switch mnist dataset mirror to a more reliable one

vnvo2409:
Recently, our CI ( `test_documentation` ) is falling because of 503 from `http://yann.lecun.com/exdb/mnist/`. So I think we should switch to a more reliable mnist mirror.

The example is also passed through `black` and `pyupgrade` so there are some unrelated changes.


kvignesh1420 (approved):
Thanks @vnvo2409

PR #1334, opened 2021-03-21
Rename tf.experimental.columnar to tf.experimental.avro

oliverhu:
Rename tfio.experimental.columnar to tfio.experimental.avro, as avro is not columnar store.


terrytangyuan (approved):

burgerkingeater:
@yongtang can you please comment? I think the suggestion to keep it under columnar is from https://github.com/tensorflow/io/pull/879#issuecomment-611598028

yongtang:
@oliverhu columnar is to categorize avro into the same category as parquet/feather/csv, as they essentially are column data. The intention is to limit the number of top level python module as the current number is growing too big.

Note though under tfio.experimental.avro might be fine as `tfio.experimenta` is the top level module, though I think it will be hard to place avro as `tfio.avro` when it is moved out of experimental. Maybe something like tfio.experimental.columnar.avro? (note we are currently also reviewing the APIs and want to come up with a consolidated way of naming. /cc @kvignesh1420 )

oliverhu:
@yongtang I understand your pov, it is just Avro is not columnar.. it is row based like CSV, so we can call it `tfio.experimental.row.avro`. But I don't think there are many row based schema nowadays in data processing world. My suggestion is to keep Avro separate and group others (parquet, ORC, arrow etc) into columnar.

kvignesh1420:
@yongtang I am currently working on the PR which restructures the library into a cleaner layout. Will keep you guys updated on it.

yongtang:
@kvignesh1420 Any insight on avro?

kvignesh1420:
@yongtang as parquet/avro/csv etc are file formats, we can have a common api layout for them, such as:
```
tfio.experimental.formats.avro
tfio.experimental.formats.parquet 
etc
```
 WDYT?

oliverhu:
Not sure if `tfio.formats` solves the problem here. Almost everything at top level is currently a format:

- arrow
- audio
- bigquery
- ignite
- image

etc.

terrytangyuan:
Perhaps the module structure should try to be flat and then in the reference we try to categorize them (one may fit in multiple categories too).

yongtang:
I think image, audio, and video are kind of special as they consists of a collections of formats.

oliverhu:
That is true...on a second thought, perhaps organize in 

- tfio.generic (parquet, avro, arrow) <- these are generic formats that could host anything
- tfio.audio (mp3) 
- tfio.streaming (HLS)
- tfio.cloud (s3, adls etc)

...

kvignesh1420:
I have raised the initial PR https://github.com/tensorflow/io/pull/1340 for refactoring the API's. Subsequent PR's will perform the cleanup of `tensorflow_io/kafka`, `tensorflow_io/arrow`, `tensorflow_io/bigquery`, `tensorflow_io/ignite` and then we can group the API's based on generic formats or specialized formats etc. I think we can close this PR as of now and discuss after the initial cleanup/refactor is done. Then as @terrytangyuan said, we will have a flat ops structure and we can group them as necessary into public API's.

yongtang:
@kvignesh1420 bigquery is used heavily, we may want to be a little careful about API changes. /cc @vlasenkoalexey

kvignesh1420:
@yongtang I was thinking of just restructure the layout of the `bigquery` folder within core and expose the API's in the same way as before. However, this begs the question that what comes under core and what doesn't?

oliverhu:
...I'll close the ticket after we finish the discussion.

PR #1336, opened 2021-03-22
Bump abseil-cpp to 6f9d96a1f41439ac172ee2ef7ccd8edf0e5d068c

yongtang:
This PR bumps abseil-cpp to 6f9d96a1f41439ac172ee2ef7ccd8edf0e5d068c
to fix the build issue.

See related changes in  tensorflow/tensorflow@1c9eeb9

Signed-off-by: Yong Tang <yong.tang.github@outlook.com>

Issue #1335, opened 2021-03-22
decode_flac not working in tfio-nightly 0.18.0 and tf-nightly 2.5.0 


lminer:
I'm trying to get the nightly version of tensorflow 2.5.0-dev to work with the nightly version of tensorflow-io.

However, when running a simply `tf.data` dataset that incorporates `tfio.audio.decode_flac`, I get a `NotImplementedError`.

Versions are:
- tb-nightly-2.5.0a20210322 
- tensorflow-io-nightly-0.18.0.dev20210318195435
- tf-estimator-nightly-2.5.0.dev2021032201
- tf-nightly-2.5.0.dev20210322

Error is below:

```
../lib/python3.7/site-packages/tensorflow/python/data/ops/dataset_ops.py:1931: in map
    preserve_cardinality=True)

../lib/python3.7/site-packages/tensorflow/python/data/ops/dataset_ops.py:4520: in __init__
    use_legacy_function=use_legacy_function)

../lib/python3.7/site-packages/tensorflow/python/data/ops/dataset_ops.py:3706: in __init__
    self._function = fn_factory()

../lib/python3.7/site-packages/tensorflow/python/eager/function.py:3130: in get_concrete_function
    *args, **kwargs)

../lib/python3.7/site-packages/tensorflow/python/eager/function.py:3095: in _get_concrete_function_garbage_collected
    graph_function, _ = self._maybe_define_function(args, kwargs)

../lib/python3.7/site-packages/tensorflow/python/eager/function.py:3439: in _maybe_define_function
    graph_function = self._create_graph_function(args, kwargs)

../lib/python3.7/site-packages/tensorflow/python/eager/function.py:3284: in _create_graph_function
    capture_by_value=self._capture_by_value),

../lib/python3.7/site-packages/tensorflow/python/framework/func_graph.py:998: in func_graph_from_py_func
    func_outputs = python_func(*func_args, **func_kwargs)

../lib/python3.7/site-packages/tensorflow/python/data/ops/dataset_ops.py:3681: in wrapped_fn
    ret = wrapper_helper(*args)

../lib/python3.7/site-packages/tensorflow/python/data/ops/dataset_ops.py:3611: in wrapper_helper
    ret = autograph.tf_convert(self._func, ag_ctx)(*nested_args)
[...]

PR #1339, opened 2021-03-23
Release nightly even if test fails

yongtang:
Our tests on some platforms have not been very stable recently. That prevents the nightly releases. For example, PR #1336 may potentially have #1335 addressed as well but we don't have tensorflow-io-nightly due to test failure.

This PR removes the dependency so that tensorflow-io-nightly can be released even if the tests fails. (as tensorflow-io-nightly is supposed to be not the same grade as final release anyway).


kvignesh1420 (approved):
LGTM.

PR #1343, opened 2021-03-28
improvements for `s3` environements variables

vnvo2409:
116dc39: Now, we have to set 3 envs `S3_DISABLE_MULTI_PART_DOWNLOAD`, `S3_MULTI_PART_UPLOAD_CHUNK_SIZE` and `S3_MULTI_PART_DOWNLOAD_CHUNK_SIZE` before importing `tensorflow-io` if we want to change the default behavior. This PR defers the reading:
- `S3_DISABLE_MULTI_PART_DOWNLOAD` now should be set before the first calling to `s3` filesystem ( read in `GetS3Client` )
- `S3_MULTI_PART_UPLOAD_CHUNK_SIZE` should be set before uploading something to `s3` ( read in `GetTransferManager` in `Aws::Transfer::TransferDirection::UPLOAD` ).
- `S3_MULTI_PART_DOWNLOAD_CHUNK_SIZE ` should be set before reading something from `s3` ( read in `GetTransferManager` in `Aws::Transfer::TransferDirection::DOWNLOAD` ).

6e9e86d: This commit use [`OverrideEndpoint`](https://github.com/aws/aws-sdk-cpp/blob/b343488fc5351d7278c0b9482988321c9b4387bf/aws-cpp-sdk-s3/source/S3Client.cpp#L185-L203) of `s3_client` instead of forcing users to set `S3_ENDPOINT` and `S3_USE_HTTPS`. `OverrideEndpoint` supports both with and without `uri` ( in the latter case it will fallback to `S3_USE_HTTPS`). We could eventually remove `S3_USE_HTTPS` and `S3_VERIFY_SSL` since users might accidentally set `S3_USE_HTTPS=0` in a production environment which is extremely dangerous. ( we could remove those envs in this PR I think )

### DIFF
--- a/tensorflow_io/core/plugins/s3/s3_filesystem.cc
+++ b/tensorflow_io/core/plugins/s3/s3_filesystem.cc
@@ -241,6 +241,9 @@ static void GetS3Client(tf_s3_filesystem::S3File* s3_file) {
             tf_s3_filesystem::AWSLogSystem::ShutdownAWSLogging();
           }
         });
+    int temp_value;
+    if (absl::SimpleAtoi(getenv("S3_DISABLE_MULTI_PART_DOWNLOAD"), &temp_value))
+      s3_file->use_multi_part_download = (temp_value != 1);
   }
 }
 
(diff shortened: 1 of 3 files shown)


vnvo2409:
In a related event, I've just sent a PR to `localstack` to fix the `416` status code. After this PR https://github.com/localstack/localstack/pull/3797 is merged, we could enable `S3_DISABLE_MULTI_PART_DOWNLOAD`, testing against productions is already fine.

yongtang (approved):
PR looks good, thanks for the improvement!

vnvo2409:
@yongtang should we remove these envs in this PR ?

yongtang:
@vnvo2409 I think we can remove the env in this PR. It will not impact users in product environment anyway.

vnvo2409:
Good news! My PR localstack/localstack#3797 is merged.

@yongtang Could you merge this PR ? Linux tests have passed.

yongtang:
> PR localstack/localstack#3797 is merged.

@vnvo2409 Great effort!

fsonntag:
@yongtang @vnghia 

We just updated our TF, and realized that `S3_VERIFY_SSL` is gone now. This is a breaking change and we can't use it anymore now. I.e. our dev instances of Ceph, where we connect via the S3 interface, don't have SSL setup properly. Using the `http` endpoint doesn't help.

Is it possible to bring this functionality back? I think it should be up to the user if/how they want to configure the S3 connection.  
If you're okay with it, I can send a PR.

yongtang:
@fsonntag PR are welcomed👍

fsonntag:
Thanks, @yongtang . Opened one: https://github.com/tensorflow/io/pull/1790

PR #1346, opened 2021-03-29
bump `aws` dependencies

vnvo2409:
- Bump `aws-sdk-cpp` to `1.8.172`
- Bump `localstack` to `0.12.9`

PR #1347, opened 2021-03-29
remove unnecessary code in `s3`

vnvo2409:
While playing around with #1346 , I found some interesting things. Let see if we got any segfault here.

---

1c2f08b: only needed if we use a custom memory management.

---

65aa6a3: 

Remove `aws_crypto`.

The reason why it was written in the first place was because at that time, `aws-sdk-cpp` did not support `openssl/boringssl` ( in 2015 based on the copyrights of https://github.com/tensorflow/tensorflow/blob/7aa4dd01e28e6aaf4b452f5204ff950305302e1b/tensorflow/core/platform/s3/aws_crypto.h#L1 ).

However, in 2016, `aws-sdk-cpp` was completely refactored to support `openssl/boringssl` via commit aws/aws-sdk-cpp@907dcdc .

Therefore, we don't need `aws_crypto` anymore.


vnvo2409:
Close, we still got segfault :'(

PR #1348, opened 2021-03-31
modify env var for loading modular file systems

kvignesh1420:
This change is a result of https://github.com/tensorflow/tensorflow/commit/aa35d4755409474f6619c68af1bfca168db24518 and https://github.com/tensorflow/tensorflow/commit/f8aee66b8c227da41273a8d3cef558d9dcf431de
The former approach requires the users to install `tensorflow-io` on a priority basis to continue normal operations and set `TF_ENABLE_LEGACY_FILESYSTEM` env var to use TensorFlow's legacy implementation. Thus, to ensure backwards compatibility, we are using a different env var (`TF_USE_MODULAR_FILESYSTEM`) and assume it to be unset by default. This will enable the users to load the tfio modular fs plugins only when the env var `TF_USE_MODULAR_FILESYSTEM` is set.

cc: @mihaimaruseac @yongtang @vnvo2409

### DIFF
--- a/tensorflow_io/core/plugins/file_system_plugins.cc
+++ b/tensorflow_io/core/plugins/file_system_plugins.cc
@@ -29,6 +29,7 @@ void TF_InitPlugin(TF_FilesystemPluginInfo* info) {
                                              sizeof(info->ops[0])));
   tensorflow::io::az::ProvideFilesystemSupportFor(&info->ops[0], "az");
   tensorflow::io::http::ProvideFilesystemSupportFor(&info->ops[1], "http");
+  // Load plugins only when the environment variable is set
   if (load_plugin == "true" || load_plugin == "1") {
     tensorflow::io::s3::ProvideFilesystemSupportFor(&info->ops[2], "s3");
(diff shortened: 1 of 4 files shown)


kvignesh1420:
@mihaimaruseac was the tensorflow-nightly package uploaded to pypi with this commit: https://github.com/tensorflow/tensorflow/commit/f8aee66b8c227da41273a8d3cef558d9dcf431de?

kvignesh1420:
The plugins from `tfio` are now being loaded only when the `TF_USE_MODULAR_FILESYSTEM` env var is set. We can discuss this during the SIG call.

yongtang:
it looks like tf-nightly on https://pypi.org/project/tf-nightly/#files includes the commit. However, python 3.9 on linux is missing so we may have to wait for one more day to get the tf-nightly with python 3.9 on linux.

yongtang:
Looks like new tf-nightly has been released with python 3.9 on Linux. Think we can merge this PR now.

yongtang (approved):

PR #1359, opened 2021-04-11
Split gs plugin out as an independent wheel

yongtang:
This PR splits gs plugin out as an independent wheel
so that it is possible for tensorflow to depends on
gs file system only without breaking existing users.

This is discussed in our SIG IO meeting.

With this PR there will be two wheels:
1. tensorflow-io
2. tensorflow-io-plugin-gs

To use gs file system, user will use `import tensorflow_io_plugin_gs`
explicitly so that existing users will not be impacted (if they are
still using tensorflow's default gs file system).


yongtang:
cc @mihaimaruseac FYI. As was discussed in the meeting, this gs plugin change will likely only to be added to tensorflow at later versions (2.6 or higher).

kvignesh1420 (commented):
@yongtang had a first pass through this PR and was wondering if we should have a separate directory for `plugins` at the root of the project. So the structure would be `plugins/tensorflow_io_plugin_gs`, `plugins/tensorflow_io_plugin_hdfs` etc, instead of overloading the root structure. WDYT?

yongtang:
@kvignesh1420 The reason `tensorflow_io_plugin_gs` has been placed at the top level is because gs will be the only plugin that will be outside of the tensorflow-io wheel package (due to backward compatibility reasons with many tutorials). the hdfs and aws plugin will always be part of the tensorflow-io package so there is no need to split out. At some point (likely 2-3 or even longer tensorflow releases) the gs plugin may be folded to tensorflow-io wheel package like other plugins as well.

tensorflow_io_plugin_gs/core/python/ops/__init__.py
```diff
@@ -0,0 +1,78 @@
...
+        filenames.append(f)
+    # Function to load the library, return True if file system library is loaded
+    if lib == "op":
+        load_fn = tf.load_op_library
+    elif lib == "dependency":
+        load_fn = lambda f: ctypes.CDLL(f, mode=ctypes.RTLD_GLOBAL)
+    elif lib == "fs":
+        load_fn = lambda f: tf.experimental.register_filesystem_plugin(f) is None
+    else:
+        load_fn = lambda f: tf.compat.v1.load_file_system_library(f) is None
```
kvignesh1420, line 58: since the functionality is dedicated to loading the `fs` plugin, the extra conditions can be removed for simplicity.

yongtang:
Thanks @kvignesh1420 . The PR has been updated. Please take a look.

kvignesh1420:
@yongtang  Just wanted to confirm the functionality by seeing the test results on linux. However, since github actions are failing, I triggered them again. Otherwise LGTM.

tensorflow_io_plugin_gs/core/file_system_plugin_gs.cc
```diff
@@ -0,0 +1,36 @@
+/* Copyright 2019 The TensorFlow Authors. All Rights Reserved.
```
burgerkingeater, line 1: please keep the date updated.

tensorflow_io_plugin_gs/core/file_system_plugin_gs.cc
```diff
@@ -0,0 +1,36 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+==============================================================================*/
+
+#include "tensorflow_io_plugin_gs/core/file_system_plugin_gs.h"
+
+#include "absl/strings/ascii.h"
+
+void TF_InitPlugin(TF_FilesystemPluginInfo* info) {
```
burgerkingeater, line 20: is this file copied from somewhere? if so can we document the source ?

tensorflow_io_plugin_gs/core/file_system_plugin_gs.h
```diff
@@ -0,0 +1,38 @@
+/* Copyright 2019 The TensorFlow Authors. All Rights Reserved.
```
burgerkingeater, line 1: ditto on the date.

tensorflow_io_plugin_gs/core/python/ops/__init__.py
```diff
@@ -0,0 +1,71 @@
+# Copyright 2018 The TensorFlow Authors. All Rights Reserved.
```
burgerkingeater, line 1: ditto

vnvo2409:
I would like to ask if it is possible to remove `google-cloud-cpp` out of Tensorflow dependencies ? It is used only by `tensorflow/c/experimental/filesystem/gcs/` ( I think we could remove this directory too ? ) and it causes some segfault when I tried updating `google-cloud-cpp` for `pubsub` and `bigquery` stuff.

tensorflow_io_plugin_gs/core/file_system_plugin_gs.cc
yongtang: @burgerkingeater Thanks. updated.

tensorflow_io_plugin_gs/core/file_system_plugin_gs.h
yongtang: @burgerkingeater  updated.

tensorflow_io_plugin_gs/core/python/ops/__init__.py
yongtang: @burgerkingeater  updated.

tensorflow_io_plugin_gs/core/file_system_plugin_gs.cc
yongtang: @burgerkingeater The API is defined in tensorflow/tensorflow/c/experimental/filesystem/filesystem_interface.h though we add the implementation here. I have updated the comment of the code.

yongtang:
@burgerkingeater I have updated the PR with comments addressed. Please take a look.

burgerkingeater (approved):

yongtang:
@vnvo2409 the plan from SIG IO meeting is that we will take multiple steps:
1. We create a tensorflow_io_plugin_gs (thin) wheel that only consists of gs plugin.
2. In next version of tensorflow, tensorflow will have a dependency of tensorflow_io_plugin_gs (thin) wheel and the built-in gcs plugin in tensorflow will be replaced and removed.
3. After tensorflow depending on tensorflow_io_plugin_gs (thin) wheel, we will remove the google-cloud-cpp and any gcs related libraries from tensorflow
4. After that the only google-cloud-cpp will be in tensorflow_io_plugin_gs (which is in tensorflow/io repo but is a separate thin wheel), then we can update the google-cloud-cpp here.

/cc @mihaimaruseac @vnvo2409

vnvo2409:
> After tensorflow depending on tensorflow_io_plugin_gs (thin) wheel, we will remove the `google-cloud-cpp` and any gcs related libraries from tensorflow

Indeed, current [cloud filesystem](https://github.com/tensorflow/tensorflow/tree/master/tensorflow/core/platform/cloud) in TF does not use `google-cloud-cpp`. So removing `google-cloud-cpp` won't cause any breakage here ( but we will have to remove `tensorflow/c/experimental/filesystem/plugins/gcs` as well. )

yongtang:
> Indeed, current cloud filesystem in TF does not use google-cloud-cpp. 

@vnvo2409 If current cloud filesystem in TF does not use `google-cloud-cpp`, maybe we can give it a try to not include google-cloud-cpp in `libtensorflow_framework.so`? That should be enough to allow us to upgrade `google-cloud-cpp` in tensorflow/io side.

vnvo2409:
I've just checked and `google-cloud-cpp` isn't included in `libtensorflow_framework.so`. Apologies for the back and forth.

mihaimaruseac:
Note that the plan for the plugins is to have them in separate wheels with separate version numbers. This would allow people to only install the plugin they need (and thus get bandwidth savings).

yongtang:
@mihaimaruseac @vnvo2409 @kvignesh1420 @terrytangyuan The gcs-only plugin wheel is available in:
 
https://pypi.org/project/tensorflow-io-plugin-gs-nightly/0.18.0.dev20210426025327/

(It is highly for now. We will release a tensorflow-io-plugin-gs formal release soon).

PR #1370, opened 2021-04-18
Fix Kafka test setup issue by adding wait period

yongtang:

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -289,8 +289,8 @@ jobs:
           bash -x -e tests/test_azure/start_azure.sh
           bash -x -e tests/test_sql/sql_test.sh
           bash -x -e tests/test_gcloud/test_gcs.sh gcs-emulator
-          bash -x -e tests/test_pulsar/pulsar_test.sh
           bash -x -e tests/test_hdfs/hdfs_test.sh
+          echo bash -x -e tests/test_pulsar/pulsar_test.sh
       - name: Test Linux
         run: |
           set -x -e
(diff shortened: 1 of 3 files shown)


yongtang:
The kafka test issue has been fixed. The issue was that GitHub Actions changes to ipv6 and `localhost` domain is having trouble to map to 127.0.0.1.

The pulsar is still having issues. I assume this might be related to ipv6 vs ipv4 as well.

/cc @kvignesh1420 @terrytangyuan @burgerkingeater

terrytangyuan (approved):

yongtang:
@BewareMyPower the test setup on pulsar is failing and is likely caused by ipv4 vs ipv6. Do you know if there is a way to only bind to ipv4 for pulsar?

BewareMyPower:
I'll take a look soon. BTW, could you try to configure `advertisedAddress=127.0.0.1` in `conf/standalone.conf` using `sed` command?

burgerkingeater (approved):

kvignesh1420 (approved):
@yongtang seems like the Kafka, Pulsar issues are resolved but some fs based issues are popping up randomly. However, based on the purpose of this PR, we can merge it.

yongtang:
@kvignesh1420 Thanks for the latest commit fix! As the master branch is failing I will merge this PR for now. 

The pulsar is ok on macOS though it still fails on Linux, likely due to ipv4/ipv6 (as was mentioned earlier). The PR disabled pulsar on Linux for now. I will open an issue to track pulsar on linux test.

PR #1371, opened 2021-04-19
Enable Pulsar test on Linux

yongtang:
This PR is to expose the pulsar test issue on Linux

/cc @BewareMyPower I tried  configure `advertisedAddress=127.0.0.1` in `conf/standalone.conf` (suggested in https://github.com/tensorflow/io/pull/1370#issuecomment-822151685) though the pulsar still cannot be setup on Github Actions CI. Maybe you can take a look if you get a chance?

### DIFF
--- a/tests/test_pulsar.py
+++ b/tests/test_pulsar.py
@@ -25,7 +25,7 @@ default_pulsar_timeout = 5000
 
 
 @pytest.mark.skipif(
-    sys.platform in ("win32", "linux"),
+    sys.platform in ("win32",),
     reason="TODO Pulsar not setup properly on Windows/Linux yet",
 )
 def test_pulsar_simple_messages():
@@ -47,7 +47,7 @@ def test_pulsar_simple_messages():
 
 
 @pytest.mark.skipif(
-    sys.platform in ("win32", "linux"),
+    sys.platform in ("win32",),
     reason="TODO Pulsar not setup properly on Windows/Linux yet",
 )
 def test_pulsar_keyed_messages():
@@ -73,7 +73,7 @@ def test_pulsar_keyed_messages():
 
 
 @pytest.mark.skipif(
-    sys.platform in ("win32", "linux"),
+    sys.platform in ("win32",),
     reason="TODO Pulsar not setup properly on Windows/Linux yet",
 )
 def test_pulsar_resubscribe():
@@ -120,7 +120,7 @@ def test_pulsar_resubscribe():
 
 
 @pytest.mark.skipif(
-    sys.platform in ("win32", "linux"),
+    sys.platform in ("win32",),
     reason="TODO Pulsar not setup properly on Windows/Linux yet",
 )
 def test_pulsar_invalid_arguments():
@@ -178,7 +178,7 @@ def test_pulsar_invalid_arguments():
 
 
 @pytest.mark.skipif(
-    sys.platform in ("win32", "linux"),
+    sys.platform in ("win32",),
     reason="TODO Pulsar not setup properly on Windows/Linux yet",
 )
 def test_pulsar_write_simple_messages():
@@ -208,7 +208,7 @@ def test_pulsar_write_simple_messages():
 
 
 @pytest.mark.skipif(
-    sys.platform in ("win32", "linux"),
+    sys.platform in ("win32",),
     reason="TODO Pulsar not setup properly on Windows/Linux yet",
 )
 def test_pulsar_write_keyed_messages():
--- a/tests/test_pulsar.py
+++ b/tests/test_pulsar.py
@@ -25,8 +25,7 @@ default_pulsar_timeout = 5000
 
 
 @pytest.mark.skipif(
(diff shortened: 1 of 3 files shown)


BewareMyPower:
I have a questions about where did you configure `advertisedAddress=127.0.0.1`?

And I'm not sure how the `pulsar_test.sh` failed? Or where can I find the related logs?

Since `bin/pulsar-daemon start standalone` start a pulsar standalone in background, I think we can dump the logs for debugging when it failed.

For example, add 

```
cat logs/*.log
```

after https://github.com/yongtang/io/blob/b79ba92a714e3669d08a9fcf4b3aed99d2e66888/tests/test_pulsar/pulsar_test.sh#L44.

or somewhere else.

yongtang:
@BewareMyPower I have updated the PR and now the logs shows a little more information. I am seeing lots of:
```
23:04:27.241 [pulsar-client-io-1-1] WARN  org.apache.pulsar.client.impl.ConnectionPool - Failed to open connection to localhost:6650 : io.netty.channel.AbstractChannel$AnnotatedConnectException: finishConnect(..) failed: Connection refused: localhost/127.0.0.1:6650
```

Wondering if you have any idea where the issue might be?

BewareMyPower:
The log is from Pulsar client. The client tried to connect `localhost:6650` but failed.

I think we can add an option `--url pulsar://127.0.0.1:6650` to `bin/pulsar-client produce` commands like https://github.com/tensorflow/io/blob/946d0549499b9bf8faa78851a20487430f4c551c/tests/test_pulsar/pulsar_test.sh#L57

Also, it's equivalent to change the `conf/client.conf` from the default

```properties
brokerServiceUrl=pulsar://localhost:6650/
```

to

```properties
brokerServiceUrl=pulsar://127.0.0.1:6650/
```

kvignesh1420:
@yongtang a bump might be needed to re-trigger Github actions.

BewareMyPower:
It looks like the problem is that pulsar standalone still failed to bind 127.0.0.1 so that `curl http://127.0.0.1:8080/xxx` always failed.

After taking a look at the configs, maybe you should also set `bindAddress`.

```
bindAddress=127.0.0.1
advertisedAddress=127.0.0.1
```

Here's the default config file for pulsar standalone: https://github.com/apache/pulsar/blob/master/conf/standalone.conf

kvignesh1420:
thanks for the fix @yongtang @BewareMyPower!

yongtang:
Looks like the issue has been fixed. Thanks @BewareMyPower for the help.

/cc @kvignesh1420 @terrytangyuan @burgerkingeater

tests/test_pulsar.py
```diff
@@ -25,7 +25,7 @@
 
 
 @pytest.mark.skipif(
-    sys.platform in ("win32", "linux"),
+    sys.platform in ("win32",),
     reason="TODO Pulsar not setup properly on Windows/Linux yet",
```
kvignesh1420, line 29: @yongtang The reason be changed only to `win32` here. Also, can we try this fix on win as well?

tests/test_pulsar.py
```diff
@@ -73,7 +73,7 @@ def test_pulsar_keyed_messages():
...
```
kvignesh1420, line 77: same as above

tests/test_pulsar.py
```diff
@@ -120,7 +120,7 @@ def test_pulsar_resubscribe():
...
```
kvignesh1420, line 124: same as above

tests/test_pulsar.py
```diff
@@ -208,7 +208,7 @@ def test_pulsar_write_simple_messages():
...
```
kvignesh1420, line 212: same as above

kvignesh1420 (changes requested):
minor fixes to be made.

tests/test_pulsar.py
BewareMyPower: It's because Pulsar server cannot run on Windows platform yet, see http://pulsar.apache.org/docs/en/standalone/#system-requirements

kvignesh1420: I see. Thanks!

yongtang:
@kvignesh1420 Can you take a look and see if update is needed?

kvignesh1420:
@yongtang can we change the `pytest.skip` reason from `"TODO Pulsar not setup properly on Windows/Linux yet"` to `"TODO Pulsar not setup properly on Windows yet"` as you have already fixed the linux based issues? Only this change is required.

yongtang:
@kvignesh1420 Ah yes that needs to be changed. Let me update.

yongtang:
@kvignesh1420 The PR has been updated. Please take a look.

kvignesh1420 (approved):
@yongtang thanks for the fix.

kvignesh1420:
@yongtang seems like a lint fix is needed.

yongtang:
@kvignesh1420 Updated. Thanks.

PR #1374, opened 2021-04-24
[Try] Try to build orc 

yongtang:


vnvo2409:
@yongtang could we use this `libhdfspp` for our hdfs filesystem ( maybe an improvement for the future 🤔 ) ?

yongtang:
> @yongtang could we use this libhdfspp for our hdfs filesystem ( maybe an improvement for the future 🤔 ) ?

@vnvo2409 I think that should be possible. Our hdfs filesystem actually only need a hdfs.h to get the data type and struct information so it is fairly lightweight. I guess any version of hdfs header file will be enough.

PR #1376, opened 2021-04-26
Install gs plugin for api and benchmarking checks

kvignesh1420:
This PR handles the failures in API compatibility checks and benchmarking jobs due to the dependency on the new `tensorflow-io-plugin-gs`.

### DIFF
--- a/.github/workflows/api.yml
+++ b/.github/workflows/api.yml
@@ -41,6 +41,7 @@ jobs:
           rm -rf tensorflow_io
           echo ${{ matrix.version }} | awk -F: '{print $1}' | xargs python -m pip install -U
           echo ${{ matrix.version }} | awk -F: '{print $2}' | xargs python -m pip install --no-deps -U
+          if [[ "${{ matrix.version }}" == *"tensorflow-io-nightly"* ]]; then python -m pip install tensorflow-io-plugin-gs-nightly ; fi
           python -m pip install pytest-benchmark boto3 google-cloud-storage==1.32.0
           python -m pip freeze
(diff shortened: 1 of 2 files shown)


yongtang:
@kvignesh1420 I think `tensorflow-io-nightly` depends on `tensorflow-io-plugin-gs-nightly` so it will be installed directly. Though the gs plugin will not be invoked through `import tensorflow_io`, instead it will only be invoked through `import tensorflow_io_plugin_gs` (to allow tensorflow integration in later versions).

kvignesh1420:
@yongtang if that's the case, this PR can be closed. However for an unknown reason, the compatibility tests are failing for `https` fs plugin. The tests pass locally though.

yongtang:
@kvignesh1420 I will take a look.

yongtang:
@kvignesh1420 I see where the issue is: while `tensorflow-io-nightly` correctly depends on `tensorflow-io-plugin-gs-nightly`, the api and benchmark test actually uses `python -m pip install --no-deps -U` to install, which does not install dependency due to `--no-deps` (because it needs to adjust the tensorflow's dependency version). Because of that, the `tensorflow-io-plugin-gs-nightly` is not getting installed. I think we still need to install `tensorflow-io-plugin-gs-nightly` manually with the same version as `tensorflow-io-nightly`. Sorry for the back and forth.

kvignesh1420:
@yongtang no problem. I have re-opened the PR. PTAL.

yongtang (approved):
LGTM

PR #1380, opened 2021-04-27
bump `absl` to match `tf-2.5-rc1`

vnvo2409:
- Related commit: https://github.com/tensorflow/tensorflow/commit/404f96bfaff94185b65a0a19373022354cf78cd5
- In this `absl` version, there is a `cleanup` library that we could replace some parts of `gcs` code.


vnvo2409:
Hmm this commit does not get into `TF 2.5rc1` yet.

PR #1384, opened 2021-04-28
switch to `Aws::S3::Model::ListObjectsV2Request`

vnvo2409:
After going through this PR https://github.com/tensorflow/tensorflow/pull/11089, I find that we are using `list-objects-v1` for compatibility with 3rd-party `s3` implementations. The implementations mentioned in that PR were Eucalyptus Walrus, Scality S3 server, Minio ( https://github.com/tensorflow/tensorflow/pull/11089#issuecomment-315017768 ).

- [Eucalyptus Walrus](https://github.com/eucalyptus/eucalyptus) has stopped updating since 2017. There is one [fork](https://github.com/Corymbia/eucalyptus) that seems to be the main replacement and it is supporting `list-objects-v2`.
- Scality S3 server supports `list-objects-v2`.
- Minio supports `list-objects-v2`.

So I think we could eventually use `list-objects-v2` for our filesystem.

---

Also in that PR, I've noticed that `s3` filesystem was first compiled with `ENABLE_NO_ENCRYPTION` and we had to provied our own crypto functions through `s3_crypto`. But in https://github.com/aws/aws-sdk-cpp/commit/907dcdc, `aws-sdk-cpp` has added supports for `boringssl/openssl`. Therefore, I think `s3_crypto` is no longer needed. WDYT @yongtang ?

### DIFF
--- a/tensorflow_io/core/plugins/s3/s3_filesystem.cc
+++ b/tensorflow_io/core/plugins/s3/s3_filesystem.cc
@@ -28,7 +28,7 @@ limitations under the License.
 #include <aws/s3/model/GetObjectRequest.h>
 #include <aws/s3/model/HeadBucketRequest.h>
 #include <aws/s3/model/HeadObjectRequest.h>
-#include <aws/s3/model/ListObjectsRequest.h>
+#include <aws/s3/model/ListObjectsV2Request.h>
 #include <aws/s3/model/UploadPartCopyRequest.h>
 #include <stdlib.h>
 #include <string.h>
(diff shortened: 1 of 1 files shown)


yongtang:
> Therefore, I think s3_crypto is no longer needed. 

@vnvo2409 Yes I think s3_crypto is not needed anymore.

vnvo2409:
@yongtang Please take a look !

yongtang (approved):
LGTM, Thanks for the PR!

PR #1390, opened 2021-05-01
Add batch support for resample

### LINKED ISSUE #1366: Add batch support to audio resample
As this library is usually combined with Tensorflow, I think that the resample function should accept batches of audios in the form of tensors of shape [batch_size, samples, channels]. 

I know this can be easily solved by iterating over the batch dimension, but all the Tensorflow functions make this transparent to the user, that's why I was expecting the same here

yongtang:
This PR adds batch support for resample so that
it is possible to pass 3D input of `[batch, sample, channel]`
shape for tfio.audio.resample.

The implementation is done through `tf.vectorized_map`

This PR fixes #1366.


kvignesh1420 (approved):
LGTM, thanks @yongtang

PR #1391, opened 2021-05-03
improvements for `test_filesystem.py`

vnvo2409:
> I remember there might be an issue where az is only available through kernel ops, or through gfile types of APIs but not available to both at the same time.

It is accessible through kernel ops but it didn't work correctly so I disable `az` on `win32` for now. I thought that https://github.com/tensorflow/tensorflow/pull/43594 should solve it, could you check it again ?

398dc7e is an example for showing how flexible these tests could be.

@yongtang could you take a look ?

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -435,7 +435,6 @@ jobs:
           python -m pip install google-cloud-bigquery-storage==0.7.0 google-cloud-bigquery==1.22.0 fastavro
           (python -m pytest -s -v test_bigquery.py)
           (python -m pytest -s -v test_dicom.py)
-          (python -m pytest -s -v test_filesystem.py)
 
   release:
     name: Release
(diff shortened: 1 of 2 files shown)


vnvo2409:
@yongtang Tests on Windows throw `az` isn't implemented. Is it expected or not ?

yongtang:
@vnvo2409 I remember there might be an issue where az is only available through kernel ops, or through gfile types of APIs but not available to both at the same time. I was wondering if this was caused by different dll was loaded for those two route to find the file system registration? (e.g., if hit one route then another route with a separate dll is not available)?

vnvo2409:
Windows throws `https` isn't implemented as well so I disable all `test_filesystem` on Windows for now.

vnvo2409:
Macos are falling because of another reasons.

yongtang (approved):
LGTM. Thanks!

PR #1395, opened 2021-05-04
Convert to use tensorflow's third party BUILD files whenever possible

yongtang:
In tensorflow-io, while there are many third-party libraries such as
arrow/avro that was not used by tensorflow, quite a few libraries such
as boringssl/curl/nasm are used by tensorflow.

In the past we used to maintain our own BUILD files for third party libraries
such as curl. Which potentially causes duplication (with tensorflow), and may
cause version incompatibility if we are out of sync with tensorflow's third party
libraries.

This PR uses tensorflow's BUILD files whenever possible so that we can reduce
overlap.

Note several libraries needs special handling in tensorflow-io, e.g., snappy
in tensorflow-io needs inclusion of snappy-c. We can see if it is possible to
update upstream tensorflow instead.

Note: the core part of the PR is the following:
```
http_archive(
    name = "org_tensorflow",
    sha256 = "354e58a2883210575ae7deacc9ab4a7ce00a3c60315151c2efbb734255e73713",
    strip_prefix = "tensorflow-2.5.0-rc1",
    urls = [
        "https://github.com/tensorflow/tensorflow/archive/refs/tags/v2.5.0-rc1.tar.gz",
    ],
)

load("@org_tensorflow//tensorflow:workspace3.bzl", "tf_workspace3")
tf_workspace3()
load("@org_tensorflow//tensorflow:workspace2.bzl", "tf_workspace2")
tf_workspace2()
load("@org_tensorflow//tensorflow:workspace1.bzl", "tf_workspace1")
tf_workspace1()
load("@org_tensorflow//tensorflow:workspace0.bzl", "tf_workspace0")
tf_workspace0()
```


vnvo2409:
I think `gs` test will fail because of a breaking change introduced in `google-cloud-cpp` 1.19.0 related to the emulator. Could we manage our own `google-cloud-cpp` instead ?

yongtang:
@vnvo2409 We use our `com_github_googleapis_google_cloud_cpp` which is in:
https://github.com/tensorflow/io/blob/f53af01d7d88a5bbf070783f16617f432fc12070/WORKSPACE#L92-L108

yongtang:
Bump the tensorflow source package to TF 2.5.0. I think this PR is ready for review now. /cc @terrytangyuan @kvignesh1420  @burgerkingeater

yongtang:
@terrytangyuan @kvignesh1420 @burgerkingeater All tests pass now. Please take a look.

yongtang:
This PR has a net code reduction of 1300 lines:
```
+551 −1,870 
```

terrytangyuan (approved):
This looks much cleaner now!

PR #1396, opened 2021-05-04
add missing symbol to `gcs_filesystem`

vnvo2409:
It seems that I forgot to add some function pointers to `gcs_filesystem` 😕


yongtang (approved):
LGTM

PR #1405, opened 2021-05-07
use `pytest-xdist` to speed up `pytest`

vnvo2409:


vnvo2409:
It's caused test hanging, close

PR #1406, opened 2021-05-07
don't run tests on forks

vnvo2409:
I think we don't need to run these actions on forks which is a waste of computational resource

/cc @kvignesh1420 @yongtang 

For example, https://github.com/vnvo2409/io/actions

image

### DIFF
--- a/.github/workflows/api.yml
+++ b/.github/workflows/api.yml
@@ -13,6 +13,7 @@ env:
 jobs:
 
   macos:
+    if: github.repository == 'tensorflow/io' # Don't do this in forks
     name: macOS ${{ matrix.python }} + ${{ matrix.version }}
     runs-on: macos-latest
     strategy:
@@ -53,6 +54,7 @@ jobs:
           python -m pytest -s -v tests/test_filesystem.py
 
   linux:
+    if: github.repository == 'tensorflow/io' # Don't do this in forks
     name: Linux ${{ matrix.python }} + ${{ matrix.version }}
     runs-on: ubuntu-20.04
     strategy:
@@ -92,6 +94,7 @@ jobs:
           python -m pytest -s -v tests/test_filesystem.py
 
   windows:
+    if: github.repository == 'tensorflow/io' # Don't do this in forks
     name: Windows ${{ matrix.python }} + ${{ matrix.version }}
     runs-on: windows-latest
     strategy:
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -7,6 +7,7 @@ on:
 
 jobs:
   macos:
+    if: github.repository == 'tensorflow/io' # Don't do this in forks
     name: macOS ${{ matrix.python }} + ${{ matrix.version }}
     runs-on: macos-latest
     strategy:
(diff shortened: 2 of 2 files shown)


.github/workflows/build.yml
```diff
@@ -69,6 +73,8 @@ jobs:
           python3 -c 'import tensorflow as tf; print(tf.version.VERSION)'
 
   linux:
+    # Don't do this when pushing master on forks
+    if: github.repository == 'tensorflow/io' ||  github.event_name == 'pull_request'
```
vnvo2409, line 77: @terrytangyuan I exclude `api` and `benchmark` because I think generally they aren't something the contributors want to run. With the `build` actions, the contributors could still run the CI by creating a PR against their own `master`. I think it is better than running automatically on pusing `master`

terrytangyuan: I don't have a preference here. @yongtang WDYT?

yongtang: @vnvo2409 @terrytangyuan Another alternative is to enable build 1) if code is on `tensorflow/io`, or 2) if code is on forked repo and not `master` branch?

The reason is, normally people will create a `non-master` branch on forked repo to prepare the pull request. And some (like me) may want to run the build on forked repo before submitting the PR.

Currently if I submit a PR, I comment out:
https://github.com/tensorflow/io/blob/master/.github/workflows/build.yml#L5-L6
so that a CI build can be triggered on my forked repo. If build pass, I will add the lines back and submit the PR.

Since I always keep master branch up to date with `tensorflow/io`, I don't need to trigger CI build on master branch of the forked repo.

vnvo2409: > The reason is, normally people will create a non-master branch on forked repo to prepare the pull request. 

Yeah me too!

> Currently if I submit a PR, I comment out:
https://github.com/tensorflow/io/blob/master/.github/workflows/build.yml#L5-L6
so that a CI build can be triggered on my forked repo. If build pass, I will add the lines back and submit the PR.

I think the problem will arise after submiting the PR, because we usually still need to push more commits after opening PR, and at that time, there will be 2 CIs running concurrently.

> Since I always keep master branch up to date with `tensorflow/io`, I don't need to trigger CI build on master branch of the forked repo.

That's the problem I am trying to solve.

kvignesh1420: The takeaway is, some fraction of contributors have to comment/uncomment code to adjust to the current flow of PR/fork builds. The approach should be to minimize the number of people making such changes on a regular basis.

If we think about first-time contributors, they might not care about these actions that might be running on their forked repositories. However, regular contributors might want to run these jobs on forked repositories before opening the PR (as per @yongtang). So @vnvo2409  as long as forked actions are NOT AFFECTING the performance of `tensorflow/io` actions, I think the workflows can be left as they are. However, If that's not the case, then we can go ahead with this PR.

kvignesh1420: Maybe we can stop forked builds for `api compatibility` and `benchmarks` as of now.

vnvo2409: > Maybe we can stop forked builds for api compatibility and benchmarks as of now.

Done

kvignesh1420 (approved):

PR #1408, opened 2021-05-08
Revert "switch to `Aws::S3::Model::ListObjectsV2Request` (#1384)"

yongtang:
We have been seeing CI failures where segmentation faults are being thrown out randomly. Several attempts have been made to resolve the issue though as the seg fault does not happen all the time, it has been hard to track down to which commit was the issue.

I have been trying to bisect the past commits, and run GitHub Actions multiple times for the past couple of days on my repo. (See branch build with name `fix`|`fix2`...`fix6`|`fix8`):
https://github.com/yongtang/io/actions

The issue points to d08b554f1d4ab5cebd93829b5431383670774c75 (or in combination of 63259e7b5ee7e3f935a7ad6852e7ab8897271821).

Not exactly sure why this is causing the issue. Though I think we want to revert the commit first as TF 2.5.0 (and our tensorflow-io 0.18.0 release is imminent).

In order to avoid false negative, I will trigger CI build multiple times to make sure this is the true fix.

This PR reverts the commit.
/cc @vnvo2409 FYI

This reverts commit d08b554f1d4ab5cebd93829b5431383670774c75.

Signed-off-by: Yong Tang <yong.tang.github@outlook.com>

PR #1389, opened 2021-05-01
Add ffmpeg support on mac

kyamagu:
This PR adds ffmpeg support for mac.

Related: #1064

PR #1410, opened 2021-05-09
Use xdist with multi-processing

yongtang:
This PR is an alternative to #1408. The idea is to use xdist with multiprocessing so that at least all tests can complete (even if test faced with segmentation fault).

Signed-off-by: Yong Tang yong.tang.github@outlook.com

### DIFF
--- a/.github/workflows/build.wheel.sh
+++ b/.github/workflows/build.wheel.sh
@@ -9,8 +9,7 @@ run_test() {
   (cd wheelhouse && $entry -m pip install tensorflow_io_plugin_gs-*-cp${CPYTHON_VERSION}-*.whl)
   (cd wheelhouse && $entry -m pip install tensorflow_io-*-cp${CPYTHON_VERSION}-*.whl)
   $entry -m pip install -q pytest pytest-benchmark pytest-xdist boto3 fastavro avro-python3 scikit-image pandas pyarrow==3.0.0 google-cloud-pubsub==2.1.0 google-cloud-bigtable==1.6.0 google-cloud-bigquery-storage==1.1.0 google-cloud-bigquery==2.3.1 google-cloud-storage==1.32.0 PyYAML==5.3.1 azure-storage-blob==12.8.1
-  (cd tests && $entry -m pytest --benchmark-disable -v --import-mode=append --forked --numprocesses=auto $(find . -type f \( -iname "test_*_v1.py" \)))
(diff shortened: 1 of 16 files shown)


vnvo2409:
@yongtang FYI, when I tried this one, one test hung 6 hours: https://github.com/tensorflow/io/runs/2526571529

yongtang:
The PR is ready for review now. Changes in this PR:
1. Converted to `pytest` + `xdist` + multiprocess(`--forked`) so that even if one test crashes, the rest of the tests will at least finish.
2. Disabled ffmpeg/aac related tests on macOS. It seems they caused lots of crash on GitHub Actions. Note only tests are disabled. The rest of the changes in PR #1389 are still very much in place. We will add the test back gradually. (cc @kyamagu )
3. The `test_filesystem.py` is disabled temporarily. The reason of the issue seems to be related to the reloading of tensorflow or shared libraries (see addition comment below). We will add the tests back gradually (cc @vnvo2409 )

In addition to the above, there are several notes about pytest+xdist+multiprocessing:
1. requires some logic to be part of the test (not at the beginning of the module/file). See https://stackoverflow.com/questions/48234032/run-py-test-test-in-different-process for one example where in answers section, eager mode/non-eager mode switch needs to be within tests. If there are included at the beginning of the file then there might be some unintended effects.
   For that reason, we made some changes in some of the tests (e.g., test_arrow.py in this PR).
   The basic rule is that, at the beginning of each file, we only want to do `import tensorflow as tf` and `import tensorflow_io as tfio`. All other additional imports or changes related to tfio or tf will be within the scope of the individual tests. cc @BryanCutler 
2. Some of our tests (e.g., in kafka) may relies `not` running concurrently. For that we use `--dist loadfile` option in pytest+xdist to make sure if tests are within the same file, they will be grouped and sent to the same worker (thus guarantees that they will not run in parallel). In the future we may consider remove this limitation with some modifications.

Finally, since we run each tests in an individual process, there is no need to split out bigquery tests from the rest of the tests (due to grpc issue). The PR made the change and treat all tests as the same now. This is the added benefit of pytest+xdist+forked cc @vlasenkoalexey 

cc @terrytangyuan @kvignesh1420 @burgerkingeater

yongtang:
Ping  @terrytangyuan @kvignesh1420 @burgerkingeater to take a look.

kvignesh1420 (approved):
@yongtang I triggered the actions once again and all of them seem to pass. LGTM as of now.

Issue #1392, opened 2021-05-03
Implementation proposal for audio resample


AlexFuster:
At Voicemod, we have made a tensorflow version of the audio resampling function from torchaudio [https://pytorch.org/audio/stable/transforms.html?highlight=resample#torchaudio.transforms.Resample](url).

I think It would be a good alternative/complement for the current implementation of audio resampling of this repo [https://www.tensorflow.org/io/api_docs/python/tfio/audio/resample](url) for 4 reasons.

1. Supports batches: I know that this was recently added to the current implementation)
2. Supports gradients: Unlike the current implementation. This feature is very useful if you want to resample inside a training graph
3. The results are as good as the ones given by the current implementation and they are close to resampy
4. For some reason, the current implementation always adds a misalignment of size 32 or so, as one can see in the image below
![image](https://user-images.githubusercontent.com/32075911/116932288-594e2c80-ac62-11eb-85ea-7e8a730d8919.png)

Tell me if you are interested

yongtang:
@AlexFuster The resample was implemented wit speex in the current repo. If an alternative implementation brings additional advantage (like matching resampy/etc) then we certainly welcome PRs.

AlexFuster:
So, do I replace the resample function in ```audio_ops.py``` or do I create a separate function?

yongtang:
@AlexFuster You can replace the resample function in audio_ops.py and replace C++ kernel in https://github.com/tensorflow/io/blob/master/tensorflow_io/core/kernels/audio_kernels.cc#L220-L299

AlexFuster:
Sorry, I forgot to add an import to the first one

PR #1411, opened 2021-05-10
replaced resample function by new implementation

AlexFuster:
#1392


AlexFuster:
Sorry, missed something

PR #1412, opened 2021-05-10
Replaced resample function by new implementation (2)

AlexFuster:
#1392

### DIFF
--- a/tensorflow_io/core/python/ops/audio_ops.py
+++ b/tensorflow_io/core/python/ops/audio_ops.py
@@ -17,7 +17,7 @@
 import sys
 
 import tensorflow as tf
-
+import math
 from tensorflow_io.core.python.ops import core_ops
 
 
--- a/tensorflow_io/core/python/ops/audio_ops.py
+++ b/tensorflow_io/core/python/ops/audio_ops.py
@@ -418,10 +418,10 @@ def _get_sinc_resample_kernel(rate_in, rate_out, lowpass_filter_width):
     return tf.expand_dims(kernels,axis=1)*scale, width
 
 def resample(input, rate_in, rate_out, lowpass_filter_width= 6):
-    """Resamples the waveform at the new frequency. This matches Kaldi’s OfflineFeatureTpl ResampleWaveform which uses a LinearResample (resample a signal at linearly spaced intervals to upsample/downsample a signal). LinearResample (LR) means that the output signal is at linearly spaced intervals (i.e the output signal has a frequency of rate_out). It uses sinc/bandlimited interpolation to upsample/downsample the signal.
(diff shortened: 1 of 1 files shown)


yongtang:
@AlexFuster if you have black installed,  can you just run black audio_ops.py and see if the file is formatted automatically?

AlexFuster:
The tests seem to be taking too long

AlexFuster:
ok, here we go again

AlexFuster:
Oh yeah, I forgot to mention It does just support floating point audios. Is this a problem?

yongtang:
> Oh yeah, I forgot to mention It does just support floating point audios. Is this a problem?

@AlexFuster The existing implementation support both int16 and float32. Though I think it will not be too difficult to add int16 support by casting to float32 and back.

yongtang:
@AlexFuster The test failures are legitimate. The reason being
1. `Tensor` in tensorflow may not be treated as python native value (e.g, `1`, `1.1`, etc). Things like `int(x)` may not work when `x` is a tensor. Instead, an explicit op of `tf.cast(x, tf.int32)` is needed.
2. A `tensor`'s shape may not be known statically before hand. So `x.shape` could just return `None` in case shape is not known yet.
3. When `x` is a tensor, the `if...else...` logic does not work as tensor's value is not statically available (before run). Instead a TF specific op of `tf.case` or `tf.cond` has to be used (you might notice the original implementation consists of `tf.case` to replace `if...elif...else`.

You may want to take a look at the test failures and see if some change is needed. I will also take a look.

AlexFuster:
I don't think the error comes from any of those 3 problems. It is just `tf.nn.conv1d` not accepting `int16` values as inputs. I'll try adding int16 support by casting and casting back as you said

yongtang:
With `int16` tests may pass, though in the original implementation the resample actually support passing `rate_in`, `rate_out` as tensor and input with unknown shape. I think with new implementation this ability is removed as I could see `gcd` and `int(rate_in)` will cause the failure if a tensor with unknown rank is passed for rate_in. The input may also cause some issue if it is a tensor with unknown shape. We can get the tests pass and address the rest in follow up PRs though.

AlexFuster:
Why would you want a tensor with a shape other than () being passed as `rate_in`?

yongtang:
> Why would you want a tensor with a shape other than () being passed as rate_in?
@AlexFuster Sorry it should be  `a tensor without constant value`. In graph mode (such as within tf.data.Dataset pipeline), user may pass `input`, `rate_in`, `rate_out` as tensor without const value. In that case, the graph mode will fail. They can be avoided by using ops that will work for tensor. For example, `int(rate_in)` can be converted to use `tf.cast(rate_in, tf.int32)`, `math.gcd` can be replaced with ops as `tf.experimental.numpy.gcd` I think.

AlexFuster:
There's a problem `tf.nn.conv1d` does not accept a tensor as a value for stride. So the operation `resampled = tf.nn.conv1d(waveform, kernel, stride=rate_in, padding="VALID")` is nor working for me when `rate_in` is a tensor. It does if I use `rate_in.numpy()`, but that would not be compatible with graph mode

AlexFuster:
@googlebot I fixed it.

rasul247 (approved):

PR #1414, opened 2021-05-11
fix api compatibility check failures

kvignesh1420:
This is a follow-up of https://github.com/tensorflow/io/pull/1413 as the API compatibility checks are failing and the fix is now applied.

### DIFF
--- a/.github/workflows/api.yml
+++ b/.github/workflows/api.yml
@@ -4,6 +4,9 @@ on:
   push:
     branches:
       - master
+  pull_request:
+    branches:
+      - master
   schedule:
     - cron: "0 12 * * *"
 
--- a/.github/workflows/api.yml
+++ b/.github/workflows/api.yml
@@ -23,7 +23,7 @@ jobs:
       fail-fast: false
       matrix:
         python: ['3.8']
-        version: ['tensorflow==2.5.0rc3:tensorflow-io-nightly', 'tf-nightly:tensorflow-io', 'tf-nightly:tensorflow-io-nightly']
+        version: ['tensorflow:tensorflow-io-nightly', 'tf-nightly:tensorflow-io', 'tf-nightly:tensorflow-io-nightly']
     steps:
       - uses: actions/checkout@v2
       - uses: actions/setup-python@v1
@@ -66,7 +66,7 @@ jobs:
       fail-fast: false
       matrix:
         python: ['3.8']
-        version: ['tensorflow==2.5.0rc3:tensorflow-io-nightly', 'tf-nightly:tensorflow-io', 'tf-nightly:tensorflow-io-nightly']
+        version: ['tensorflow:tensorflow-io-nightly', 'tf-nightly:tensorflow-io', 'tf-nightly:tensorflow-io-nightly']
     steps:
       - uses: actions/checkout@v2
       - uses: actions/setup-python@v1
@@ -108,7 +108,7 @@ jobs:
       fail-fast: false
       matrix:
         python: ['3.8']
-        version: ['tensorflow==2.5.0rc3:tensorflow-io-nightly', 'tf-nightly:tensorflow-io', 'tf-nightly:tensorflow-io-nightly']
+        version: ['tensorflow:tensorflow-io-nightly', 'tf-nightly:tensorflow-io', 'tf-nightly:tensorflow-io-nightly']
     steps:
       - uses: actions/checkout@v2
       - uses: actions/setup-python@v1
--- a/.github/workflows/api.yml
+++ b/.github/workflows/api.yml
@@ -43,7 +43,7 @@ jobs:
           python --version
           df -h
           rm -rf tensorflow_io
-          rm -rf tensorflow_io_plugin_gs
+          rm -rf tensorflow_io_gcs_filesystem
           export TF_USE_MODULAR_FILESYSTEM=1
           echo ${{ matrix.version }} | awk -F: '{print $1}' | xargs python -m pip install -U
           echo ${{ matrix.version }} | awk -F: '{print $2}' | xargs python -m pip install --no-deps -U
@@ -85,7 +85,7 @@ jobs:
           python --version
           df -h
           rm -rf tensorflow_io
-          rm -rf tensorflow_io_plugin_gs
+          rm -rf tensorflow_io_gcs_filesystem
           export TF_USE_MODULAR_FILESYSTEM=1
           echo ${{ matrix.version }} | awk -F: '{print $1}' | xargs python -m pip install -U
           echo ${{ matrix.version }} | awk -F: '{print $2}' | xargs python -m pip install --no-deps -U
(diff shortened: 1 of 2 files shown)


kvignesh1420:
@yongtang the gcs tests are failing with:
```
All attempts to get a Google authentication bearer token failed, returning an empty token.
Retrieving token from files failed with "Not found: Could not locate the credentials file.".
Retrieving token from GCE failed with:
"Failed precondition: Error executing an HTTP request: libcurl code 6 meaning 'Couldn't resolve host name', error details:
Could not resolve host: metadata".

tensorflow.python.framework.errors_impl.NotFoundError: Error executing an HTTP request:
HTTP response code 404 with body
'<?xml version='1.0' encoding='UTF-8'?><Error><Code>NoSuchBucket</Code><Message>
The specified bucket does not exist.</Message></Error>'
```

did we encounter this before?

kvignesh1420:
The tests are now failing in https://github.com/tensorflow/io/pull/1414/checks?check_run_id=2558044056 due to
```
tensorflow.python.framework.errors_impl.UnimplementedError: File system scheme 'https' not implemented (file: 'https://www.apache.org/licenses/LICENSE-2.0.txt')
```
and in https://github.com/tensorflow/io/pull/1414/checks?check_run_id=2558044231 due to:
```
tensorflow.python.framework.errors_impl.UnimplementedError: File system scheme 'az' not implemented (file: 'az://devstoreaccount1/aztest')
```

vnvo2409:
@kvignesh1420 I think it's because `tf-nightly` is built with [`absl lts_2021_03_24 LTS`](https://github.com/tensorflow/tensorflow/tree/master/third_party/absl)

vnvo2409:
@kvignesh1420 I think it won't be that easy. Since `TF` team doesn't seem to upgrade `absl` in the branch `r2.5`, we will have to choose one: either compatibility with `master` (`tf-nightly`) or `r2.5` (release). I think we have to raise an issue in TF core. @yongtang WDYT ?

kvignesh1420:
Since `tf:nightly` and `tf:2.5.0rc3` checks are mutually exclusive as of now, I prefer to be consistent with `tf:2.5.0rc3` as this will help in `tfio:0.18.0` release.

kvignesh1420:
cc: @yongtang

kvignesh1420:
cc: @yongtang can you PTAL.

yongtang:
@kvignesh1420 @vnvo2409 I tends to think that the issue might be because there is a dependency of `libtensorflow_io_plugins.so` and `libtensorflow_framework.so`. Because of that dependency, some symbols that should have been built from static third-party libraries within tensorflow-io has been instead linked to symbols within `libtensorflow_framework.so`.

My understanding is that, we only need C API (e.g., `TF_Log`) from `libtensorflow_framework.so` and anything else like `absl` etc should comes from our third-party version.

Maybe there is a way for us to not link `libtensorflow_framework.so` (except C API like `TF_Log`)?

vnvo2409:
> Maybe there is a way for us to not link `libtensorflow_framework.so` (except C API like `TF_Log`) ?

@yongtang I tried to use `llvm-strip` but I am not sure how to integrate it with `bazel`. Could you show me how to do it properly ?

yongtang:
@kvignesh1420 With PR #1423 I think some of the incompatibility issue may have been solved. Maybe we can give it a try again? We can keep `api.yml ` change and leave out `conftest.py` to see if some tests can pass.

Note currently we are not generating nightly build to pypi.org due to an unrelated windows bash issue (see PR #1427). Once PR #1427 is merged and a new nightly build is available on pypi.org, we can re-trigger the test on this PR.

kvignesh1420:
@yongtang if the current conftest.py is removed(commented in this case), then the tests are failing due to exit code 5.
https://github.com/tensorflow/io/pull/1414/checks?check_run_id=2600111979. It is a `pytest` failure where no tests were collected. Reference: https://github.com/pytest-dev/pytest/pull/2320/files

kvignesh1420:
The current api compatiblity checks status are as shown below. 
- The discrepancy between tf master and tf nightly is causing the failures for linux/mac os (`tf-nightly:tensorflow-io`). 
- Windows based issues still persists from our end even after https://github.com/tensorflow/io/pull/1427 was merged.

yongtang:
Thanks @kvignesh1420. It looks like once we have another tensorflow-io release (0.19.0) then at least the API compatibility tests on macOS and Linux will all pass.

On windows the situation is a little tricky. If I remember the issue was that tensorflow on Windows was compiled with --config=monolithic and that caused multiple instances of `Env::Default` in different places (on macOS/Linux there is only one instance of `Env::Default` in `libtensorflow_framework.so`.

I am currently inspecting the monolithic build of tensorflow (don't have Windows machine so I am building on Linux), and see if it is at least possible to remove the duplication of `Env::Default`.

/cc @vnvo2409 I am wondering, if we built tensorflow with `monolithic` config but combines `_pywrap_tensorflow_internal.so` and `_pywrap_file_io.so`, will that make a difference?

As far as I can see there are two usages of IO in tensorflow python. One is in kernel C++ ops (which is either in  `libtensorflow_framework.so` or `_pywrap_tensorflow_internal.so` depending on `monolithic=yes|no config`). Another is in `_pywrap_file_io.so` for gfile related operations. I think if we combine those two together in monolithic build then maybe we will end up with just one instance of `Env::Default`?

vnvo2409:
@yongtang 

> if we built tensorflow with monolithic config but combines `_pywrap_tensorflow_internal.so` and `_pywrap_file_io.so`, will that make a difference?

I haven't tried it yet but I think it could make a difference. However, there might be more then 2 instances of `Env::Default` You could see my comment here https://github.com/tensorflow/tensorflow/issues/19297#issuecomment-602224997. IIRC, some of them are created and used in compile-time only so it might not be a problem.

kvignesh1420:
Thanks for the info @yongtang @vnvo2409. I have removed the pr trigger for API compatibility checks. I think we can proceed with the working checks as of now then. PTAL.

yongtang (approved):

yongtang:
> However, there might be more then 2 instances of Env::Default

@vnvo2409  I built a local version with `monolithic` config, but only found two instances of exposed `Env::Default`:
- tensorflow/python/_pywrap_tensorflow_internal.so
- tensorflow/compiler/tf2xla/ops/_xla_ops.so

If these are the only two copies then maybe there is a way to get the issue fixed.

I build tensorflow on Linux machines though so don't know if it is the same on Windows.

vnvo2409:
@yongtang could you add a `std::cout` like i did to see how many time the `Env::Register` are called. Thank you !

yongtang:
@vnvo2409 I did look into the Register and notices that `gcs` has only been registered once but `file` and `ram` are registered multiple times:

On a side note, the multiple registration of `file` and `ram`  are not an issue. The `gcs` (only registered once) is actually an issue as even on tensorflow the `gcs` behavior is exactly we face with `http`: Kernel C++ ops work but `tf.gfile.GFile` does not: https://github.com/yongtang/io/runs/2621275868?check_suite_focus=true

I tends to believe the issue is with the duplication of `FileSystemRegistryImpl` though haven't been able to verify it.

PR #1423, opened 2021-05-15
Create stub libtensorflow_framework.so.2

yongtang:
This PR tries to resolve the issue discussed in https://github.com/tensorflow/io/pull/1414#issuecomment-841365328

On macOS and Linux, this PR creates stub `libtensorflow_framework.so.2`/`libtensorflow_framework.2.dylib` to avoid symbol mismatch.

TODO: Need to find a way to address issue on Windows as well (using .def to create stub .lib?).


yongtang:
@vnvo2409 @kvignesh1420  FYI

yongtang:
@burgerkingeater @terrytangyuan @kvignesh1420 @vnvo2409 This PR is now ready to review and it looks like everything works.

Basically, here is the summary of the changes:
1. For file system plugins, we create a stub libtensorflow_framework.so and link our code against it. If there is any unresolved symbols, during the run an error will be thrown. This also means we shouldn't encounter any issues for file system compatibility test in the future.
2. For C++ kernel ops, since we use C++ API from tensorflow, there is not a reliable way to guarantee API compatibility. However,
   a. macOS/Linux: We depends on stub libtensorflow_framework.so as well. linux/macOS so/dylib load will report an error if symbols cannot be resolved. (the behavior is the same whether we use stub one or installed one).
   b. Windows: We still depends on installed `_pywrap_tensorflow_internal.dll`. (While technically we may try `/FORCE:UNRESOLVED` option in visual studio, I am not sure the behavior for error handling is the same as before. We could explore `/FORCE:UNRESOLVED` later.

Some potential future enhancement (follow up PRs):
1. On macOS/Linux, with #1395 we may not even need installation of tensorflow anymore as we may use header files from tensorflow source directly (and we have stub .so/.dylib of our own).
2. On macOS/Linux, we can start looking into passing `visibility=hidden` in copt to future hide any local symbols.
3. On Windows, maybe we can created a C++ API list we use, and, create def file so that we don't need `_pywrap_tensorflow_internal.dll`?

yongtang:
All tests pass now.

terrytangyuan (approved):
LGTM

PR #1424, opened 2021-05-15
Export TF_InitPlugin API in windows DLL

yongtang:
Before this PR, our windows build fails with http file system but we didn't notice due to two reasons: 1. TF_InitPlugin was not exported in DLL. 2. bat error level was not populated to stop the CI build.

This PR does two things:

- This PR fixes the issue where libtensorflow_io_plugin.so does not export TF_InitPlugin in windows DLL which caused http file system to fail on windows. Now function is exported (see `__declspec(dllexport)`).
- This PR populates the error level of windows dos bat cmd so that the failure can be properly populated. (See `if %errorlevel% neq 0 exit /b %errorlevel%`)

Note: We still have some other additional failures with http which might be caused by duplication of two file system registration in tensorflow's dll. But this issue will need additional fix in follow up PRs (may need upstream tensorflow fix).

### DIFF
--- a/tensorflow_io/core/plugins/file_system_plugins.cc
+++ b/tensorflow_io/core/plugins/file_system_plugins.cc
@@ -17,7 +17,16 @@ limitations under the License.
 
 #include "absl/strings/ascii.h"
 
-void TF_InitPlugin(TF_FilesystemPluginInfo* info) {
+#if defined(_MSC_VER)
+#define TFIO_PLUGIN_EXPORT __declspec(dllexport)
+#else
+#define TFIO_PLUGIN_EXPORT __attribute__((visibility("default")))
+#endif
+
+// Please see:
+// tensorflow/tensorflow/c/experimental/filesystem/filesystem_interface.h
+// for definition of `TF_InitPlugin`
+TFIO_PLUGIN_EXPORT void TF_InitPlugin(TF_FilesystemPluginInfo* info) {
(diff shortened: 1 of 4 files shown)


tensorflow_io/core/plugins/file_system_plugins.cc
```diff
@@ -17,7 +17,16 @@ limitations under the License.
 
 #include "absl/strings/ascii.h"
 
-void TF_InitPlugin(TF_FilesystemPluginInfo* info) {
+#if defined(_MSC_VER)
+#define TFIO_PLUGIN_EXPORT __declspec(dllexport)
+#else
+#define TFIO_PLUGIN_EXPORT __attribute__((visibility("default")))
+#endif
```
vnvo2409, line 24: @yongtang IIRC, `tensorflow` already has this in `tensorflow/tensorflow/c/c_api_macros.h`. Could you try this out ? Thank you for the PR !

yongtang:
@vnvo2409 It is hard to use `tensorflow/tensorflow/c/c_api.h` directly for export dll. The reason being that the dllexport is controlled by both `TF_COMPILE_LIBRARY` and `TF_CAPI_EXPORT`. However, notice the C API we use (`TF_Log`/etc) also used these two macros. We want to use the C API from tenorflow as `import` but we want our API to be `export`. Which will conflict I think.

For that reason I added our own define.

kvignesh1420 (approved):
LGTM.

PR #1425, opened 2021-05-16
Use stub .so/.dylib for FFmpeg linkage and re-enable macOS FFmpeg test

yongtang:
This PR take a similar approaches as we have done for libtensorflow_framework.so linkage:
1. Create stub ffmpeg dependencies (e.g., libavformat.so) without content
2. Change sonata to with version (e.g., libavformat.so.57)
3. Directly depends `libtensorflow_io_ffmpeg_4.2.so`/etc with ffmpeg dependencies (e.g., libavformat.so)

In this way, system will still search for  `libavformat.so.57` which was specified in soname.

This allows us to not globally load all symbols from `libavformat.so.57` (which could be dangerous), but rely on system to resolve the symbols for us.

The end result is the reduction in code complexity (much of the original ffmpeg_ops.py code has been removed).

In addition, this PR also re-enabled all ffmpeg tests on macOS (disabled when we faced some issues earlier).

/cc @kyamagu FYI

### DIFF
--- a/.github/workflows/build.wheel.sh
+++ b/.github/workflows/build.wheel.sh
@@ -44,7 +44,6 @@ if [[ $(uname) == "Linux" ]]; then
   export CLASSPATH=$(${HADOOP_HOME}/bin/hadoop classpath --glob)
   export
 elif [[ $(uname) == "Darwin" ]]; then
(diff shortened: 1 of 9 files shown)


kvignesh1420 (approved):
LGTM.

PR #1428, opened 2021-05-17
Apply visibility=hidden further

yongtang:
This PR updates third-party libraries as much as possible
by passing macros or patch visibility defines so that
most of the third-party libraries can be compiled with symbols
hidden. This will significally reduce the chance of symbol collision.

Note `boost` is the only one left and it still exposes lots of symbols.


kvignesh1420 (approved):

hrw:
Reverting libwebp part fixes #1510

PR #1429, opened 2021-05-17
fix benchmark checks as per latest tfio release

kvignesh1420:
This PR fixes the failing benchmark checks after the 0.18.0 release of tfio.


kvignesh1420:
The PR triggered run is available here https://github.com/tensorflow/io/runs/2603529795

yongtang (approved):
LGTM, Thanks!

PR #1430, opened 2021-05-18
run `test_filesystem` without `pytest_xdist`

vnvo2409:
This PR enables `test_filesystem` without `pytest_xdist`, which I think that makes sense because:

- `pytest_xdist` doesn't work well with module-level fixtures. These fixtures are still called by each test which causes some errors like `BucketAlreadyExists`.
- `pytest_xdist` doesn't work well with `monkeypatch` and parameterized tests. Meanwhile, these tests use a lot of `monkeypatch` mechanism to control the envs and many behaviors, as well as parameterized tests.

After resolving many symbol collisions thank to @yongtang , I was able to run 4 CIs without `segfault`: https://github.com/vnvo2409/io/actions?query=branch%3Aenable-fs-test


yongtang:
@vnvo2409 Can you rename the test file to `test_standalone_filesystem.py`? If we want to have special cases that run outside of xdist, it might make sense to have prefix `test_standalone_xxx.py` so that we can capture similar standalone runs in the future. `.github/workflows/build.wheel.sh` can also be modified to capture all files with `test_standalone_xxx.py` and run individually.

vnvo2409:
@yongtang we might have a more elegant solution, that's putting this test and any test that needed to be run without `xdist` and use a specific `conftest.py` for that directory to disable `xdist`, what do you think ?

yongtang:
> putting this test and any test that needed to be run without xdist and use a specific conftest.py for that directory to disable xdist, 

@vnvo2409 Since we already have a `conftest.py` this make sense I think.

vnvo2409:
@yongtang Renaming is simpler so I go with it. Could you take a look ? Thank you !

yongtang (approved):

PR #1433, opened 2021-05-19
fix recent changes to filesystem test target

kvignesh1420:
This PR renames the `test_filesystem` pytest target to `test_standalone_filesystem` to fix API compatibility check failures.

Checks ref: https://github.com/tensorflow/io/runs/2623622764
root cause PR ref: https://github.com/tensorflow/io/pull/1430

cc: @vnvo2409  @yongtang


vnvo2409:
Thank you ! I missed that part.

yongtang (approved):
LGTM. Thanks!

PR #1436, opened 2021-05-21
Move options to `bazelrc` and use job container for linux

vnvo2409:
This PR changes 2 things:
- Move build options to `.bazelrc` instead of `build.yml`. This way, we don't need to repeat `BAZEL_OPTIMIZATION: --copt=-msse4.2 --copt=-mavx --compilation_mode=opt` and it will be eaiser to add another build configs like `asan`, `tsan` and `msan`.
- Use [`jobs.container`](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idcontainer) for linux build, so we don't have to use `docker` directly and we could eventually share a common script ( see `macos-bazel` and `linux-bazel` ) for both Linux and MacOS build. 

Testing script is harder since we have many things to setup. I am trying to compress everything into a docker container so we could replace this
https://github.com/tensorflow/io/blob/87e2b4d7191ff8569e7a18d6109839b8d2f3156e/.github/workflows/build.yml#L311-L321

with something like `docker-compose up` and contributors ( including me ) will have an easier way to setup the environment for testing locally. WDYT @yongtang ?

### DIFF
--- a/tools/build/configure.py
+++ b/tools/build/configure.py
@@ -114,9 +114,9 @@ def write_config():
                     bazel_rc.write('build --action_env TF_CUDA_VERSION="10.1"\n')
                     bazel_rc.write('build --action_env TF_CUDNN_VERSION="7"\n')
             # Needed for tf rules
-            bazel_rc.write('build --experimental_repo_remote_exec\n')
+            bazel_rc.write("build --experimental_repo_remote_exec\n")
             # Enable platform specific config
-            bazel_rc.write('build --enable_platform_specific_config\n')
+            bazel_rc.write("build --enable_platform_specific_config\n")
             # Needed for GRPC build
             bazel_rc.write('build:macos --copt="-DGRPC_BAZEL_BUILD"\n')
             # Stay with 10.14 for macOS
@@ -124,6 +124,27 @@ def write_config():
             bazel_rc.write('build:macos --linkopt="-mmacosx-version-min=10.14"\n')
             # MSVC (Windows): Standards-conformant preprocessor mode
             bazel_rc.write('build:windows --copt="/Zc:preprocessor"\n')
+            # Config for CI and release build
+            bazel_rc.write("build:ci --copt=-msse4.2\n")
+            bazel_rc.write("build:ci --copt=-mavx\n")
+            bazel_rc.write("build:ci --compilation_mode=opt\n")
+            bazel_rc.write("build:mac_ci --copt=-Wunguarded-availability\n")
+            bazel_rc.write(
+                "build:linux_ci --crosstool_top=//third_party/toolchains/gcc7_manylinux2010:toolchain\n"
+            )
+            bazel_rc.write(
+                "build:linux_ci_gpu --crosstool_top=//third_party/toolchains/gcc7_manylinux2010-nvcc-cuda10.1:toolchain\n"
+            )
+            bazel_rc.write("build:output_ci --noshow_progress\n")
+            bazel_rc.write("build:output_ci --noshow_loading_progress\n")
+            bazel_rc.write("build:output_ci --verbose_failures\n")
+            bazel_rc.write("build:output_ci --test_output=errors\n")
+            # GCS cache (read-only by default)
+            bazel_rc.write(
+                "build:cache --remote_cache=https://storage.googleapis.com/tensorflow-sigs-io\n"
+            )
+            bazel_rc.write("build:cache --remote_upload_local_results=false\n")
+
             bazel_rc.close()
     except OSError:
         print("ERROR: Writing .bazelrc")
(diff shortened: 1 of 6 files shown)


vnvo2409:
FYI, I successfully span up a MacOS Actions with Docker: https://github.com/vnvo2409/ci/runs/2639182573?check_suite_focus=true, with the `docker-compose` for emulate testing enviroment and servers, we could increase testing coverage on MacOS to the same as on Linux.

tools/build/configure.py
```diff
@@ -114,16 +114,37 @@ def write_config():
...
             bazel_rc.write('build:macos --copt="-DGRPC_BAZEL_BUILD"\n')
             # Stay with 10.14 for macOS
             bazel_rc.write('build:macos --copt="-mmacosx-version-min=10.14"\n')
             bazel_rc.write('build:macos --linkopt="-mmacosx-version-min=10.14"\n')
             # MSVC (Windows): Standards-conformant preprocessor mode
             bazel_rc.write('build:windows --copt="/Zc:preprocessor"\n')
+            # Config for CI and release build
+            bazel_rc.write("build:ci --copt=-msse4.2\n")
+            bazel_rc.write("build:ci --copt=-mavx\n")
+            bazel_rc.write("build:ci --compilation_mode=opt\n")
```
yongtang, line 130: Maybe we can change `build:ci` to `build:optimization` to better represent the option like?
```
build:optimization --copt=-msse4.2
build:optimization --copt=-mavx
build:optimization --compilation_mode=opt
```

tools/build/configure.py
```diff
@@ -114,16 +114,37 @@ def write_config():
...
+            bazel_rc.write("build:mac_ci --copt=-Wunguarded-availability\n")
```
yongtang, line 131: I think `--copt=-Wunguarded-availability` can be part of the `build:macos` as there is no impact even for normal build?
```
build:macos --copt=-Wunguarded-availability
```

tools/build/configure.py
```diff
@@ -114,16 +114,37 @@ def write_config():
...
+            bazel_rc.write(
+                "build:linux_ci --crosstool_top=//third_party/toolchains/gcc7_manylinux2010:toolchain\n"
+            )
+            bazel_rc.write(
+                "build:linux_ci_gpu --crosstool_top=//third_party/toolchains/gcc7_manylinux2010-nvcc-cuda10.1:toolchain\n"
+            )
+            bazel_rc.write("build:output_ci --noshow_progress\n")
+            bazel_rc.write("build:output_ci --noshow_loading_progress\n")
+            bazel_rc.write("build:output_ci --verbose_failures\n")
+            bazel_rc.write("build:output_ci --test_output=errors\n")
```
yongtang, line 141: Maybe we can fold all into part of the normal build, as they are not impactful on the binary output even for normal build?
```
build --noshow_progress
build --noshow_loading_progress
build --verbose_failures
build --test_output=errors
```

We can also add `--experimental_ui_max_stdouterr_bytes=-1` here as well:
```
build --experimental_ui_max_stdouterr_bytes=-1
```

tools/build/configure.py
```diff
@@ -114,16 +114,37 @@ def write_config():
...
+            # GCS cache (read-only by default)
+            bazel_rc.write(
+                "build:cache --remote_cache=https://storage.googleapis.com/tensorflow-sigs-io\n"
+            )
+            bazel_rc.write("build:cache --remote_upload_local_results=false\n")
```
yongtang, line 146: Nice 👍

yongtang:
@vnvo2409 I do have some hesitation about using docker on macOS. I remember we  encountered quite some issue with docker on macOS as it is not completely reliable (flaky) with GitHub Actions. Maybe we can defer this to a later time?

tools/build/configure.py
vnvo2409: Thank you !

tools/build/configure.py
vnvo2409: Done !

tools/build/configure.py
vnvo2409: Done !

tools/build/configure.py
vnvo2409: Done !

vnvo2409:
@yongtang How about a `docker-compose` for Linux only ?

tools/build/configure.py
vnvo2409: This actually wil invalidate our `macOS` build cache. The `Bazel macOS` seems to be invalidated as well. However, it is only a temporary problem, we are fine.

yongtang:
> How about a docker-compose for Linux only ?

I think that could be a solution. Though we may want to allow people to launch only one or two services (not bring everything in one batch) in case some people is only debugging one issue.

yongtang (approved):

vnvo2409:
> Though we may want to allow people to launch only one or two services (not bring everything in one batch) in case some people is only debugging one issue.

It won't be a problem. Because people could use `docker-compose up {service_name}` to launch only one service at a time or `docker-compose up` to launch everything.

PR #1438, opened 2021-05-28
Add default value to --project args in setup.py

yongtang:
In order to support tensorflow-io-gcs-filesystem wheel
we added the option of `--project` in setup.py
to selectively pass either `tensorflow-io` or ` tensorflow-io-gcs-filesystem`

However, it looks like after that, the `Used by` map in GitHub does not shown
up (likely caused by setup.py does not work without `--project` value anymore).

This PR add default value (`tensorflow-io`) to setup.py to hopefully
get the `Used by` map back in GitHub.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -148,7 +148,7 @@ jobs:
           set -x -e
           python -m pip install -U wheel setuptools
           python --version
-          python setup.py --project tensorflow-io --data bazel-bin -q bdist_wheel --plat-name macosx_10_14_x86_64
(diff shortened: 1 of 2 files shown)


kvignesh1420 (approved):
LGTM. Let's see if that mapping comes back.

kvignesh1420:
@yongtang can you check this reference and see if changes are required in repo "Settings": 

https://docs.github.com/en/code-security/supply-chain-security/understanding-your-software-supply-chain/exploring-the-dependencies-of-a-repository#changing-the-used-by-package.

PR #1443, opened 2021-06-01
Allow configuring gcs file system in runtime.

yongtang:
This is a working in progress PR to have set_configuration support for configuring gcs in runtime.

This PR depends on https://github.com/tensorflow/tensorflow/pull/49951

FYI @vnvo2409 

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -153,10 +153,10 @@ pip_install()
 
 http_archive(
     name = "org_tensorflow",
-    sha256 = "233875ea27fc357f6b714b2a0de5f6ff124b50c1ee9b3b41f9e726e9e677b86c",
-    strip_prefix = "tensorflow-2.5.0",
+    sha256 = "204c027f82e09266ba4412f6bdbcb1cb678ffc1ee36d89193392f79545106293",
+    strip_prefix = "tensorflow-eeaf1597c43d1913d0127566064209d7ad4359b5",
     urls = [
-        "https://github.com/tensorflow/tensorflow/archive/refs/tags/v2.5.0.tar.gz",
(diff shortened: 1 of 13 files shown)


yongtang:
This PR is ready for review now. The tensorflow dependency has also been bumped to 2.6.0rc0 to match the upcoming 2.6.0 release.

yongtang:
With tensorflow 2.6.0 the s3 and hdfs only exists on tensorflow-io so we do want to make sure everything works fine.

kvignesh1420 (approved):
@yongtang this looks good. Thanks!!

kvignesh1420:
> With tensorflow 2.6.0 the s3 and hdfs only exists on tensorflow-io so we do want to make sure everything works fine.

we can update our tests to verify the "configuration on fly" op for s3 and hdfs as well. WDYT?

yongtang:
@kvignesh1420 Yes. In fact starting tensorflow 2.6, user will have to install tensorflow-io to get S3 and HDFS support. And they don't need `TF_USE_MODULAR_FILESYSTEM=1` to switch on/off.

`TF_USE_MODULAR_FILESYSTEM=1` is only needed for gcs switch (we plan on also remove once we can get GCS/TPU in place within modular file system).

We will need to make sure we can release tensorflow-io 0.20.0 (next version)  the same time tensorflow 2.6 is released.

jsimsa:
@yongtang Could you please contribute the Python API and C++ operation for setting file system options introduced by this PR to the core TF repository? There is a need for that API internally and it would be more convenient for that API to exist in core TF instead of tensorflow_io. Thank you.

yongtang:
@jsimsa  Sure I can take a look.

jsimsa:
Hi @yongtang, do you have an ETA for this? This is currently blocking an internal user.

jsimsa:
Thank you for the update Yong.

PR #1449, opened 2021-06-03
bump `aws-sdk-cpp` to 1.8.186

vnvo2409:
Since the symbol collision is resolved, I think we could try to bump `aws-sdk-cpp`.

### DIFF
--- a/tensorflow_io/core/plugins/s3/s3_filesystem.cc
+++ b/tensorflow_io/core/plugins/s3/s3_filesystem.cc
@@ -134,37 +134,6 @@ static Aws::Client::ClientConfiguration& GetDefaultClientConfig() {
   absl::MutexLock l(&cfg_lock);
 
   if (!init) {
-    const char* region = getenv("AWS_REGION");
-    // TODO (yongtang): `S3_REGION` should be deprecated after 2.0.
-    if (!region) region = getenv("S3_REGION");
-    if (region) {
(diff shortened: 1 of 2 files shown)


tensorflow_io/core/plugins/s3/s3_filesystem.cc
```diff
@@ -134,37 +134,6 @@ static Aws::Client::ClientConfiguration& GetDefaultClientConfig() {
...
-          }
-        }
-        Aws::Config::AWSConfigFileProfileConfigLoader loader(config_file);
-        loader.Load();
-        auto profiles = loader.GetProfiles();
-        if (!profiles["default"].GetRegion().empty())
-          cfg.region = profiles["default"].GetRegion();
-      }
-    }
     // if these timeouts are low, you may see an error when
```
yongtang, line 137: @vnvo2409 Is deletion above on purpose?

vnvo2409: @yongtang Yes as reading region will be done automatically by [`aws-sdk-cpp`](https://github.com/aws/aws-sdk-cpp/wiki/What’s-New-in-AWS-SDK-for-CPP-Version-1.8#client-configuration-now-reads-environment-variables-configuration-file-and-ec2-metadata-to-get-default-aws-region).

yongtang (approved):
LGTM, thanks!

Issue #1440, opened 2021-05-29
Improve project directory layout


kvignesh1420:
The following is a proposal to restructure the `tensorflow-io` directory layout for improving the modularity and maintainability of the project.

The current directory layout (with tree depth 2) is as follows:
```
tensorflow_io
├── BUILD
├── __init__.py
├── arrow
│   ├── BUILD
│   ├── README.md
│   ├── __init__.py
│   ├── kernels
│   ├── ops
│   └── python
├── bigquery
│   ├── BUILD
│   ├── README.md
│   ├── __init__.py
│   ├── kernels
│   ├── ops
│   └── python
├── core
│   ├── BUILD
│   ├── __init__.py
│   ├── go
│   ├── grpc
│   ├── kernels
│   ├── ops
│   ├── plugins
│   ├── python
│   └── swift
├── ignite
│   ├── README.md
│   ├── __init__.py
│   └── python
└── kafka
    ├── __init__.py
    └── python
```

It can be observed that:
- `kafka`, `ignite`, `bigquery` and `arrow` have their own directories with `kernels`, `ops` and `python` to expose their set of API's.
- `core` encompasses all other APIs.

The following are the changes that can be made to improve the consistency with this layout.

move:
- [x] `arrow/kernels` to `core/kernels` (https://github.com/tensorflow/io/pull/1442)
- [x] `arrow/ops` to `core/ops` (https://github.com/tensorflow/io/pull/1442)
- [x] `arrow/python` to `core/python` (https://github.com/tensorflow/io/pull/1442)
- [x] `bigquery/kernels` to `core/kernels` (https://github.com/tensorflow/io/pull/1446)
- [x] `bigquery/ops` to `core/ops` (https://github.com/tensorflow/io/pull/1446)
- [x] `bigquery/python` to `core/python` (https://github.com/tensorflow/io/pull/1446)
- [x] `ignite/python` to `core/python` (https://github.com/tensorflow/io/pull/1447)
- [x] `kafka/python` to `core/python` (https://github.com/tensorflow/io/pull/1447)

After this change, the layout should look something like this:
```
tensorflow_io
├── BUILD
├── __init__.py
├── core
│   ├── BUILD
│   ├── __init__.py
│   ├── go
│   ├── grpc
│   ├── kernels
│   ├── ops
│   ├── plugins
│   ├── python
│   └── swift
│── arrow.py
│── bigquery.py
│── ignite.py
│── kafka.py
```
[...]

PR #1450, opened 2021-06-03
migrate the python ops directory outside of core

kvignesh1420:
This PR addresses #1440 by moving the `tensorflow_io/core/python` directory to `tensorflow_io/python`. This enables the developers to clearly distinguish the `core` kernels/ops from python API's. Thus, aligning with the structure of the `tf/tf` repo.

Additionally, the BUILD file in `tensorflow_io/core` has been split so that we have the flexibility to place the `cc_binary` targets outside `core`.

Also, `plugins` as of now mean file-system plugins. However, I feel the name has to be changed to `filesystems` to avoid confusion.


yongtang (approved):
LGTM, Thanks!

PR #1466, opened 2021-06-25
reset mongo cursor after reaching end of collection

kvignesh1420:
This PR resets the mongo cursor after it has reached the end of the collection. This is necessary to restart the fetching process for each epoch in the training process.

Additionally:
- Remove the hard-coded `appname` to allow the users to set an `appname` from the URI itself.
- The BSON to JSON conversion has been modified to simplify parsing each dataset record into a json.


kvignesh1420:
cc: @yongtang

yongtang (approved):
LGTM

PR #1476, opened 2021-07-21
Remove build temporary folder btween wheel creation

casassg:
bdist_wheel runs install which runs the build step for setup.py. Due to running bdist_wheel for both wheels generated, the first command adds the tfio library files into the build folder, after that, the second command (tfio_gcs_filesystem) takes the tfio files that are left in build folder by previous command and adds them as data dir.

This can be avoided if you remove the build directory between the 2 wheel generation steps which effectively makes it more hermetic.

This is an issue on the production wheel!!!!

To validate, try installing tensorflow_io_gcs_filesystem alone. That will trigger installing tensorflow-io python files.

Example:

```
python -m venv env
. env/bin/activate
pip install tensorflow_io_gcs_filesystem==0.19.0
ls env/lib/python3.7/site-packages/tensorflow_io
```

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -149,6 +149,7 @@ jobs:
           python -m pip install -U wheel setuptools
           python --version
           python setup.py --data bazel-bin -q bdist_wheel --plat-name macosx_10_14_x86_64
+          rm -rf build
           python setup.py --project tensorflow-io-gcs-filesystem --data bazel-bin -q bdist_wheel --plat-name macosx_10_14_x86_64
       - name: Auditwheel ${{ matrix.python }} macOS
         run: |
@@ -376,6 +377,7 @@ jobs:
           python --version
           python -m pip install -U wheel setuptools
           python setup.py --data bazel-bin -q bdist_wheel
+          rm -rf build
           python setup.py --project tensorflow-io-gcs-filesystem --data bazel-bin -q bdist_wheel
           ls -la dist
       - uses: actions/upload-artifact@v2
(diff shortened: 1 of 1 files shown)


casassg:
Would appreciate if a release could be cut with the script generating the release doing this step of manually removing the build folder. This is blocking some of our usage of TFIO with [pex](https://github.com/pantsbuild/pex)

kvignesh1420 (approved):
I verified this fix on my mac and it works fine. Thanks for the fix @casassg 

We will release a new version (0.20.0) soon (soon after the tf 2.6.0 release) and the wheels should be fixed.

kvignesh1420:
@casassg you can use the nightly releases to verify the fix.

casassg:
@kvignesh1420 any chance to release 0.19.1 with this commit. Ideally, we need a TF 2.5 compatible release (our monorepo takes time to upgrade TF and this issue is making some of our code flaky)

yongtang:
@casassg We can create a branch on 0.19 and if you can create the same PR to 0.19 branch then I think it is possible to release 0.19.1.

kvignesh1420:
@yongtang since the PR has already been merged, we can proceed with the branch cut I guess.

casassg:
Yes, I can do the PR. Lmk once the branch is created and I can do it

yongtang:
@casassg Branch of R0.19 has been created: https://github.com/tensorflow/io/tree/R0.19

casassg:
There we go: https://github.com/tensorflow/io/pull/1478

PR #1482, opened 2021-07-26
Fix decode_video returning only the first frame

### LINKED ISSUE #1471: tfio.experimental.ffmpeg.decode_video returns only first frame repeated in some contexts
When using `tensorflow-io>=0.18` in an Ubuntu 18.04 based image (e.g. the official Tensorflow images) `tfio.experimental.ffmpeg.decode_video(data)` gives rise to a warning that `libavformat.so.58` is missing and then returns a tensor of correct shape but where all the frames are the same.

**Steps to reproduce**
I used the small.mp4 test file from the repo here.

`test.py`
```
import numpy as np
import tensorflow as tf
import tensorflow_io as tfio


data = tf.io.read_file("small.mp4")
video = tfio.experimental.ffmpeg.decode_video(data)

assert video.shape == (166, 320, 560, 3)

# Just checking difference between first and last frame.
diff = np.abs(video[0] - video[-1]).sum()
assert diff != 0
```

`Dockerfile`
```
FROM tensorflow/tensorflow:2.5.0
RUN apt-get update && apt-get install -y ffmpeg
RUN pip install tensorflow_io==0.19.0
COPY small.mp4 .
COPY test.py .
ENTRYPOINT ["python", "test.py"]
```

**Output**
```
/usr/local/lib/python3.6/dist-packages/tensorflow_io/python/ops/ffmpeg_ops.py:43: UserWarning: could not load libtensorflow_io_ffmpeg_4.2.so: unable to open file: libtensorflow_io_ffmpeg_4.2.so, from paths: ['/usr/local/lib/python3.6/dist-packages/tensorflow_io/python/ops/libtensorflow_io_ffmpeg_4.2.so']
caused by: ['libavformat.so.58: cannot open shared object file: No such file or directory']
  warnings.warn("could not load {}: {}".format(library, e))
  File "test.py", line 10, in <module>
    assert diff != 0
AssertionError
```

With`tensorflow_io==0.17.0` it does not fail like this.

kyamagu:
This PR fixes the bug in `tfio.experimental.ffmpeg.decode_video` returning the repeated first frame.

Fix #1471

### DIFF
--- a/tensorflow_io/core/kernels/ffmpeg_kernels.cc
+++ b/tensorflow_io/core/kernels/ffmpeg_kernels.cc
@@ -669,7 +669,7 @@ class FFmpegVideoStream : public FFmpegStream {
     char* base = ((char*)(value->flat<uint8>().data()));
     int64 datasize = height_ * width_ * channels_;
     for (size_t i = 0; i < frames_.size(); i++) {
-      memcpy(base, reinterpret_cast<char*>(frames_buffer_.front().get()),
+      memcpy(base, reinterpret_cast<char*>(frames_buffer_[i].get()),
              datasize);
       base += datasize;
     }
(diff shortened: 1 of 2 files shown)


kvignesh1420 (approved):
LGTM. Thanks @kyamagu

PR #1492, opened 2021-08-04
Clarifications for bigquery connector

vlasenkoalexey:
- Added code snippet describing how to get better performance when reading data from BQ with large batch sizes or wide rows
- Clarified meaning of the requested_streams argument

### DIFF
--- a/tensorflow_io/bigquery.md
+++ b/tensorflow_io/bigquery.md
@@ -87,9 +87,14 @@ def main():
       data_format=BigQueryClient.DataFormat.AVRO)
   dataset = read_session.parallel_read_rows()
   dataset = dataset.batch(10)
+  dataset = dataset.prefetch(10)
+  # You can optionally cache dataset, so data is loaded from BigQuery only once
+  # during first iteration. For subsequent iterations data is going to be loaded
+  # from cache.
+  dataset = dataset.cache()
 
   row_index = 0
-  for row in dataset.prefetch(10):
+  for row in dataset:
     print("row %d: %s" % (row_index, row))
     row_index += 1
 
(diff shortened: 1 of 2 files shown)


tensorflow_io/bigquery.md
```diff
@@ -95,7 +103,28 @@ if __name__ == '__main__':
 
 ```
 
-It also supports reading BigQuery column with repeated mode (each field contains array of values with primitive type: Integer, Float, Boolean, String, but RECORD is not supported). In this case, selected_fields needs be a dictionary in a
+In some cases when when row is too wide (has 30+ columns) or if you are doing batching with large batch sizes it might be benefitial to do batching before interleave to get a better performance. Here is an exaple showing how to do that:
```
kvignesh1420, line 106: duplicate `when` in first line.

kvignesh1420: `benefitial` --> `beneficial` in second line

kvignesh1420: `exaple` --> `example` in third line

tensorflow_io/bigquery.md
```diff
@@ -95,7 +103,28 @@ if __name__ == '__main__':
...
+streams_ds = tf.data.Dataset.from_tensor_slices(streams)
+dataset = streams_ds.interleave(
+    read_rows,
+    cycle_length=streams_count,
+    num_parallel_calls=streams_count,
+    deterministic=False)
+...
+```
+
+Connector also supports reading BigQuery column with repeated mode (each field contains array of values with primitive type: Integer, Float, Boolean, String, but RECORD is not supported). In this case, selected_fields needs be a dictionary in a
```
kvignesh1420, line 127: The last line is incomplete

tensorflow_io/bigquery.md
vlasenkoalexey: so many typos, thanks for reviewing

tensorflow_io/bigquery.md
vlasenkoalexey: updated

kvignesh1420 (approved):
LGTM. Thanks @vlasenkoalexey

yongtang (approved):
LGTM, Thanks @vlasenkoalexey

PR #1508, opened 2021-08-25
<immintrin.h> is x86-64 only

hrw:
do not use it on other architectures


yongtang:
Thanks @hrw . The PR looks good. The CI error is unrelated and is being fixed in PR #1509.

hrw:
There will be more changes. I am now in 'hack & slash' phase trying to build it and then will extract changes into small commits like this one.

Issue #1504, opened 2021-08-17
Impossible to load directory of parquet files using map due to tf.io.gfile.isdir 


GeraudK:
Hi,

I'm trying to load a list of parquet files in a directory. For this I do:
```
files = tf.data.Dataset.list_files(dir_path)
dataset = files.map(lambda x: tfio.IODataset.from_parquet(x, columns=...))
```
I'm getting the following error:
```
TypeError: Expected binary or unicode string, got <tf.Tensor 'f:0' shape=() dtype=string>
```
Which seems to be coming from `tf.io.gfile.isdir` in v0.20.0 [here](https://github.com/tensorflow/io/blob/master/tensorflow_io/python/ops/parquet_dataset_ops.py#L30)

You can try the following example to replicate the error:
```
tf.io.gfile.isdir(tf.constant("/path/to/file"))
```
And you will get the same error. Is there anyway to use map or interleave to load a directory of parquet files?

Thank you for you help!

kvignesh1420:
@GeraudK I see. Will take a look.

kvignesh1420:
@GeraudK I have opened PR #1521 to address this issue.

On a related note, @yongtang you seem to be working on this issue with this PR: https://github.com/tensorflow/tensorflow/pull/33590. Any plans on going forward with it?

kvignesh1420:
@GeraudK as PR #1521 has been merged, you can use `tensorflow-io-nightly` for the fix. Thanks for reporting the issue.

Issue #1468, opened 2021-06-28
IODataset.from_parquet segfaults when filename is a directory


GuyPozner:
When passing a directory of parquet parts to IODataset.from_parquet filename, it produces a segmentation fault, which can be reproduced using this python script:

Python script to reproduce the issue:

```python
import os

import numpy as np
from pyspark.sql import SparkSession, Row
from pyspark.sql.types import StructType, StructField, BinaryType, IntegerType
import tensorflow as tf
import tensorflow_io as tfio

NUM_SAMPLES = 1000
PARQUET_DIRPATH = "/tmp/dummy.parquet"

def create_parquet():
    xs = np.random.randint(0, high=10, size=NUM_SAMPLES, dtype=np.uint8).flatten().tolist()
    ys = np.random.randint(0, high=10, size=NUM_SAMPLES, dtype=np.uint8).flatten().tolist()
    schema = StructType([StructField("xs", IntegerType(), True),
                         StructField("ys", IntegerType(), True)])
    
    spark = SparkSession.builder.master('local[1]').getOrCreate()
    sc = spark.sparkContext
    
    def row_generator(i):
        row = {"xs": xs[i], "ys": ys[i]}
        return row

    rows_rdd = sc.parallelize(range(NUM_SAMPLES)).map(row_generator).map(lambda x: Row(**x))
    spark.createDataFrame(rows_rdd, schema).write.mode("overwrite").parquet(PARQUET_DIRPATH)
    print("Done creating parquet.")

def parquet_file_to_dataset():
    example_struct = {"xs": tf.int32, "ys": tf.int32}
    filename = [f for f in os.listdir(PARQUET_DIRPATH)
                if f.endswith(".parquet")][0]
    filepath = os.path.join(PARQUET_DIRPATH, filename)
    dataset = tfio.IODataset.from_parquet(filepath, example_struct)
    return dataset

def parquet_dir_to_dataset():
    example_struct = {"xs": tf.int32, "ys": tf.int32}
    dataset = tfio.IODataset.from_parquet(PARQUET_DIRPATH, example_struct)
    return dataset

def print_first_example(dataset):
    example = next(iter(dataset))
    print(example)

if __name__ == "__main__":
    print(f"Tensorflow Version: {tf.__version__}")
    print(f"Tensorflow_io Version: {tfio.__version__}")
    create_parquet()
    ds1 = parquet_file_to_dataset()
    print_first_example(ds1)
    ds2 = parquet_dir_to_dataset()
[...]

PR #1521, opened 2021-09-03
fix the parquet dataset creation with list_files and map.

kvignesh1420:
This PR addresses the issue: https://github.com/tensorflow/io/issues/1504 along with https://github.com/tensorflow/io/issues/1468. The change moves the `filename` IsDirectory check to the cc layer from python layer. This allows the users to use `tf.data.Dataset.list_files(...)` and `map()` to load multiple parquet files into a dataset.
Tests have been added to assert this behaviour.


kvignesh1420:
cc: @yongtang @GeraudK

yongtang (approved):
LGTM, Thanks!

PR #1526, opened 2021-09-09
Fix temporary file issue in s3

### LINKED ISSUE #1519: S3 driver creates temporary files in current directory
As a result if a process is executed on an environment where whole FS is read-only except `/tmp` directory - uploading to S3 fails with `Could not append to the internal temporary file`.

That's not the case with GCS which tries to find tmp dir:
https://github.com/tensorflow/io/blob/6110049ff73ed5dc85046737f71057401b679eac/tensorflow_io_gcs_filesystem/core/gcs_helper.cc#L46 =>
https://github.com/tensorflow/tensorflow/blob/17774a11b576186e439174298af82b096dcd5582/tensorflow/core/platform/path.cc#L329

Compare with S3 which just passes null prefix downstream:
https://github.com/tensorflow/io/blob/6110049ff73ed5dc85046737f71057401b679eac/tensorflow_io/core/filesystems/s3/s3_filesystem.cc#L358

TF S3 implementation just uses `/tmp`:
https://github.com/tensorflow/tensorflow/blob/17774a11b576186e439174298af82b096dcd5582/tensorflow/core/platform/s3/s3_file_system.cc#L57

Quick workaround can be to `os.chdir("/tmp")` if acceptable.

yongtang:
This PR tries to fix the temporary file issue in s3 where temp files are created within the current directory.

This PR maintains the behavior of old s3 file system.

This PR fixes #1519.

### DIFF
--- a/tensorflow_io/core/filesystems/s3/s3_filesystem.cc
+++ b/tensorflow_io/core/filesystems/s3/s3_filesystem.cc
@@ -355,9 +355,15 @@ typedef struct S3File {
         s3_client(s3_client),
         transfer_manager(transfer_manager),
         outfile(Aws::MakeShared<Aws::Utils::TempFile>(
-            kS3FileSystemAllocationTag, nullptr, "_s3_filesystem_XXXXXX",
+            kS3FileSystemAllocationTag,
+#if defined(_MSC_VER)
+            nullptr,
+#else
+            "/tmp/_s3_filesystem_XXXXXX",
+#endif
             std::ios_base::binary | std::ios_base::trunc | std::ios_base::in |
-                std::ios_base::out)) {}
+                std::ios_base::out)) {
+  }
 } S3File;
 
 void Cleanup(TF_WritableFile* file) {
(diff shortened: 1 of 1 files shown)


terrytangyuan (approved):

tensorflow_io/core/filesystems/s3/s3_filesystem.cc
```diff
@@ -355,9 +355,15 @@ typedef struct S3File {
         s3_client(s3_client),
         transfer_manager(transfer_manager),
         outfile(Aws::MakeShared<Aws::Utils::TempFile>(
-            kS3FileSystemAllocationTag, nullptr, "_s3_filesystem_XXXXXX",
+            kS3FileSystemAllocationTag,
+#if defined(_MSC_VER)
+            nullptr,
```
kvignesh1420, line 360: @yongtang where will be the temporary files be created if it is a nullptr?

yongtang:
@kvignesh1420 The temp file will be on the same directory on Windows. Added a comment in the code which matches:
https://github.com/tensorflow/tensorflow/blob/17774a11b576186e439174298af82b096dcd5582/tensorflow/core/platform/s3/s3_file_system.cc#L52-L58

kvignesh1420 (approved):
LGTM. Thanks @yongtang

PR #1535, opened 2021-09-29
Fix HDFS append to non-existent file error

burgerkingeater:
This PR fixes the error below:
Writing to HDFS from the snapshot writer generates the following exception
```
Caused by: org.apache.hadoop.ipc.RemoteException(java.io.FileNotFoundException): Failed to append to non-existent file 
```

This is because HDFS expects files opened for "append" to already exist. This change checks the existence of files, and if it does not exist, then opens them for "write" instead of "append".

### DIFF
--- a/tests/test_hdfs.py
+++ b/tests/test_hdfs.py
@@ -40,3 +40,54 @@ def test_read_file():
     content = tf.io.read_file("hdfs://{}:9000/file.txt".format(address))
     print("CONTENT: {}".format(content))
     assert content == body
+
+
+@pytest.mark.skipif(
+    sys.platform in ("win32", "darwin"),
+    reason="TODO HDFS not setup properly on macOS/Windows yet",
+)
+def test_append_non_existing_file():
+    """Test case for append a non-existing HDFS file"""
+
+    address = socket.gethostbyname(socket.gethostname())
+    print("ADDRESS: {}".format(address))
+
+    body = b"1234567"
+    filepath = "hdfs://{}:9000/non-existing.txt".format(address)
+    f = tf.io.gfile.GFile(filepath, "a")
+    f.write(body)
+    f.flush()
+
+    content = tf.io.read_file(filepath)
+    print("CONTENT: {}".format(content))
+    assert content == body
+    f.close()
+
+
+@pytest.mark.skipif(
+    sys.platform in ("win32", "darwin", "linux"),
+    reason="TODO HDFS not setup properly on macOS/Windows yet. For linux, current HDFS setup will throw:"
+    "java.io.IOException: Failed to replace a bad datanode on the existing pipeline due to no more"
+    "good datanodes being available to try. ",
+)
+def test_append_existing_file():
+    """Test case for append an existing HDFS file"""
+
+    address = socket.gethostbyname(socket.gethostname())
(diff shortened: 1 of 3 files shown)


burgerkingeater:
@kvignesh1420 can you take a look? Thanks.

kvignesh1420:
@burgerkingeater thanks for the fix, do you have a test case by any chance that we can add to the suite?

yongtang:
@burgerkingeater I think adding a simple test in python will be preferable for ease of debugging?

burgerkingeater:
@kvignesh1420 can you take a look? Thanks

burgerkingeater:
@kvignesh1420 gently ping..

kvignesh1420:
@burgerkingeater will take a look.

tests/test_hdfs.py
```diff
@@ -40,3 +40,54 @@ def test_read_file():
...
+    assert content == body
+    f.close()
+
+
+@pytest.mark.skipif(
+    sys.platform in ("win32", "darwin", "linux"),
+    reason="TODO HDFS not setup properly on macOS/Windows yet. For linux, current HDFS setup will throw:"
+    "java.io.IOException: Failed to replace a bad datanode on the existing pipeline due to no more"
+    "good datanodes being available to try. ",
+)
```
kvignesh1420, line 72: @burgerkingeater can we enable this test for Linux and check? Those CI failures were due to flaky runs.

burgerkingeater:
This particular failure as commented in reason section is not flaky test
failure. I have reran few times and it’s happening consistently.

tests/test_hdfs/hdfs_test.sh
```diff
@@ -21,6 +21,16 @@ HADOOP_RELEASE_TAG="3.2.1"
...
+# Add following properties
+# to prevent following error when closing the hdfs client:
+#
+# java.io.IOException: Failed to replace a bad datanode on the existing pipeline due to no more good datanodes
+# being available to try.
+
+echo "
+HDFS_CONF_dfs_client_block_write_replace___datanode___on___failure_enable=true
+HDFS_CONF_dfs_client_block_write_replace___datanode___on___failure_policy=NEVER" >> hadoop.env
+
```
kvignesh1420, line 33: @burgerkingeater I see, I thought this was fixing that issue.

kvignesh1420 (commented):
@burgerkingeater we can skip that test for now. However, did you get a chance to check if this test passes with a different(local/on-prem) hadoop cluster?

burgerkingeater:
@kvignesh1420 I haven't tested in on-prem hadoop cluster due to the setup complexity.

As the best practice, we should leverage automated unit tests to guard the change as much as possible. That's why I'm curious about the motivation of the fact we deleted unit test (https://github.com/tensorflow/tensorflow/blob/v2.6.0/tensorflow/core/platform/hadoop/hadoop_file_system_test.cc).

burgerkingeater:
@kvignesh1420  can you take a look again?

kvignesh1420 (approved):
@burgerkingeater we removed the cpp unit tests as the python level testing would implicitly cover cpp level functionality. Let's merge this PR as of now so that the append issue is fixed. Sorry for the delay.

PR #1536, opened 2021-10-03
Create SECURITY.md

Jeverett3000:


mihaimaruseac (changes requested):
It should point to the tensorflow's one, instead of being the default file

PR #1537, opened 2021-10-03
Add support for Azure Shared Access Signatures

janbernloehr:
This PR adds support for Azure Shared Access Signatures to tf io. Previously, account keys were the only supported means of authentication. However, account keys are often not available in corporate environments since they do not support fine-grained access control or expiration.


janbernloehr:
> @googlebot I signed it!

janbernloehr:
Some internal feedback I got was that since sas tokens can only be created one per container, that we should support multiples of them. I will update the PR with that feature as well and also add some documentation.

yongtang:
@janbernloehr Can you fix the lint errors? You can run:
```
bazel run //tools/lint:lint -- clang
```

janbernloehr:
Most tests have passed now in at least one config. Maybe triggering again and this time we are lucky?

PR #1542, opened 2021-10-11
fix: longjmp causes uninitialized stack frame in libcurl

janbernloehr:
This fixes "longjmp causes uninitialized stack frame in libcurl" which we run into for longer trainings reading from azure blob storage.

For more info on the error see https://stackoverflow.com/questions/9191668/error-longjmp-causes-uninitialized-stack-frame


yongtang:
The CI build issue is being fixed in #1544

yongtang:
@janbernloehr Thanks for the PR! As #1544 has been merged, can you try rebasing your PR to the latest master in order to fix the CI failures?

Issue #1548, opened 2021-10-21
Support a variant for tensorflow-rocm


elukey:
Hi everybody!

At the Wikimedia Foundation we use AMD ROCm GPUs, and we are currently testing their latest drivers/tools (4.3.1) and the related new `tensorflow-rocm` 2.6 (a fork of the official `tensorflow` that supports AMD GPUs). The main issue that we are encountering is related to Hadoop HDFS support, since IIUC it has been offloaded to `tensorflow-io`, that in turns requires `tensorflow` when installed (that doesn't play well, in my tests, with `tensorflow-rocm`).

I created https://github.com/ROCmSoftwarePlatform/tensorflow-upstream/issues/1461 on the ROCm end as well.

Since `tensorflow-rocm` is a drop-in replacement of `tensorflow` (or at least, it should be) I am wondering if you could add a variant to `setup.py` to variate the dependency to tensorflow if needed. For example, `pip install tensorflow-io[rocm]` would try to install `tensorflow-rocm` and not `tensorflow`. Would it be possible/feasible in your opinion?

Thanks in advance,

Luca

yongtang:
@elukey That sounds good to me. Never used tensorflow-rocm before so I may need to spend some time to take a look before a PR. But in the meantime, if you can create a PR that will be better.

elukey:
@yongtang any feedback? :)

yongtang:
@elukey Sorry I am out of office and have limited access to my usual workstation. Tt is a little slowed to setup the build env. I will try to get env setup soon to work on this issue while on the road..

elukey:
Ack thanks! @jayfurmanek if you have time can you check/review the pull request in the meantime?

jayfurmanek:
I don't have review authority on tensorflow-io, but I did try your PR and it does work as expected.

PR #1551, opened 2021-10-22
setup.py: allow tensorflow variants stated as extras_require settings

elukey:
The AMD ROCm project maintains a fork of the Tensorflow package
called 'tensorflow-rocm', a drop-in replacement for 'tensorflow'.
The fact that tensorflow-io states only 'tensorflow' as specific
dependency makes it hard to use 'tensorflow-rocm', since both
don't work whell in the same environment.

This change follows what's done for tensorflow-addons in
tensorflow/addons@61ebf40,
namely:
- Move the 'tensorflow' requirement to extras_require.
- Add new extras_require for tensorflow-{rocm,cpu,gpu}

### DIFF
--- a/.github/workflows/build.wheel.sh
+++ b/.github/workflows/build.wheel.sh
@@ -8,7 +8,8 @@ run_test() {
   CPYTHON_VERSION=$($entry -c 'import sys; print(str(sys.version_info[0])+str(sys.version_info[1]))')
   (cd wheelhouse && $entry -m pip install tensorflow_io_gcs_filesystem-*-cp${CPYTHON_VERSION}-*.whl)
   (cd wheelhouse && $entry -m pip install tensorflow_io-*-cp${CPYTHON_VERSION}-*.whl)
-  $entry -m pip install -q pytest pytest-benchmark pytest-xdist boto3 fastavro avro-python3 scikit-image pandas pyarrow==3.0.0 google-cloud-pubsub==2.1.0 google-cloud-bigtable==1.6.0 google-cloud-bigquery-storage==1.1.0 google-cloud-bigquery==2.3.1 google-cloud-storage==1.32.0 PyYAML==5.3.1 azure-storage-blob==12.8.1 azure-cli==2.29.0
+  TF_VERSION=$(/usr/bin/grep tensorflow tensorflow_io/python/ops/version_ops.py | /usr/bin/cut -d '"' -f 2)
+  $entry -m pip install -q $TF_VERSION pytest pytest-benchmark pytest-xdist boto3 fastavro avro-python3 scikit-image pandas pyarrow==3.0.0 google-cloud-pubsub==2.1.0 google-cloud-bigtable==1.6.0 google-cloud-bigquery-storage==1.1.0 google-cloud-bigquery==2.3.1 google-cloud-storage==1.32.0 PyYAML==5.3.1 azure-storage-blob==12.8.1 azure-cli==2.29.0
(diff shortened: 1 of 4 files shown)


elukey:
@yongtang I tried to update the pull request with what suggested by @jayfurmanek. From a quick local test (bdist wheel + pip install) it seems that it works as intended.

One caveat remains though - in order to build the wheel I had to install `tensorflow` in my pypi venv. This is due to how __init__.py of `ops` and `api` are called from `setup.py` includes, but it should be fine from what I can see.

Lemme know your thoughts, I tried the least invasive change and since I am not super familiar with the package I'll be open to any direction/suggestion you want to take from here. Thanks!

jayfurmanek:
https://github.com/tensorflow/io/issues/1548

yongtang:
@elukey Sorry for the late reply. Can you resolve the merge conflict so that we can trigger the CI build?

elukey:
@yongtang Hi! No problem, thanks a lot for your help :) The pull request should be updated, lemme know!

elukey:
I just noticed the issue, namely https://github.com/tensorflow/io/runs/4175314538?check_suite_focus=true

This is what I was referring to above, namely that tensorflow is needed to build the package. Going to figure out a workaround.

elukey:
@yongtang I may have found a simpler and less invasive way to apply the fix, PR updated :)

kvignesh1420:
@elukey I have approved and started the CI jobs, will review based on their status. Thanks!

elukey:
@kvignesh1420 Thanks! I tried to add a workaround to `build.wheel.sh` to install `tensorflow`, lemme know if it is good to test!

elukey:
The patch works, but for some reason the Test 3.7 Windows fails to find `numpy`. This seems unrelated, is it an open issue?

yongtang:
@elukey While Linux & Mac uses `io/.github/workflows/build.wheel.sh` for testing setup, the Windows setup is done directly in:
https://github.com/tensorflow/io/blob/master/.github/workflows/build.yml#L385-L438

 Can you add similar installation of tensorflow package on Windows and give it a try?

elukey:
@yongtang PR updated, not sure if the fix is enough (don't know how to test it properly).

yongtang:
@elukey  I think we are almost there. The Windows build is failing which is likely because Windows cmd on GitHub Actions does not work well with `"'.

Can you replace the Windows install tensorflow line to the following:
```
(grep tensorflow tensorflow_io/python/ops/version_ops.py | sed -e "s/require = //g" | xargs python -m pip install)
```

I think the above line will resolve the issue.

elukey:
@yongtang done!

elukey:
mmm it fails for
```
D:\a\io\io>(/usr/bin/grep tensorflow tensorflow_io/python/ops/version_ops.py   | sed -e "s/require = //g"   | xargs python -m pip install ) 
The system cannot find the path specified.
```

elukey:
@yongtang all working now!

kvignesh1420 (approved):
Thanks for the PR @elukey

elukey:
Really happy that the pull request is getting approved :)
If it gets merged, would it be feasible to release a 0.22.1 version of `tensorflow-io` to allow `tensorflow-rocm` users to use it?

yongtang (approved):

yongtang:
Thanks @elukey for the PR. I think ideally we can release a 0.23.0 instead as it is a change that requires user changing the installation.

PR #1559, opened 2021-11-11
Enforce pyupgrade to Python 3.7+

yongtang:
Since we only support Python 3.7+, it makes sense to enforce pyupgrade to 3.7+.


kvignesh1420 (approved):
LGTM. Thanks @yongtang

PR #1560, opened 2021-11-14
Remove python 3.6 from setup.py description.

yongtang:
This PR removes python 3.6 from setup.py description
as support has been dropped.

In addition, this PR also:
1. Reformat setup.py with black and pyupgrade (py3.7+).
2. Re-introduce bacel cache for Windows build, as our build has been stablized with TF 2.7.


kvignesh1420 (approved):
LGTM

PR #1565, opened 2021-11-17
az: Improve error messages

janbernloehr:
This PR adds additional information for all error messages of the azure bindings. In particular, in case the azure storage sdk raises an exception, the status code, error code, and error message are propagated. Also, the storage path which the exception belongs to is added. Finally, `TF_VLog` calls are added to the most relevant functions.


kvignesh1420 (approved):
Thanks for the improvements @janbernloehr

PR #1588, opened 2021-12-14
upgrading docker devel image to use python3.7 and updating documentation

vlasenkoalexey:
Devel docker image is not longer usable because latest release doesn't have whls for Python 3.6, therefore updating  Python to 3.7.
Also updated documentation, references to python3 probably should be replaced by python since support for python2 was dropped.

### DIFF
--- a/docs/development.md
+++ b/docs/development.md
@@ -217,35 +217,6 @@ TFIO_DATAPATH=bazel-bin \
     'python3 -m pytest -s -v tests/test_serialization.py'
 ```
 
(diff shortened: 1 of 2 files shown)


kvignesh1420 (approved):
LGTM.

kvignesh1420:
The lint failure should be fixed with https://github.com/tensorflow/docs/commit/741a39f0a765dc87569d259469c4c0e994637a11

yongtang (approved):

PR #1614, opened 2022-01-20
Implement `InputDatasets` for BigQueryDatasetOp

dulacp:
Fixes the following warnings:

```
E tensorflow/core/framework/dataset.cc:580] UNIMPLEMENTED: Cannot compute input sources for dataset of type IO>BigQueryDataset, because the dataset does not implement `InputDatasets`.
E tensorflow/core/framework/dataset.cc:584] UNIMPLEMENTED: Cannot merge options for dataset of type IO>BigQueryDataset, because the dataset does not implement `InputDatasets`.
```

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -304,11 +304,11 @@ http_archive(
 http_archive(
     name = "dcmtk",
     build_file = "//third_party:dcmtk.BUILD",
-    sha256 = "a05178665f21896dbb0974106dba1ad144975414abd760b4cf8f5cc979f9beb9",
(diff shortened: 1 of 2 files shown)


yongtang:
@dulacp I am wondering if the DCMTK failure is only a temporarily thing? I think it will likely succeed with another run even without the change?

yongtang:
Never mind, since DCMTK move to github officially, we will need to change it anyway.

yongtang (approved):
Never mi

PR #1616, opened 2022-01-21
Fix parquet in graph execution

dulacp:
My first commit adds a failing test `test_parquet_dataset_ordered_dict_in_graph_execution` to demonstrate the issue to resolve. In a nutshell the shape is set to `<unknown>` in graph execution, which is undesirable and not consistent with the eager execution.

Here is an extract of the failing test output:

```
E         + OrderedDict([('boolean_field',
E         -               TensorSpec(shape=(), dtype=tf.bool, name=None)),
E         ?                                ^^
E         +               TensorSpec(shape=<unknown>, dtype=tf.bool, name=None)),
E         ?                                ^^^^^^^^^
E         -              (b'int32_field',
E         ?               -
E         +              ('int32_field',
E         -               TensorSpec(shape=(), dtype=tf.int32, name=None)),
E         ?                                ^^
E         +               TensorSpec(shape=<unknown>, dtype=tf.int32, name=None)),
E         ?                                ^^^^^^^^^

...
```

My second commit is fixing this test with a workaround.
Let me know if you have a better fix in mind.

Cheers

### DIFF
--- a/tensorflow_io/python/ops/parquet_dataset_ops.py
+++ b/tensorflow_io/python/ops/parquet_dataset_ops.py
@@ -81,7 +81,8 @@ class ParquetIODataset(tf.data.Dataset):
                 indices_start = tf.data.Dataset.range(0, shape[0], step)
                 indices_stop = indices_start.skip(1).concatenate(
                     tf.data.Dataset.from_tensor_slices(
-                        tf.convert_to_tensor([shape[0]], tf.int64))
+                        tf.convert_to_tensor([shape[0]], tf.int64)
+                    )
                 )
                 dataset = tf.data.Dataset.zip((indices_start, indices_stop))
 
--- a/tensorflow_io/python/ops/parquet_dataset_ops.py
+++ b/tensorflow_io/python/ops/parquet_dataset_ops.py
@@ -59,9 +59,12 @@ class ParquetIODataset(tf.data.Dataset):
                 return dtype
 
             if columns is not None:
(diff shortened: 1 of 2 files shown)


yongtang:
@dulacp  Wondering if you can take a look at the test failure as it seems to be related.

dulacp:
I refactored my code and changed the approach to fix all tests. Indeed the failure was related, now all tests should pass :)

tensorflow_io/python/ops/parquet_dataset_ops.py
```diff
@@ -58,15 +47,28 @@ def dtype_f(dtypes, components, column):
                 dtype = tf.as_dtype(dtype.numpy())
                 return dtype
 
-            if columns is not None:
+            if not tf.executing_eagerly():
+                assert columns is not None
+                assert isinstance(columns, dict)
```
kvignesh1420, line 52: @dulacp I think we should raise exceptions with proper messages here instead of an `assert` so that users can better understand the issue.

dulacp: Thanks for pointing it out. I changed this into a `ValueError` exception and added two tests to validate the behavior and avoid regressions.

dulacp (commented):
I fixed your comment in my latest commit, raising a proper exception.

yongtang (approved):

kvignesh1420 (approved):
LGTM. Thanks @dulacp

PR #1628, opened 2022-02-07
retrieve previous version from the tags list and avoid recalculation

kvignesh1420:
This PR modifies the logic to retrieve the previous version of the package by using the tags list instead of explicit calculations.

cc: @yongtang I think with this change, we don't need to go down the `semver` path as this current logic achieves the purpose. However, we can use it in the future based on requirements now that we have it noted in our discussions.


yongtang (approved):
LGTM, Thanks!

PR #1633, opened 2022-02-09
decrease max bytes read from hdfs

lyeeeee:
The origin max bytes allowed read/write from/to hdfs is INT_MAX-2.
But it may raise JVM OOM in some JVM implementations.
For example:
java version "1.8.0_91"
Java(TM) SE Runtime Environment (build 1.8.0_91-b14)
Java HotSpot(TM) 64-Bit Server VM (build 25.91-b14, mixed mode)

it will raise oom when size == 2147483645

Here we choose INT_MAX - 8.(java.util.ArrayList use it as the max length).


kvignesh1420 (approved):
Thanks for the change @lyeeeee 

[Reference](https://hg.openjdk.java.net/jdk8/jdk8/jdk/file/tip/src/share/classes/java/util/ArrayList.java) for  where this limitation was imposed.

PR #1635, opened 2022-02-10
Add release workflow

yongtang:
This PR adds a release workflow that can be triggered through Actions => All Workflows => Release => Run Workflow. A tag and a commit has to be provided.

The action will fetch the candidate binary from dockerhub, cut the head of the RELEADE.md, and create a release.

Follow up steps:
1. The release is a draft (draft: true) now. Once we are feeling confident, we can switch to draft: false
2. The publishing to pypi is comment out for now. Once we are feeling confident, we can uncomment.

Note: The old release.yml has been renamed to release.note.yml to better represent the content.

A try run is available in another repo: https://github.com/yongtang/io/runs/5134338413?check_suite_focus=true

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,28 +1,72 @@
-name: Release Bot
+name: Release
 
 on:
-  push:
-    branches:
-      - master
+  workflow_dispatch:
+    inputs:
+      version:
+        description: "Version (e.g., v0.24.0)"
+        required: true
+      commit:
+        description: "Commit (e.g., 92b44e1)"
+        required: true
 
 jobs:
   release:
-    name: README.md
+    name: Release
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
         with:
           fetch-depth: 0
-      - run: git tag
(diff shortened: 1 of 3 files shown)


.github/workflows/release.yml
```diff
@@ -1,28 +1,72 @@
...
+          body_path: CURRENT.md
+          name: ${{ steps.info.outputs.name }}
+          tag_name: ${{ steps.info.outputs.tag }}
+          target_commitish: ${{ steps.info.outputs.commit }}
+          draft: true
+      #- uses: pypa/gh-action-pypi-publish@master
+      #  with:
+      #    user: __token__
+      #    password: ${{ secrets.github_tensorflow_io_nightly }}
+
```
kvignesh1420, line 72: @yongtang can we add a step to push it to `pypi-test` and then to `pypi`? WDYT?

yongtang:
@kvignesh1420 Thanks for the review. I have update the PR to include a step to push to test.pypi.org. Once we can see it works once or twice then we will be able to push pypi.org in live.

kvignesh1420 (approved):
Thanks @yongtang

PR #1637, opened 2022-02-10
Disable failing test on macOS

yongtang:
Recently brew upgraded ffmpeg to 5.0 and as such some tests are
failing on macOS.

This PR disables failing tests on macOS.

In order to fix the issue, we will need to upgrade the ffmpeg
dependency to 5.0 (a follow up).


kvignesh1420 (commented):
@yongtang this test should also be disabled:
```
test_audio.py::test_audio_io_dataset_with_dataset[from_ffmpeg]
```

kvignesh1420 (approved):

PR #1656, opened 2022-04-01
Update log level from fetal to error when loading the libhdfs.so failed

zuston:
Sorry i'm not familiar with the codebase of tensorflow/io. But i found that when no libhdfs.so in the path of HADOOP_HDFS_HOME environment var, it will directly absort when using tensorflo2.8

Howerver, in our linux machine, it exist in /usr/lib64/libhdfs.so. And it should fall back to load /usr/lib64 path.

Besides everything is OK in tensorflow2.4


zuston:
@burgerkingeater @yongtang Could u help check this PR? Thanks

yongtang (approved):
LGTM. Thanks! (and sorry for the late review)

PR #1657, opened 2022-04-03
Update filter_ops.py

trabenx:
currently for every value of the 'mode' parameter other than 'None' a NameError exception is thrown (NameError: name 'upper' is not defined)
This is due to misuse of the upper() method


yongtang:
@trabenx  Looks good to me though can you sign the CLA as was mentioned in the comment https://github.com/tensorflow/io/pull/1657#issuecomment-1086881382 ?

kvignesh1420 (approved):

PR #1662, opened 2022-04-18
Fix release note, and prepare for 0.25.0 release

yongtang:


yongtang:
This is a mechanical change to create release note so will just merge in.

PR #1673, opened 2022-04-27
Bump orc version to latest stable 1.6.14

372046933:
Got latest stable information from here
https://orc.apache.org/docs/releases.html#all-releases


yongtang (approved):
LGTM. Sorry for the late response and thanks for the contribution!

372046933:
BTW, the reason of `undefine HAS_BUILTIN_OVERFLOW_CHECK`, i.e. not using `__builtin_mul_overflow` and `__builtin_add_overflow` is for portability concern. Is that right?

yongtang:
@372046933 yes the build failed on windows without the change of undefine HAS_BUILTIN_OVERFLOW_CHECK

PR #1702, opened 2022-08-10
Add arrow binary data type support

yongtang:

### DIFF
--- a/tests/test_arrow.py
+++ b/tests/test_arrow.py
@@ -1230,6 +1230,41 @@ class ArrowDatasetTest(ArrowTestBase):
 
         os.unlink(f.name)
 
+    def test_arrow_feather_dataset_binary(self):
+        """test_arrow_feather_dataset_binary"""
+        import tensorflow_io.arrow as arrow_io
+
+        from pyarrow import feather as pa_feather
+        import pyarrow as pa
+        import numpy as np
+
+        local_path = os.path.join(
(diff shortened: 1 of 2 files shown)


tests/test_arrow.py
```diff
@@ -1230,6 +1230,41 @@ def test_arrow_list_feather_columns(self):
...
+            os.path.dirname(os.path.abspath(__file__)), "test_image", "sample.png"
+        )
+        with open(local_path, "rb") as f:
+            data = [f.read()]
+            table = pa.Table.from_arrays([data], ["data"])
+
+        chunk_size = 1000
+        with tempfile.NamedTemporaryFile(delete=False) as f:
+            pa_feather.write_feather(table, f, chunksize=chunk_size)
+
```
kvignesh1420, line 1251: @yongtang thanks for adding this feature. I saw that `StringType` is a sub-class of `BinaryType` [Ref](https://arrow.apache.org/docs/cpp/api/datatype.html#_CPPv4N5arrow10StringTypeE) so can we add one more test where we read from a string typed pa table?

yongtang:
@kvignesh1420 The PR has been updated to include an additional test case. Please take a look.

kvignesh1420:
@yongtang LGTM, let's see if the tests pass this time. The lint test fix might need a separate PR.

yongtang:
@kvignesh1420  Thanks. Will merge in for now and will create a separate PR for follow up the other fix.

PR #1703, opened 2022-08-16
Fix lint build failure

yongtang:
Fix lint build failure


yongtang:
Will merge as lint failures are fixed now.

Issue #1736, opened 2022-11-25
3.11 wheels seem to block themselves


llchan:
I'm currently seeing this:
```
Processing ./tensorflow_io_gcs_filesystem-0.28.0-cp311-cp311-manylinux_2_12_x86_64.manylinux2010_x86_64.whl
ERROR: Package 'tensorflow-io-gcs-filesystem' requires a different Python: 3.11.0 not in '<3.11,>=3.7'
```
My guess is that the [setup.py line](https://github.com/tensorflow/io/blob/45ee85c420b47842b00236e3ad3c420fbc731516/setup.py#L169) did not get updated when the rest of the 3.11 wheel stuff was set up?

PR #1740, opened 2022-11-30
Allow installing for Python 3.11

agkphysics:
This fixes issue #1736. Since Python 3.11 wheels are already available, this PR just makes `setup.py` consistent with that.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -156,6 +156,7 @@ setuptools.setup(
         "Programming Language :: Python :: 3.8",
         "Programming Language :: Python :: 3.9",
         "Programming Language :: Python :: 3.10",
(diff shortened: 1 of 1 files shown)


vient:
@yongtang is it possible to reupload PyPI 3.11 packages with this fix backported?

feliperyan:
Seems to still be causing issues for `poetry add tensorflow`

larsschellhas:
Limiting the allowed Python version in my pyprojet.toml helped!

Just set the python limits to this and it should work fine: ">=3.11,<3.12"

It's because there is no tensorflow-io-gcs-filesystem package that is compatible with versions higher than 3.11

PR #1831, opened 2023-08-18
Bump TF 2.14.0rc0

yongtang:


mnowakhabana:
@yongtang Do you have plan to complete this change? Any ETA ?

third_party/libapr1.BUILD
```diff
@@ -99,7 +99,6 @@ cc_library(
         "tables/apr_tables.c",
         "threadproc/unix/proc.c",
         "threadproc/unix/procsup.c",
-        "threadproc/unix/signals.c",
```
aostrowski-hbn, line 102: Can we restore this line? Removing it breaks importing the package on Linux due to `undefined symbol: apr_proc_kill`

yongtang: Updated

PR #1957, opened 2024-03-11
Handle tensorflow version pinning in setup.py

sdenton4:


mihaimaruseac (commented):
I think this would make sense, but it impact release process as now all these need to be bumped.

Instead we rely on `tensorflow` specifying the dependency with a pin and users installing only compatible versions

sdenton4:
The "users only install compatible versions" is broken from a user-perspective, though - it means downstream users need to keep track of the special constraints between versions of tf and tf-io, and manage the constraints manually. This commonly leads to broken installations and confused users. 

The entire point of specifying dependencies explicitly is that they can be handled automatically, as the tooling can automatically find a set of versions satisfying all constraints, without burdening users of libraries depending on your library... I would really advise taking some momentary pain (changing the release process) to make it easier for everyone to use your libraries.

Issue #1912, opened 2024-01-01
S3 filesystem pure virtual method called; terminate called without an active exception


rivershah:
I am getting a core dump during interpreter teardown, when using the s3 filesystem. Can I please be given guidance how to handle this issue. Please see script to reproduce inside docker:

`FROM tensorflow/tensorflow:2.14.0-gpu`

The following environment variables are set

```
"AWS_ACCESS_KEY_ID": xxx,
"AWS_SECRET_ACCESS_KEY": xxx,
"AWS_ENDPOINT_URL_S3": xxx,
"AWS_REGION": "us-east-1",
"S3_USE_HTTPS": "1",
"S3_VERIFY_SSL": "1",
"S3_DISABLE_MULTI_PART_DOWNLOAD": "0",
"S3_ENDPOINT": xxx,
```

```
import os

import tensorflow as tf
import tensorflow_io as tfio

def illustrate_core_dump():
    print(f"tf version: {tf.__version__}")
    print(f"tfio version: {tfio.__version__}")
    filename = f"{os.environ['CLOUD_MOUNT']}/tmp/test_tfrecord.tfrecord"
    assert filename.startswith("s3://"), "problem appears to be be for s3 filesystem only"
    ds = tf.data.TFRecordDataset(filename, "GZIP")

    for i in ds:
        print(f"i.shape: {i.shape}")


if __name__ == "__main__":
    illustrate_core_dump()
    print("reaches here successfully")
    print("something broken during destruction and tf")

    # during interpreter teardown if s3 filesystem used we will get
    # pure virtual method called
    # terminate called without an active exception
    # Aborted (core dumped)

    # gs:// and file:// do not exhibit this issue which don't rely on tfio
```

```
TF_CPP_MIN_LOG_LEVEL=0 python notebooks/illustrate_core_dump.py 
2024-01-01 18:07:11.253238: E tensorflow/compiler/xla/stream_executor/cuda/cuda_dnn.cc:9342] Unable to register cuDNN factory: Attempting to register factory for plugin cuDNN when one has already been registered
2024-01-01 18:07:11.253287: E tensorflow/compiler/xla/stream_executor/cuda/cuda_fft.cc:609] Unable to register cuFFT factory: Attempting to register factory for plugin cuFFT when one has already been registered
[...]

PR #2119, opened 2025-01-18
fix(filesystem): Fix crash when using S3

dante-lee:
- Related: #1912 
- Just splitted `GetExecutor` function.
  - `executor_pool_size` is retrieved from `GetExecutorPoolSize`(new implemented)
  - And `GetExecutor` receives new argument named `executor_pool_size`
- Checked exception by pure virtual method call does not happend anymore.

### DIFF
--- a/tensorflow_io/core/filesystems/s3/s3_filesystem.cc
+++ b/tensorflow_io/core/filesystems/s3/s3_filesystem.cc
@@ -215,13 +215,13 @@ static int GetExecutorPoolSize() {
   return temp_value;
 }
 
-static void GetExecutor(tf_s3_filesystem::S3File* s3_file) {
+static void GetExecutor(tf_s3_filesystem::S3File* s3_file, int executor_pool_size) {
   absl::MutexLock l(&s3_file->initialization_lock);
 
   if (s3_file->executor.get() == nullptr) {
     s3_file->executor =
         Aws::MakeShared<Aws::Utils::Threading::PooledThreadExecutor>(
-            kExecutorTag, kExecutorPoolSize);
+            kExecutorTag, executor_pool_size);
   }
 }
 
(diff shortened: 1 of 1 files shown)


tensorflow_io/core/filesystems/s3/s3_filesystem.cc
```diff
@@ -203,34 +203,35 @@ static void GetS3Client(tf_s3_filesystem::S3File* s3_file) {
...
+// This function returns executor_pool_size that is
+// used in initialization of executor.
+// Ref: https://github.com/tensorflow/io/issues/1856
+static int GetExecutorPoolSize() {
   int temp_value;
   const char* executor_pool_size = getenv("S3_EXECUTOR_POOL_SIZE");
   if (executor_pool_size == nullptr ||
       !absl::SimpleAtoi(executor_pool_size, &temp_value))
     temp_value = kExecutorPoolSize;
+  return temp_value;
```
mihaimaruseac, line 215: If the if condition fails, `temp_value` is unitialized

mihaimaruseac (changes requested):
This is semantically equivalent with the existing code, there is no actual change from this PR.

dante-lee:
@mihaimaruseac 
Yes you are right. But exception is gone with this change.(Teste on Ububtu 22.04) I'm looking for reason why it works, but if it works how about applying it? Since it seems like to happen quite a long time and it prevents to use tf >= 2.16.

tensorflow_io/core/filesystems/s3/s3_filesystem.cc
dante-lee: Seems like other codes in this source code initialize value from environment like this(actually I did not change any logic except for function splitting as you know) I understood it works like this

1. env var does not exist -> initialized with kExecutorPoolSize
2. env var exists
  2.1. cannot be parsed into integer -> initialized with kExecutorPoolSize
  2.2. can be parsed into integer -> initialized with given env var

If my understanding is wrong plz tell me. Thanks.

dante-lee:
@yongtang Can you review this PR? To resolve #1912 I saw that you released v0.37.1 but it seems it doesn't resolved issue. And #1857 which you've approved seems related to this.

dante-lee:
I just rebuilt tensorflow-io with command [here](https://github.com/tensorflow/io/issues/1912#issuecomment-1942855023) and it seems like that v0.37.1 without modifying source code works fine(i.e no pure virtual method called exception). Therefore I think this code change will not resolve crash.

I gonna close this PR and open new PR when I found root cause.

